PR #13, opened 2018-07-30
optimizer class, with tests

dangilman:
I added the optimizer class. 

There is an initial run with a particle swarm algorithm, followed by a downhill simplex "amoeba" routine that Chuck recommends. 

The Particle Swarm implementation is slightly edited from the default CosmoHammer routine - I coded up some new convergence criteria that will terminate the swarm once the standard deviation of the last N samples is less than some critical value. The critical value can be pretty lax, since the particle swarm run doesn't need to actually converge fully. I added some tests, too. 

Finally, I added a function in util.py that sorts images by position (used in the tests).


sibirrer:
Impressive work. The tests for python3 crash, due to print statement convention:
https://travis-ci.org/sibirrer/lenstronomy/jobs/409940528

sibirrer:
Some other tests failed
https://travis-ci.org/sibirrer/lenstronomy/jobs/409940527
try on your local machine:
>>> py.test test/test_LensModel/test_Optimizer/

dangilman:
Hi,

I think I figured out what was wrong (the multiplane flag was set to ‘False’ in the ‘test_multiplane' routine…..…… ) 

For the print statement, should I just remove it for now to be python 3 compatible? Or do you want to change it on your end? 

- Daniel

sibirrer:
for the print statement, you can keep them, just use print("this text gets printed") as a function call with brackets()

PR #20, opened 2018-08-05
Re optimization update

dangilman:
The user can specify 're_optimize' which will adjust the initial particle swarm optimization to start clustered around the input 'kwargs_lens' parameters; this speeds up the particle swarm step considerably as long as the starting lens model is reasonably good. There is also a new option 'particle_swarm' that allows the user to skip the particle swarm optimization altogether, and go straight to the downhill simplex optimization. 

I've also added a 'default_settings' file with default parameters used in the optimization.

### DIFF
--- a/lenstronomy/LensModel/Optimizer/default_settings.py
+++ b/lenstronomy/LensModel/Optimizer/default_settings.py
@@ -1,0 +1,31 @@
+"""
+This file contains default settings for the optimization class. If the settings are not explicitly specified in the
+call to Optimizer, the setting will default to the parameters in this class
+"""
+
+# the tolerance on lens model source position (used to fit image positions)
+default_tol_source = 1e-5
+# the tolerance on lens model magnification (used to fit magnifications)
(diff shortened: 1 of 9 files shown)


sibirrer:
In the tests, I put the initialisation of the class methods in a def setup() definition. This caused a merge conflict which I resolved in inciting on the setup() initialisation. Some of your updated test functions might require variables that were not defined as class variables. Sorry for that.

sibirrer:
python3 seems to be a bit more restrictive in comparing variables with different types than python2:

self = <lenstronomy.LensModel.Optimizer.multi_plane.MultiPlaneOptimizer object at 0x7f8c5a21b7f0>
    def _test_convergence(self):
    
>       if self._counter > self._n_particles and np.mean(self.src_penalty[-self._n_particles:]) < self._pso_convergence_mean:
E       TypeError: unorderable types: float() < NoneType()
lenstronomy/LensModel/Optimizer/multi_plane.py:146: TypeError

Sorry for that.

PR #30, opened 2018-08-16
Approx shear

dangilman:
Hi I'm wondering if you can take a look at something: I notice that when computing the finite magnification with the shifting background approximation the images look significantly different, in that the images produced by guessing an SIS macromodel tend to be bigger (thicker and more stretched). I have a hunch this is because the SIS deflection field is constant while the real one has a a quadrupole from the shear component. So, I added a feature in the Background class that will add a shear correction to the SIS guess for computing the finite magnifications. Have a look at the MultiPlaneLensing class and the test_multi_plane class. Do you think this will work in general?

### DIFF
--- a/lenstronomy/LensModel/Optimizer/multi_plane.py
+++ b/lenstronomy/LensModel/Optimizer/multi_plane.py
@@ -1,6 +1,7 @@
 from lenstronomy.LensModel.lens_model import LensModel
 import numpy as np
 from lenstronomy.Util.util import approx_theta_E
+import matplotlib.pyplot as plt
 
 class MultiPlaneLensing(object):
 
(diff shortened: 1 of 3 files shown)


sibirrer:
Hi Daniel, I see. What probably needs to match is the Hessian (shear and convergence) of the guessed profile and the proposed macro-model. You are right, you can probably accurately enough correct for that by computing the local Hessian of the two models and then relative to that distort the rays. I would make it a bit more general such that the full Hessian matrix correction is applied. The convergence term is probably very similar in your example as you are using fixed power-law profiles in the testing.

sibirrer:
accepted more recent pull request

PR #32, opened 2018-08-22
Composite class

dangilman:
About the single background approximation thing: Fixing the definition of the delta_betas seemed to immediately make everything else work, I added separate test functions for that class in test_single_background. I've also started playing around with it with full realizations and so far so good 

About the TNFW: thanks for catching that! I replaced the expression in the _g(x, tau) function with a (hopefully correct) version and I get reasonable looking deflection angles, and all the tests still pass. I also added a test function that compares the hessian to numerically computed derivatives of the deflections.


sibirrer:
thanks Daniel,
would you mind making a simple test of TNFW deflections against gravlens?
I am adding a numerical differential test to the test_numeric_differentials.py and make sure the function _F() does not return nan if X==1 (there is a divide by zero at the moment)

PR #36, opened 2018-09-20
Hybrid background approx

dangilman:
The single background approximation now does full ray tracing for a while until it gets close to a solution, then computes the delta beta's and does the approximation for the rest of the optimization.


sibirrer:
Hi Daniel, can you make some tests of the new lines?
Thanks,
Simon

PR #37, opened 2018-10-05
Ray shooting partial step

dangilman:
I added a routine in LensModel/multiplane.py that does "ray_shooting_partial", but returns the comoving ray positions, a list of lens plane redshifts (unique redshifts, not the full list), and the comoving distances to each lens plane. 

I am using this function now in a new approach to the optimization that will hopefully speed things up.


sibirrer:
Hi Daniel, is the coverage decreased because you removed redundant parts of the code or because new lines are not tested. If it is the first, I am happy to approve the merge. Thanks, Simon

dangilman:
Hi, 

Yeah the first request I made had untested lines (actually from a previous pull request) so in a later one I removed the untested part of the code (actually removed the whole single background part) 

- Daniel

dangilman:
Wait sorry never mind, looks like there are a few untested lines. Let me fix that and submit a new pull request 

- Daniel

PR #43, opened 2018-11-06
Adding the possibility to use the multi-threading in cosmoHammer

martin-millon:
Minor modification to add a keyword argument in FittingSequence.fit_sequence(threadCount=8) and propagate it to cosmoHammer.
It reduces computational time by a factor 2-3 for a PSO fitting on my laptop, using 8 cores. 
This should not be used simultaneously with mpi.


sibirrer:
is it possible to delete the .pyc files from the pull request? Thanks.

martin-millon:
Yes. It's done.

PR #44, opened 2018-11-25
Cored Burkert profile

dangilman:
implemented a Burkert profile with a variable core size


sibirrer:
Hi Daniel, thank you for implementing the cored Burkert profile!
Would you mind removing the plotting parts (I know, they are not getting executed, but they will be marked as not tested code) and check the documentation of the new class whether the references are right? Thank you!

PR #46, opened 2018-11-25
core Burkert

dangilman:
I removed a couple lines from coreBurkert.py, but I don't see what plotting commands you're referring to


sibirrer:
I was looking not on your last commit so I saw some intermediate files. Sorry, should have looked at your last commit only.

PR #49, opened 2018-11-26
coreBurkert

dangilman:
fixed tests, changed to theta_Rs


sibirrer:
thank you! Some python2 conventions make some problems:
https://travis-ci.org/sibirrer/lenstronomy/jobs/459576230

sibirrer:
hmm, int to float conversion is different in python2 and 3. This might cause the remaining discrepancy.
https://travis-ci.org/sibirrer/lenstronomy/jobs/459605616

PR #50, opened 2018-11-28
HYBRID lens model

dangilman:
A hybrid lens model profile for defining a superposition of two lens models with a user-specified interpolating function. I want to use this to blend a cored Burkert profile into an NFW profile as r_core -> 0. It can also be extended to other composite lens models, like Sersic+NFW

### DIFF
--- a/lenstronomy/LensModel/Profiles/hybrid.py
+++ b/lenstronomy/LensModel/Profiles/hybrid.py
@@ -16,9 +16,10 @@ class HYBRID(object):
     for each lens class
 
     For example, interpolating between an NFW profile and a coreBURKERT profile,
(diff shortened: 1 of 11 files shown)


dangilman:
Hi,

Yeah I worried about that. I could do a superposition of lens models outside of lenstronomy, but I worry that computations would be twice as slow because the length of lens_model_list doubles. Even though inside the HYBRID class one has to compute deflection angles sequentially, so the total number of function calls to e.g. ‘derivatives' would be the same, since the main lens_model class uses list comprehension to sort through all of these profiles I worry things would get quite slow if you just double the length of lens_model_list. But its not urgent, we can discuss it later

- Daniel

sibirrer:
I don't expect much of speed up as the data crunching still happens in the numpy arrays. A good thing is to profile your code to see where the bottle necks are, and if you notice an unexpected one, you can write an issue. I close this pull request in the meantime.

PR #54, opened 2019-01-31
Cored nfw

dangilman:
I implemented everything except for the potential. mathematica is giving me a hard time with the integral. For the time being I just default to the potential of a regular nfw profile


sibirrer:
Hi Daniel, did you push this to master? I just wonder because travis-ci was not triggered as it did in some earlier pull requests.
Maybe it's better if there is a raise statement for the potential to make sure no-one is using it in case.

dangilman:
Hi, 

Hmm I pushed it to my fork and then created a pull request, usually that is all it takes. And I raise a warning in the potential computation

- Daniel

PR #65, opened 2019-03-11
Numerical deflections

dangilman:


sibirrer:
Hi Daniel, I see that in your new function you pass **kwargs to the interpolation and kwargs['center_x'] to the coordinate transform. I'd assume you do not want to pass the rotation and centering parameters to the interpolation function and have those things split. So you can explicitly pass those arguments that are common to the function and those specific to the underlying interpolation, such as center_x, center_y, **kwargs_interp

dangilman:
Right I’ll change it

PR #74, opened 2019-05-08
small edit to optimizer initialization 

dangilman:
when using a shear chi^2 penalty, initialize particle swarm with the desired external shear value


sibirrer:
is the coverage decreased because of lines previously been tested have been removed or new lines added are not tested?

dangilman:
I think the former but I’ll check

Daniel

sibirrer:
is is possible to add some tests to increase coverage?

PR #76, opened 2019-05-24
elliptical sersic model with ellipticity in convergence 

dangilman:


sibirrer:
Hi Daniel, this looks good. Would you mind:
1. rename the function into 'SERSIC_ELLIPSE_KAPPA' and the existing one in 'SERSIC_ELLIPSE_POTENTIAL?
2. make test to see that the computed convergence from the Hessian actually corresponds to the scaled surface brightness of an elliptical Sersic profile? So far, the tests are just testing vs some numerical numbers.
Thanks you!
Simon

PR #83, opened 2019-06-03
Lensed position convention

dangilman:


sibirrer:
Hi Daniel, I like how you implemented this in a general way!
In LensedLocation __call__, it works when there is only one lens model in one plane indicated as physical positions. If there are two or more, we have to take into account the ray-tracing of the corrected position of the <z lenses before computing the higher z lens. So there needs to be a sorting of the indexes in increasing redshift and then applying the multi-plane with the updated new_kwargs.

Minor comments:
- document the argument(s) you added in the __init__ statements
- are the 'check_convention' statements in the function arguments needed or should this only be set by the class instance themselves? (are there reasonable cases where you want to use this argument differently within the same instance of LensModel? If not necessary, keeping it simpler is better.

Thanks a lot
Simon

lenstronomy/LensModel/Optimizer/multi_plane_optimizer.py
```diff
@@ -33,51 +33,62 @@ def __init__(self, full_lensmodel, x_pos, y_pos, lensmodel_params, z_source,
...
-            self._foreground = Foreground(halo_lensmodel, self._z_macro, x_pos, y_pos)
+            self._foreground = Foreground(halo_lensmodel, macromodel_lensmodel, self._z_macro, x_pos, y_pos)
 
         self._halo_args = halo_args
 
-        self._model_to_vary = ToVary(macromodel_lensmodel, self._z_macro)
-        self._macro_args = macro_args
+        self._halo_lensmodel = halo_lensmodel
+
+    def set_kwargs(self, kwargs_lens_full):
```
sibirrer: do you want this function to be public? Just to check. Otherwise make it _set_kwargs()

lenstronomy/LensModel/image_position_convention.py
```diff
@@ -0,0 +1,47 @@
...
+
+        new_kwargs = deepcopy(kwargs_lens)
+
+        for ind in self._inds:
+
+            theta_x = kwargs_lens[ind]['center_x']
+            theta_y = kwargs_lens[ind]['center_y']
+            zstop = self._multiplane._redshift_list[ind]
+            x, y, _, _ = self._multiplane.ray_shooting_partial(0, 0, theta_x,
+                            theta_y, 0, zstop, kwargs_lens, check_convention = False)
```
sibirrer: don't you need here the new_kwargs being used to ray-trace to the next lens model? (at least after the first one). If you tested this case, the test should fail at the moment.

lenstronomy/LensModel/lensed_image_positions.py
```diff
@@ -0,0 +1,35 @@
...
+
+from copy import deepcopy
+
+class PhysicalLocation(object):
+
+    def __call__(self, kwargs_lens):
+
+        return kwargs_lens
+
+class LensedLocation(object):
```
sibirrer: add a short description of what this class does.

lenstronomy/LensModel/lensed_image_positions.py
```diff
@@ -0,0 +1,35 @@
+__author__ = 'dgilman'
+
+from copy import deepcopy
+
+class PhysicalLocation(object):
```
sibirrer: add a short description of the role of this class (I know, trivial but it still helps reading the code)

lenstronomy/LensModel/lensed_image_positions.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __call__(self, kwargs_lens):
+
+        new_kwargs = deepcopy(kwargs_lens)
+
+        for ind in self._inds:
+
+            theta_x = kwargs_lens[ind]['center_x']
+            theta_y = kwargs_lens[ind]['center_y']
+            zstop = self._multiplane._redshift_list[ind]
```
sibirrer: here you access a private function of another class (meaning not necessary stable and possibly not so well documented). I can make a 'proper' function that returns the redshift of a specific lens model index (basically executes the private function underneath).

lenstronomy/LensModel/lensed_image_positions.py
```diff
@@ -0,0 +1,35 @@
...
+            x, y, _, _ = self._multiplane.ray_shooting_partial(0, 0, theta_x,
+                            theta_y, 0, zstop, kwargs_lens, check_convention = False)
+
+            D = self._multiplane._T_z_list[ind]
```
sibirrer: this is also a private function of another class. Don't worry at the moment (also for my comment above). I will think of how we can make it accessible for outside use (and better documented what these variables actually are).

lenstronomy/LensModel/multi_plane.py
```diff
@@ -2,17 +2,21 @@
...
     Multi-plane lensing class
 
     The lens model deflection angles are in units of reduced deflections from the specified redshift of the lens to the
     sourde redshift of the class instance.
     """
 
-    def __init__(self, z_source, lens_model_list, lens_redshift_list, cosmo=None, numerical_alpha_class=None):
+    def __init__(self, z_source, lens_model_list, lens_redshift_list, cosmo=None, numerical_alpha_class=None,
+                 observed_convention_index = False):
         """
```
sibirrer: please add documentation of what numerical_alpha_class and observed_convention_index are.

test/test_LensModel/test_Optimizer/test_multi_plane.py
```diff
@@ -21,42 +21,51 @@ def setup(self):
...
         back_redshifts = [0.55, 0.6, 0.6, 0.74, 0.74, 0.8]
         self.back_args = [{'theta_Rs': 0.004, 'center_y': 0.1, 'center_x': 1, 'Rs': 0.13},
                           {'theta_Rs': 0.001, 'center_y': 0.2, 'center_x': 0.7, 'Rs': 0.11},
                           {'theta_Rs': 0.003, 'center_y': -0.1, 'center_x': -1, 'Rs': 0.13},
                           {'theta_Rs': 0.0008, 'center_y': 0.42, 'center_x': 0.1, 'Rs': 0.13},
-                          {'theta_Rs': 0.0014, 'center_y': 0.42, 'center_x': 1.08, 'Rs': 0.13},
+                          {'theta_E': 0.3, 'center_y': 0.7, 'center_x': 1.08},
                           {'theta_Rs': 0.006, 'center_y': 0.5, 'center_x': 0.75, 'Rs': 0.16}]
 
+        self.convention_idx = [14,18]
```
sibirrer: can you make a test when you switch those two indexes and demand that the answer is the same?

test/test_LensModel/test_multi_plane.py
```diff
@@ -228,6 +229,56 @@ def test_pseudo_multiplane(self):
...
+        z_source = 1.5
+        lens_model_list = ['SIS', 'SIS','SIS', 'SIS']
+        redshift_list = [0.5, 0.5, 0.8, 0.7]
+
+        kwargs_lens = [{'theta_E': 1, 'center_x':0, 'center_y': 0},
+                       {'theta_E': 0.4, 'center_x': 0, 'center_y': 0.2},
+                       {'theta_E': 1, 'center_x': 1.8, 'center_y': -0.4},
+                       {'theta_E': 0.31, 'center_x': 1., 'center_y': 0.7}]
+
+        index_list = [[1], [1,2]]
```
sibirrer: can you add a test where you switch the convention and demand them to be the same result? [2, 1] vs [1, 2]

sibirrer:
Hi Daniel, thanks for the update. I am a bit peaky here, sorry (and maybe more demanding than when I write my own code) but I think this will improve the code in the long run. About multiple changing positions at different lens planes: I don't think the current code works correctly as the ray tracing of the second displaced deflector does not take the displacement of the first deflector into account. At least this is what I read out of the code.

dangilman:
Hi, I think I addressed everything, nice catch on the bug with computing the kwargs_lens for a list of models

PR #91, opened 2019-06-18
Nested Sampling algorithms in Fitting Sequence

aymgal:
Hi Simon,

Here is my contribution for adding nested samplers **MultiNest**, **DyPolyChord** and **Dynesty** to the Fitting Sequence.

I wrote unit tests, hopefully in a correct manner. They all run on my computer, but I expect those for MultiNest and DyPolyChord won't run online : they both require compiled Fortran code. How do you want to proceed ? Dynesty is pure python, should be ok.

Side change : since I was not able to exclude some specific commits from this PR, a small fix in the computation of deflection angles from the convergence is also included : [https://github.com/sibirrer/lenstronomy/blob/master/lenstronomy/LensModel/numerical_profile_integrals.py](url)


ajshajib:
Nested sampling is a great addition. Do you have any stat on the difference in runtime between dynesty and emcee/cosmohammer to sample the same number of points?

aymgal:
> Nested sampling is a great addition. Do you have any stat on the difference in runtime between dynesty and emcee/cosmohammer to sample the same number of points?

Capture d’écran 2019-06-18 à 11 00 44

For now I only did some quick tests on a very simple lens (no point sources nor cosmo). Here is are some results (5 sampled parameters) : 
- PSO+MCMC (cosmohammer) : **~68 s** for 5000 points
PSO_MCMC

- MultiNest : **~204 s**, for 5440 points
MN

lenstronomy/Sampling/likelihood.py
```diff
@@ -1,5 +1,7 @@
 __author__ = 'sibirrer'
 
+import numpy as np
```
sibirrer: do you need numpy here imported when having the lines out commented?

lenstronomy/Util/sampling_util.py
```diff
@@ -0,0 +1,51 @@
...
+from scipy import special
+
+
+# transform the unit hypercube to pysical parameters for sampling
+
+
+SQRT2 = np.sqrt(2)
+
+
+def unit2gaussian(x, mu, sigma):
```
sibirrer: can you describe a bit more what this function does?

lenstronomy/Util/sampling_util.py
```diff
@@ -0,0 +1,51 @@
...
+        cube_ = cube
+        cube = np.zeros_like(cube_)
+    for i in range(num_dims):
+        val = cube_[i] if copy else cube[i]
+        low, upp = lowers[i], uppers[i]
+        cube[i] = unit2uniform(val, low, upp)
+    return cube
+
+
+def cube2args_gaussian(cube, lowers, uppers, means, sigmas, num_dims, copy=False):
```
sibirrer: can you describe a bit more what this function does?

lenstronomy/Util/sampling_util.py
```diff
@@ -0,0 +1,51 @@
...
+    """from Handley+15 (PolyChord paper)"""
+    return vmin + (vmax - vmin) * x
+
+
+def uniform2unit(theta, vmin, vmax):
+    """from Handley+15 (PolyChord paper)"""
+    return (theta - vmin) / (vmax - vmin)
+
+
+def cube2args_uniform(cube, lowers, uppers, num_dims, copy=False):
```
sibirrer: can you describe a bit more what this function does?

sibirrer:
Hi Aymeric, thanks a lot for this! This is great! The only thing is that the FromFile class might be split into two more specific tasks.
Simon

lenstronomy/Sampling/likelihood.py
aymgal: Indeed, will be removed

lenstronomy/Util/sampling_util.py
aymgal: is that ok ?

    """
    mapping from a uniform distribution on unit hypercube
    to gaussian distribution on parameter space, with mean 'mu' and std dev 'sigma'

    from Handley+15, eq. (A9)
    """

lenstronomy/Util/sampling_util.py
aymgal: Is that ok?

"""
    mapping from uniform distribution on unit hypercube
    to gaussian distribution on parameter space, with mean 'mu' and std dev 'sigma'

    from Handley+15, eq. (A9)
    """

lenstronomy/Util/sampling_util.py
aymgal: Is that ok ?

```suggestion
def cube2args_uniform(cube, lowers, uppers, num_dims, copy=False):
   """
    mapping from uniform distribution on unit hypercube 'cube'
    to uniform distribution on parameter space

    :param cube: list or 1D-array of parameters on unit hypercube
    :param lowers: lower bounds for each parameter
    :param uppers: upper bounds for each parameter
    :param num_dims: parameter space dimension (= number of parameters)
    :param copy: If False, this function modifies 'cube' in-place. Default to False.
    :param return: hypercube mapped to parameters space
    """
```

lenstronomy/Util/sampling_util.py
aymgal: Is that ok ?

```suggestion
def cube2args_gaussian(cube, lowers, uppers, means, sigmas, num_dims, copy=False):
    """
    mapping from uniform distribution on unit hypercube 'cube'
    to truncated gaussian distribution on parameter space, 
    with mean 'mu' and std dev 'sigma'

    :param cube: list or 1D-array of parameters on unit hypercube
    :param lowers: lower bounds for each parameter
    :param uppers: upper bounds for each parameter
    :param means: gaussian mean for each parameter
    :param sigmas: gaussian std deviation for each parameter
    :param num_dims: parameter space dimension (= number of parameters)
    :param copy: If False, this function modifies 'cube' in-place. Default to False.
    :param return: hypercube mapped to parameters space
    """
```

aymgal:
> Hi Aymeric, thanks a lot for this! This is great! The only thing is that the FromFile class might be split into two more specific tasks.
> Simon



> Hi Aymeric, thanks a lot for this! This is great! The only thing is that the FromFile class might be split into two more specific tasks.
> Simon

Actually I coded 'FromFile' when your 'INTERPOL' was not yet implemented in lenstronomy. Normally I deleted that file in one of the commits... so that's strange, do you see it still included in the last commit ?

sibirrer:
> > Hi Aymeric, thanks a lot for this! This is great! The only thing is that the FromFile class might be split into two more specific tasks.
> > Simon
> 
> Actually I coded 'FromFile' when your 'INTERPOL' was not yet implemented in lenstronomy. Normally I deleted that file in one of the commits... so that's strange, do you see it still included in the last commit ?



> > Hi Aymeric, thanks a lot for this! This is great! The only thing is that the FromFile class might be split into two more specific tasks.
> > Simon
> 
> Actually I coded 'FromFile' when your 'INTERPOL' was not yet implemented in lenstronomy. Normally I deleted that file in one of the commits... so that's strange, do you see it still included in the last commit ?

I see. Yes, I see it in the latest commit. So maybe it did not get properly deleted in the git branch that you are pushing.

lenstronomy/Util/sampling_util.py
sibirrer: thanks!

lenstronomy/Util/sampling_util.py
sibirrer: thanks!

aymgal:
I just committed the changes + additional doc in Fitting Sequence

sibirrer:
Awesome, thank you! Let's see whether test are passing...

PR #95, opened 2019-06-19
Further improvements in test coverage

aymgal:
Ok now I think the maximum number of lines are tested (what's left is almost only those that required packages to be installed properly)


sibirrer:
Thank you very much!

PR #109, opened 2019-07-05
Feature/nfw gaussian decomposition

ajshajib:
Add Gauss decomposed elliptical NFW (2D) profile and Gauss decomposed spherical cored-truncated NFW (3D to 2D) profile as lens models.


sibirrer:
Thank you!

PR #111, opened 2019-07-06
add ray_shooting_partial_steps to multi_plane

dangilman:
the function was in multi_plane_base, but not in the main multi_plane class where things get called by the user


sibirrer:
Can you just quickly write a test function for this? Thanks!

dangilman:
added the test function

sibirrer:
Thank you!

PR #115, opened 2019-07-10
Bugfix/mpi polychord sampler

ajshajib:
This PR fixes the following issues:

- Check if coordinate array is empty to prevent errors in fastell.
- Fix competing file handling at post-processing in PolyChord sampler with MPI.

Additionally, I added a few lines to catch exceptions in likelihood computation and print the `param_array` along with the usual error message. This is mainly for debugging purpose and can be reverted later if not necessary to keep.


lenstronomy/Plots/output_plots.py
```diff
@@ -734,7 +734,7 @@ def source_plot(self, ax, numPix, deltaPix_source, v_min=None,
         text_description(ax, d_s, text=text, color="w", backgroundcolor='k',
                          flipped=False, font_size=font_size)
         source_position_plot(ax, coords_source, self._kwargs_source_partial)
-        return ax
+        return ax, source
```
sibirrer: I am thinking of another way that you can receive the 'source'. In this way, it breaks all the current notebooks and plotting scripts. Maybe you can write a specific function that just returns 'source' and then call this function to create the plots?

lenstronomy/Sampling/likelihood.py
```diff
@@ -209,7 +209,15 @@ def __call__(self, a):
...
 
     def likelihood(self, a):
-        return self.logL(a)
+        try:
+            return self.logL(a)
+        except:
+            print('The parameter array during error: {}'.format(a))
+            from mpi4py import MPI
+            print('Worker: {}'.format(MPI.COMM_WORLD.Get_rank()))
+            logL = self.logL(a)
```
sibirrer: I am not sure this works. You call here the function that was excepted before. Please add a test to this behavior

lenstronomy/LensModel/Profiles/spemd_smooth.py
```diff
@@ -145,4 +147,19 @@ def convert_params(self, theta_E, gamma, q):
         gam = (gamma-1)/2.
         q_fastell = (3-gamma)/2. * (theta_E ** 2 / q) ** gam
 
-        return q_fastell, gam
\ No newline at end of file
+        return q_fastell, gam
+
+    @staticmethod
```
sibirrer: please add a test for this definition

lenstronomy/Plots/output_plots.py
ajshajib: This sneaked in from a hack I did while plotting. I will revert it.

lenstronomy/Sampling/likelihood.py
ajshajib: The intended behavior is print the parameter array that causes an error, but still print out the full error message, hence the `logL()` function is being called anyway to throw the error. This is for debugging purpose to see what values are in the parameter array. Again, this is a hack, I can revert it, if undesirable.

lenstronomy/LensModel/Profiles/spemd_smooth.py
ajshajib: Will do.

PR #118, opened 2019-07-26
Feature/polychord settings flexibility

ajshajib:
- Let users pass all the PolyChord settings from the top level.
- Randomize the computed R_E by `LensModelExtensions.effective_einstein_radius()` uniformly in (R-division, R] so that the computed distribution of R_E does not look like stripes on a histogram or 2D distribution plot.
- Remove a typo from a label string in `output_plots.py`.


lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -234,7 +234,7 @@ def effective_einstein_radius(self, kwargs_lens_list, k=None, spacing=1000):
             if sum_mask > 0:
                 kappa_mean = np.sum(kappa*mask)/np.sum(mask)
                 if kappa_mean < 1:
-                    return r
+                    return -np.random.uniform(-r, -r+r_array[1]-r_array[0])
```
sibirrer: I don't really understand why you need to draw a random number here. It's clear that within the grinding, the precision is limited but there should not be a random component of the output

lenstronomy/Sampling/Samplers/polychord_sampler.py
```diff
@@ -65,11 +67,10 @@ def __init__(self, likelihood_module, prior_type='uniform',
             os.mkdir(self._output_dir)
 
         self._output_basename = output_basename
-        self.settings = {
-            'file_root': self._output_basename,
-            'base_dir': self._output_dir,
-            'seed': seed_increment,
-        }
+        self._seed_increment = seed_increment
+        self.settings = polychord_settings
```
sibirrer: does self.settings have to be a public variable? Otherwise make it private

lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -219,6 +219,8 @@ def nested_sampling(self, sampler_type='MULTINEST', kwargs_run={},
                         prior_type='uniform', width_scale=1, sigma_scale=1, 
                         output_basename='chain', remove_output_dir=True, 
                         dypolychord_dynamic_goal=0.8,
+                        polychord_settings = {},
```
sibirrer: no space between polychord_setting={}

lenstronomy/LensModel/lens_model_extensions.py
ajshajib: With the current code, the theta_E PDF ends up like this.

Screen Shot 2019-07-26 at 12 26 51 PM

ajshajib: The random sampling between the two stripes to have points populated there. Technically, with the current algorithm you only know theta_E to be within r[n-1] and r[n], not exactly at r[n], n being the last iteration.

sibirrer: In terms of software, I'd prefer not to mix two different tasks. Better would be to return a second argument, which is the precision of the return statement (step size in this case) and then your specific solution to draw/convolve the samples with the precision statement is then a second different step. Would that be ok for you?

ajshajib: Certainly. I will push a new commit making the necessary change.

ajshajib: Do you want me to add an extra argument to the function call, so that the precision in only returned when asked for not to break any previous codes/notebooks?

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -234,7 +236,10 @@ def effective_einstein_radius(self, kwargs_lens_list, k=None, spacing=1000):
             if sum_mask > 0:
                 kappa_mean = np.sum(kappa*mask)/np.sum(mask)
                 if kappa_mean < 1:
-                    return r
+                    if get_precision:
```
sibirrer: I like that! And now I am getting a bit picky: Can you write a test function for it? Sorry and thank you!

sibirrer:
I looked at the tests and they fail due to some different (new?) configurations of travis. I changed them yesterday for the master and simulation_development branches.
https://travis-ci.org/sibirrer/lenstronomy/jobs/564292893

I pushed a new master branch that should fix this. So if you merge yours first then it should pass.

PR #120, opened 2019-08-12
Add resuming option for dypolychord with MPI

ajshajib:
- Add ability to resume Polychord sampling.
- Refactor Polychord settings to pass. Use `kwargs_run` to pass keyword arguments to dyPolyChord and `polychord_settings` to pass keyword arguments to Polychord.

### DIFF
--- a/lenstronomy/Workflow/fitting_sequence.py
+++ b/lenstronomy/Workflow/fitting_sequence.py
@@ -256,6 +256,11 @@ class FittingSequence(object):
             samples, means, logZ, logZ_err, logL, results_object = sampler.run(kwargs_run)
 
         elif sampler_type == 'DYPOLYCHORD':
+            if 'resume_dyn_run' in kwargs_run and \
+                    kwargs_run['resume_dyn_run'] is True:
+                resume_dyn_run = True
+            else:
+                resume_dyn_run = False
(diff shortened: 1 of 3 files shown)


lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -256,6 +256,11 @@ def nested_sampling(self, sampler_type='MULTINEST', kwargs_run={},
             samples, means, logZ, logZ_err, logL, results_object = sampler.run(kwargs_run)
 
         elif sampler_type == 'DYPOLYCHORD':
+            if 'resume_dyn_run' in kwargs_run and \
```
sibirrer: is this line being tested? I am just not sure as coveralls seems to be down

test/test_Workflow/test_fitting_sequence.py
```diff
@@ -223,6 +223,13 @@ def test_fitting_sequence(self):
             'kwargs_run': {
                 'ninit': 8,
                 'nlive_const': 10,
+                #'seed_increment': 1,
```
sibirrer: do you want to add a short test that re-uses a very simple chain as intended with resume_dyne_run? Then probably all the lines of code get tested too.

PR #129, opened 2019-09-13
Fix astrometric correction

aymgal:


sibirrer:
Thanks, did not spot nor test that!

PR #139, opened 2019-12-09
Sparse optimization with SLITronomy (deprecated)

aymgal:
This PR aims to add preliminary support for pixelated surface brightness reconstruction, through sparse optimization. This is a development in the continuation of the SLIT method (https://github.com/herjy/SLIT), and is implemented as a separate package that aims to work transparently with lenstronomy : https://github.com/aymgal/SLITronomy .
 
While the core computation is externalised to SLITronomy, the following changes has been introduced in lenstronomy :
- a new light profile : **Starlets**. It implements decomposition and reconstruction in the starlet basis. A python implementation of the starlet transform is included, but faster transforms are achievable when PySAP is installed (https://github.com/CEA-COSMIC/pysap).
- a new "multi-band-type" : `'single-band-sparse'`. Based on the original `'single-band'`, it uses the new class `SingleBandMultiModelSparse` for support of pixelated sparse light distribution like Starlets. The class_creator utility has been updated on purpose.
- a new `ImageSparseFit` class, in the same spirit of `ImageLinearSolve`, importing the SLITronomy solver to reconstruct the source light distribution (lens light not supported yet, will be soon).
- adaptations of `LikelihoodModule` for support of all the above. Currently, `kwargs_sparse_solver` is a new entry in `kwargs_likelihood` to pass settings to the SLITronomy solver.

@sibirrer : my main wonders concern the implementation in the class_creator function, as a new `SingleBandMultiModelSparse `, whose name is not fully justified in the current state, I think... Do you have a better suggestion ? But I would like to have your general feedback about the other aspects to make it suitable for a merge 🙂 

Also, I know that I have to implement test routines for all of that before merging, but I just would like to make sure the implementation is ok before going on.


lenstronomy/ImSim/MultiBand/single_band_multi_model.py
```diff
@@ -167,3 +168,128 @@ def select_kwargs(self, kwargs_lens=None, kwargs_source=None, kwargs_lens_light=
         else:
             kwargs_extinction_i = [kwargs_extinction[k] for k in self._index_optical_depth]
         return kwargs_lens_i, kwargs_source_i, kwargs_lens_light_i, kwargs_ps_i, kwargs_extinction_i
+
+
+class SingleBandMultiModelSparse(ImageSparseFit):
```
sibirrer: Is the only difference of this class that it inherits ImageSparseFit rather then ImageLinearFit? I see a lot of duplication here that we might make more general.

lenstronomy/ImSim/image_linear_solve.py
```diff
@@ -224,6 +224,22 @@ def _linear_response_matrix(self, kwargs_lens, kwargs_source, kwargs_lens_light,
             n += 1
         return np.nan_to_num(A)
 
+    def _source_linear_response_matrix(self, x_grid, y_grid, kwargs_lens, kwargs_source):
```
sibirrer: this is not really the full linear response matrix as it does not do the convolution. It 'just' splits the intrinsic surface brightness in the different linear components, but not necessary in the data vector (could also be in supersampled resolution)

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,290 @@
+import numpy as np
+
+from lenstronomy.ImSim.image_model import ImageModel
+from lenstronomy.ImSim.Numerics.convolution import PixelKernelConvolution
+from lenstronomy.Util import util
+from lenstronomy.Util import image_util
+
+from slitronomy.Optimization.sparse_solver import SparseSolverSource
```
sibirrer: one thing you have to be aware of is that there can not be looped dependencies. E.g. if you have SLITronomy being dependent on lenstronomy in the installation and then lenstronomy being dependent on SLITronomy, this is not going to work well.

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,290 @@
...
+        using sparse optimization, on the data pixelated grid.
+
+        :param kwargs_lens: list of keyword arguments corresponding to the superposition of different lens profiles
+        :param kwargs_source: list of keyword arguments corresponding to the superposition of different source light profiles
+        :param kwargs_lens_light: list of keyword arguments corresponding to different lens light surface brightness profiles
+        :param kwargs_ps: keyword arguments corresponding to "other" parameters, such as external shear and point source image positions
+        :return: 1d array of surface brightness pixels of the optimal solution of the linear parameters to match the data
+        """
+        C_D_response, model_error = self._error_response(kwargs_lens, kwargs_ps, kwargs_special=kwargs_special)
+        model, param, n_pixels_source = self._solve(kwargs_lens, kwargs_source, kwargs_lens_light)
```
sibirrer: effectively the only conceptual difference between SLIT and the default lenstronomy solver is this line of code. In that sense I am thinking of whether a flag in the input statement in the current classes might be sufficient IF the numerics to compute _error_response can be put into the lower level numerical modules instead

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,290 @@
...
+        computes the image (lens and source surface brightness with a given lens model)
+        using sparse optimization, on the data pixelated grid.
+
+        :param kwargs_lens: list of keyword arguments corresponding to the superposition of different lens profiles
+        :param kwargs_source: list of keyword arguments corresponding to the superposition of different source light profiles
+        :param kwargs_lens_light: list of keyword arguments corresponding to different lens light surface brightness profiles
+        :param kwargs_ps: keyword arguments corresponding to "other" parameters, such as external shear and point source image positions
+        :return: 1d array of surface brightness pixels of the optimal solution of the linear parameters to match the data
+        """
+        C_D_response, model_error = self._error_response(kwargs_lens, kwargs_ps, kwargs_special=kwargs_special)
```
sibirrer: is C_D_respnse being used or in general this function? I see that you only use the model_error part here.
I think one of the key questions is how much of the actual numerical calculation for SLIT on the pixel level can actually be expressed jointly with the current Numerics module as an extension of this. Currently we are mixing a bit algorithm and numerics which makes the design difficult. The SLIT algorithm basically also needs a response basis per wavelet and maybe we can spend some work in generalizing the response basis return function and then worry about the numerics in the Numerics module.

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+    def num_param_linear(self, kwargs_lens, kwargs_source, kwargs_lens_light, kwargs_ps):
+        """
+
+        :return: number of linear coefficients to be solved for in the linear inversion
+        """
+        return self._num_param_linear(kwargs_lens, kwargs_source, kwargs_lens_light, kwargs_ps)
+
+    def _num_param_linear(self, kwargs_lens, kwargs_source, kwargs_lens_light, kwargs_ps):
```
sibirrer: do you need this here or is this a duplication?

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,290 @@
...
+
+        :param param: linear parameter vector corresponding to the response matrix
+        :return: updated list of kwargs with linear parameter values
+        """
+        # update amplitudes (wavelets coefficients)
+        i = 0
+        kwargs_source, i = self.SourceModel.update_linear(param, i, kwargs_list=kwargs_source)
+        kwargs_lens_light, i = self.LensLightModel.update_linear(param, i, kwargs_list=kwargs_lens_light)
+        # update other parameters if required
+        if n_pixels_source is not None:
```
sibirrer: can this be part of SourceModel.update_linear()? Then you would not need this function here.

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,290 @@
...
+            kwargs_source[0]['n_pixels'] = n_pixels_source
+        return kwargs_source, kwargs_lens_light
+
+    def reduced_residuals(self, model, error_map=0):
```
sibirrer: duplication?

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+
+        :param model:
+        :return:
+        """
+        mask = self.likelihood_mask
+        residual = (model - self.Data.data)/np.sqrt(self.Data.C_D+np.abs(error_map))*mask
+        return residual
+
+    def reduced_chi2(self, model, error_map=0):
```
sibirrer: duplication?

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,290 @@
...
+        returns reduced chi2
+        :param model:
+        :param error_map:
+        :return:
+        """
+        chi2 = self.reduced_residuals(model, error_map)
+        return np.sum(chi2**2) / self.num_data_evaluate
+
+    @property
+    def num_data_evaluate(self):
```
sibirrer: duplication?

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        number of data points to be used in the linear solver
+        :return:
+        """
+        return int(np.sum(self.likelihood_mask))
+
+
+    def image2array_masked(self, image):
```
sibirrer: these functions exist in the Numerics module and I don't think the user needs to interact with them at this high level of this class

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        returns 1d array of values in image that are not masked out for the likelihood computation/linear minimization
+        :param image: 2d numpy array of full image
+        :return: 1d array
+        """
+        array = util.image2array(image)
+        return array[self._mask1d]
+
+
+    def array_masked2image(self, array):
```
sibirrer: these functions exist in the Numerics module and I don't think the user needs to interact with them at this high level of this class

lenstronomy/LightModel/Profiles/starlets.py
```diff
@@ -0,0 +1,127 @@
...
+        operator = lambda x: self.decomposition(x, n_scales)
+        inverse_operator = lambda c: self.function(c, n_scales)
+        return util.spectral_norm(num_pix, operator, inverse_operator, num_iter=num_iter, tol=tol)
+
+
+    def _load_pysap(self):
+        if _force_no_pysap:
+            return False, None
+        try:
+            import pysap
```
sibirrer: you may add pysap in the lenstronomy requirements

lenstronomy/LightModel/Profiles/starlets.py
```diff
@@ -0,0 +1,127 @@
...
+import numpy as np
+
+from lenstronomy.Util import util
+from lenstronomy.LightModel.Profiles import starlets_slit
+
+
+_force_no_pysap = False
+
+
+class Starlets(object):
```
sibirrer: I know I don't do it well neither but some additional description and paper might be useful to explain this class

lenstronomy/LightModel/Profiles/starlets.py
```diff
@@ -0,0 +1,127 @@
...
+    def __init__(self, thread_count=1, fast_inverse=True, second_gen=False, show_pysap_plots=False):
+        self.use_pysap, pysap = self._load_pysap()
+        if self.use_pysap:
+            self._transf_class = pysap.load_transform('BsplineWaveletTransformATrousAlgorithm')
+        self._thread_count = thread_count
+        self._fast_inverse = fast_inverse
+        self._second_gen = second_gen
+        self._show_pysap_plots = show_pysap_plots
+
+    def function(self, coeffs, n_scales, n_pixels):
```
sibirrer: do you also want to provide option of where to center the pixel grid?

lenstronomy/LightModel/light_model_base.py
```diff
@@ -88,6 +88,9 @@ def __init__(self, light_model_list, smoothing=0.0000001):
             elif profile_type == 'INTERPOL':
                 from lenstronomy.LightModel.Profiles.interpolation import Interpol
                 self.func_list.append(Interpol())
+            elif profile_type == 'STARLETS':
+                from lenstronomy.LightModel.Profiles.starlets import Starlets
+                self.func_list.append(Starlets(fast_inverse=True, second_gen=False))
```
sibirrer: here we have to think of a design that allows the user to conveniently chose the class init options from the API

lenstronomy/LightModel/light_model_base.py
```diff
@@ -100,12 +103,21 @@ def surface_brightness(self, x, y, kwargs_list, k=None):
...
+            # flux not initialised here because shape depends on profile parameters
+        else:
+            pixelated_light = False
+            flux = np.zeros_like(x)
         bool_list = self._bool_list(k=k)
         for i, func in enumerate(self.func_list):
             if bool_list[i] is True:
-                out = np.array(func.function(x, y, **kwargs_list_standard[i]), dtype=float)
-                flux += out
+                if pixelated_light:
```
sibirrer: here we have to think how pixelated and continues surface brightness can co-exist. In this function there is some ambiguity what happens if 'STARLETS' is one of the options in other profiles.

lenstronomy/Sampling/Likelihoods/image_likelihood.py
```diff
@@ -25,7 +26,7 @@ def __init__(self, multi_band_list, multi_band_type, kwargs_model, bands_compute
         """
 
         self.imSim = class_creator.create_im_sim(multi_band_list, multi_band_type, kwargs_model, bands_compute=bands_compute,
-                                   likelihood_mask_list=likelihood_mask_list, band_index=0)
+                                   likelihood_mask_list=likelihood_mask_list, band_index=0, kwargs_sparse_solver=kwargs_sparse_solver)
```
sibirrer: maybe we can think of having kwargs_sparse_solver as part of kwargs_model

sibirrer:
Hi Aymeric, I see your updated PR - thanks for doing this. Let me know when I should have another look at it and/or if you require some changes more on the base level of lenstronomy!

aymgal:
Yes I updated and improved a few things. I'm still working on the SLITronomy part to make sure it's stable enough before finalising this PR, with full tests and so on.
Is also answered / reacted to a subset of your comments above. But maybe you don't see those in the github interface ?

sibirrer:
ah, I see that some conversations are resolved but they do not get on my radar. If you put something the comments here I will get notified :)

aymgal:
Yes actually I hesitated to implement it this way. However I think this offset is directly linked to the pixelated lensing operator itself, as opposed to the light profile. E.g. one could plot the wavelets coefficients with specific center_x/center_y, independently of the source grid offset. In other words, the offset due to a pixelated image-to-source mapping is independent of the 'physical' position on the coordinates grid. Not sure if I'm clear though.

One could also imagine to keep center_x / center_y and add 2 other offsets parameters to the source profile. 

Or if you think that there is no ambiguity to use center_x / center_y as grid offsets, I can adapt the implementation as you suggest.

sibirrer:
Ah I see, these are not centers of the starlets but the pixel grid numerics. Make sure that wherever the parameters appear that the documentation is clear of what they mean (I haven't checked whether it's the case).

sibirrer:
Hi Aymeric, let me know at what stage you want a PR review/approval or have question on the design level!

aymgal:
Thanks! I am currently completing test routines on the SLITronomy side, I would like this to be finished before the merge.
I am having quite hard time with tests for plotting routines, with travis builds not working anymore because of some backend issues.

sibirrer:
you can look into test/test_Plots to see how I dealt with it 
matplotlib.use('agg')
and after each plot rather than plt.show() you need plt.close()

aymgal:
Yes I had a look and followed you on these details.
But it seems to be more subtle, the package that performs wavelets transform does not install correctly anymore on the 3.6 build, still working on 2.7 though. Do you know a way to test the travis build locally on your machine? This would prevent a bunch of try-and-error commits...

sibirrer:
I see. tox might help you there, see e.g. https://opensource.com/article/19/5/python-tox

aymgal:
Thanks, very practical package indeed

aymgal:
Hi @sibirrer 
I made a few changes in the ImageSparseFit. This allows me to use the WLS engine for finding point source amplitudes, while keeping the source/lens light in a starlets representation.
As a first implementation, I added in SLITronomy a sovler that alternates between a starlet reconstruction of the source light and a linear minimization of point sources amplitudes (knowing the PS positions).
Please tell me what you think about this.

sibirrer:
Hi @aymgal, test coverage review I can only start when the tests are passing. Maybe it has to do with the scipy (or some other dependency) environment? For the other bits, I start reviewing now. Thanks!

lenstronomy/LightModel/Profiles/starlets.py
```diff
@@ -0,0 +1,177 @@
...
+    def _coeffs2pysap(self, coeffs):
+        """convert coefficients stored in numpy array to list required by pySAP"""
+        coeffs_list = []
+        for i in range(coeffs.shape[0]):
+            coeffs_list.append(coeffs[i, :, :])
+        return coeffs_list
+
+    def _load_pysap(self):
+        """load pySAP module"""
+        if _force_no_pysap:
```
sibirrer: is this a hack? So this function in the current mode never loads PySap, correct?

lenstronomy/LightModel/Profiles/starlets_slit.py
```diff
@@ -0,0 +1,116 @@
+# starlet transform from SLIT
+
+import numpy as np
+import scipy.signal as scs
+import scipy.ndimage.filters as scf
+
+
+def transform(img, n_scales, Filter='Bspline', convol2d=0, second_gen=False):
+
```
sibirrer: needs documentation

lenstronomy/LightModel/Profiles/starlets_slit.py
```diff
@@ -0,0 +1,116 @@
...
+
+        c = cnew
+     
+    wave[i+1, :, :] = c
+
+    return wave
+
+
+def inverse_transform(wave, convol2d=0, fast=True, second_gen=False):
```
sibirrer: needs documentation

lenstronomy/LightModel/light_param.py
```diff
@@ -63,6 +63,15 @@ def getParams(self, args, i):
                             raise ValueError('sigma needs to be fixed in %s.' % model)
                         kwargs['amp'] = args[i:i + num_param]
                         i += num_param
+                    elif model in ['STARLETS'] and name == 'coeffs':
+                        if 'n_scales' in kwargs_fixed and 'n_pixels' in kwargs_fixed:
+                            n_scales = kwargs_fixed['n_scales']
```
sibirrer: do I understand correctly that those variables get overwritten each time the solver is applied? That's at least what I read/see from the code.

lenstronomy/LightModel/light_param.py
```diff
@@ -95,6 +104,12 @@ def setParams(self, kwargs_list):
                             num_param = int((n_max + 1) * (n_max + 2) / 2)
                         for i in range(num_param):
                             args.append(kwargs[name][i])
+                    elif model in ['STARLETS'] and name == 'coeffs':
+                        n_scales = kwargs_fixed.get('n_scales', kwargs['n_scales'])
```
sibirrer: does this crash when there is no 'n_scales' in kwargs but in kwargs_fixed?

lenstronomy/LightModel/linear_basis.py
```diff
@@ -157,4 +170,7 @@ def add_fixed_linear(self, kwargs_fixed_list):
             if 'amp' in param_names:
                 if 'amp' not in kwargs_fixed:
                     kwargs_fixed['amp'] = 1
+            if 'coeffs' in param_names:
```
sibirrer: this can be dangerous if some other models have 'coeffs' that do not need to behave like this.

lenstronomy/Plots/plot_util.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np
 import math
+import matplotlib.pyplot as plt
```
sibirrer: do you need this import statement?

requirements.txt
```diff
@@ -10,8 +10,10 @@ numba>=0.43.1
...
 dynesty
 pymultinest
-pypolychord
+#pypolychord  # only compatible with python 2.7 ...
 nestcheck
 schwimmbad
 multiprocess>=0.70.8
-#-e git://github.com/sibirrer/fastell4py.git#egg=fastell4py
\ No newline at end of file
+-e git://github.com/aymgal/pysap.git@dev-aym#egg=pysap
```
sibirrer: is there a stable pip version? I consider this dependency not stable to be merged into the master branch. For the slit_dev branch, this is totally fine.

requirements.txt
```diff
@@ -10,8 +10,10 @@ numba>=0.43.1
...
+-e git://github.com/aymgal/slitronomy.git#egg=slitronomy
```
sibirrer: is there a stable pip version? I consider this dependency not stable to be merged into the master branch. For the slit_dev branch, this is totally fine.

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,281 @@
...
+        # TODO : make ImageModel.source_surface_brightness() to work without this overwriting.
+
+        :param kwargs_lens_light: list of keyword arguments corresponding to different lens light surface brightness profiles
+        :param unconvolved: not defined here. Here for keeping same method signatures as in super class.
+        :return: 1d array of surface brightness pixels
+        """
+        if unconvolved is True:
+            print("Warning : sparse solver for lens light does not perform deconvolution of lens light, returning convolved estimate instead")
+        # ra_grid, dec_grid = self.ImageNumerics.coordinates_evaluate
+        ra_grid, dec_grid = self.sparseSolver.lensingOperator.imagePlane.grid()
```
sibirrer: I am a bit confused why you need this line instead of 'simply' the pixel coordinates to evaluate the surface brightness (even for the STARLET basis). Or is this what this is? Then it's a bit a duplication from Data() class which is part of ImageModel

lenstronomy/ImSim/image_sparse_solve.py
```diff
@@ -0,0 +1,281 @@
...
+            if not np.all(self.PSF.psf_error_map == 0):
+                print("WARNING : SparseSolver with point sources does not support PSF error map for now !")
+            self.sparseSolver = SparseSolverSourcePS(self.Data, self.LensModel, self.SourceModel, self.ImageNumerics, 
+                                                     point_source_linear_solver=self._image_linear_solve_point_sources, #TODO: not fully satisfying
+                                                     likelihood_mask=likelihood_mask, 
+                                                     **kwargs_sparse_solver)
+            
+        self._subgrid_res_source = kwargs_sparse_solver.get('subgrid_res_source', 1)
+
+    def source_surface_brightness(self, kwargs_source, kwargs_lens=None, kwargs_extinction=None, kwargs_special=None,
```
sibirrer: I think with this routine you fight the most as it is conceptually rather different than a forward model linear inversion. I like the #TODO remarks thought.

aymgal:
Thanks a lot @sibirrer for this very detailed feedback! I will go through and address your remarks.

lenstronomy/ImSim/MultiBand/single_band_multi_model.py
```diff
@@ -167,3 +168,130 @@ def select_kwargs(self, kwargs_lens=None, kwargs_source=None, kwargs_lens_light=
...
```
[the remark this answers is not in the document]
aymgal: I created a class `ImageFit` as intermediate class, so `ImageLinearFit` and `ImageSparseFit` do not share inconsistent methods. I will implement test routines accordingly.

lenstronomy/ImSim/Numerics/convolution.py
```diff
@@ -36,6 +36,21 @@ def __init__(self, kernel, convolution_type='fft_static'):
...
+        :return: convolution kernel as an 2D image
+        """
+        return self._kernel
+
+    def copy_transpose(self):
+        """
+        
+        :return: copy of the class with kernel set to the transpose of original one
+        """
+        return PixelKernelConvolution(self._kernel.T, convolution_type=self._type)
```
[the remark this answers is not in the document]
aymgal: I renamed the method to `pixel_kernel`, as in `MultiGaussianConvolution`. This way, both 'PIXEL' and 'GAUSSIAN' are supported. The case 'NONE' is supported internally in SLITronomy, when the convolution appears to be None.

lenstronomy/ImSim/Numerics/numerics.py
```diff
@@ -136,3 +136,7 @@ def _supersampling_cut_kernel(self, kernel_super, convolution_kernel_size, super
             return kernel_cut
         else:
             return kernel_super
+
+    @property
+    def convolution_class(self):
+        return self._conv
```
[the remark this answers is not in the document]
aymgal: Well on my side it's fine if the returned object is `None`, I juste test if this the case. Would you prefer to make the `psf_type` public (or better, accessed through a property), so the user can test if the psf type is 'NONE' before trying to access the convolution class ?

lenstronomy/ImSim/image_linear_solve.py
```diff
@@ -192,8 +192,8 @@ def _linear_response_matrix(self, kwargs_lens, kwargs_source, kwargs_lens_light,
         :return:
         """
         x_grid, y_grid = self.ImageNumerics.coordinates_evaluate
-        source_light_response, n_source = self.source_mapping.image_flux_split(x_grid, y_grid, kwargs_lens,
-                                                                               kwargs_source)
+        source_light_response, n_source = self._source_linear_response_matrix(x_grid, y_grid, kwargs_lens,
```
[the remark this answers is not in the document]
aymgal: No idea why this is still here, probably deprecated. Back to previous version.

lenstronomy/ImSim/MultiBand/single_band_multi_model.py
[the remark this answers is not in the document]
aymgal: @sibirrer how would you do to "dynamically" inherit either from `ImageLinearFit` or `ImageSparseFit` ?
I see two ways : 
- the `SingleBandMultiModel` class has an attribute, e.g. `self.model`, that is one or the other class. The issue here is that it is not a direct inheritance, hence one must create aliases for a lot of attributes for the current workflow to work properly. 
- replacing the `self` by the. I'm pretty sure it's bad practice though.
- inheriting from both classes at once, but calling the right constructor in the `__init__`. Not sure it's feasible / won't lead to strange behaviours though.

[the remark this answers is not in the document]
sibirrer: @aymgal I see the issue. or maybe you can make 
- one class inherit the other and then overrule some definitions that are different? 
- create a base class with all the shared functionalities, inherits it for both classes and then only add those definitions that are not shared in the specific classes.
I am not sure it works for this specific problem thought.

lenstronomy/LightModel/light_param.py
aymgal: The `n_scales` is kept fixed (so must be in kwargs_fixed), but the `n_pixels` can indeed be updated by the solver. However both are required to know the number of starlet coefficients (`coeffs`), so the code might seem a bit odd here

lenstronomy/LightModel/light_param.py
aymgal: I'm not sure to understand. It should not crash, we either take the value from the kwargs_fixed, and if not in kwargs_fixed, take it from kwargs ?

sibirrer: if 'n_scales' is not in kwargs, then this will raise regardless of wether 'n_scales' is in kwargs_fixed (because it calls kwargs['n_scale']). Did you test this?

aymgal: Ah I see. I thought the second entry in `get()` would not be executed if the first one does not raise an error. I check and indeed everything is executed. I fixed that.

lenstronomy/LightModel/linear_basis.py
aymgal: Hmm yes indeed. I did not want to name those coefficients the `'amp'`, but in the end... it may be clearer to stay with this convention.

lenstronomy/LightModel/Profiles/starlets.py
aymgal: _force_no_pysap is False by default, so it's loading pySAP if it is installed. This is for debuging when pysap is not installed for instance.

lenstronomy/LightModel/Profiles/starlets.py
```diff
@@ -0,0 +1,177 @@
...
+        """
+        if self.use_pysap:
+            coeffs = self._transform(image, n_scales)
+        else:
+            coeffs = starlets_slit.transform(image, n_scales, second_gen=self._second_gen)
+        return coeffs
+
+    def spectral_norm(self, n_scales, n_pixels):
+        """
+        spectral norm associated to the starlet transform operator
```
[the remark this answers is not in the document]
aymgal: @sibirrer you mean more than the original paper describing the transform in the docstring ?

lenstronomy/ImSim/MultiBand/single_band_multi_model.py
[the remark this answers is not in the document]
aymgal: @sibirrer I finally created a `SingleBandMultiModel` inheriting from `ImageFit`, and which has a `self.fit` attribute which is either `ImageLinearFit` or `ImageSparseFit`.
Although it may be a bit redundant to have as an attribute a class inheriting from the same class that the class itself. I tried to figure out several ways of dealing with these architecture issues, and I after all I think that making another class `SingleBandMultiModel` for the sparse fit is not a very flexible way to proceed. It is much more generalisable to have on `SingleBandMultiModel`, one `MultiLinear` and one `JointLinear`.
What do you think about this last version ?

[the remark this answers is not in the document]
sibirrer: @aymgal hmm, there quite some difference in the in and output of the ImageLinearFit and ImageSparseFit routines such that you effectively have to modify/flag each routine which inherits from ImageFit. In that sense it might be better to have them slit at this level and create a support/base class that only contains those functions that are used (and inherited) in both options (linear and sparse inversion). From the current class, I (or anyone else) will have trouble re-designing the architecture without breaking either the linear or the sparse solver bit. That's what I most concerned of.

lenstronomy/ImSim/image_sparse_solve.py
aymgal: @sibirrer This is because the coordinates grid is handled in SLITronomy as well, so it ensures the light profile is projected on the same grid than the one used in the solver.
This may be improved in the future, with a better integration to lenstronomy.

PR #141, opened 2019-12-10
add FREEFORM lens model

dangilman:


lenstronomy/LensModel/Profiles/free_form.py
```diff
@@ -0,0 +1,41 @@
...
+
+        :param x: x position to evaluate deflection angle
+        :param y: y position to evaluate deflection angle
+        :param alpha_x: deflection at x-coordinates
+        :param alpha_y: deflection at y-coordinates
+        :param center_x: not used
+        :param center_y: not used
+        :return: deflections at image positions
+        """
+        return np.array(alpha_x), np.array(alpha_y)
```
sibirrer: I don't see the purpose of this model much as effectively you already have the answer here. The input are not really parameters but actually the solution of it for specific points,

lenstronomy/Util/util.py
```diff
@@ -541,3 +541,29 @@ def convert_bool_list(n, k=None):
...
+    :param source_y: y source position [arcsec]
+    :param xpos: x image location(s) [arcsec] (numpy array)
+    :param ypos: y image location(s) [arcsec] (numpy array)
+    :param lens_model_list: a list of other deflector models
+    :param kwargs_list: kwargs for other deflectors
+    :return: (alpha_x, alpha_y) deflections that map image positions to source position for FREEFORM lensmodel
+    """
+
+    def _func_to_minimize(alpha):
+        kwargs = [{'potential': 0, 'alpha_x': alpha[0], 'alpha_y': alpha[1]}] + kwargs_list
```
sibirrer: this is very specific that you require the 'FREEFORM' model to be the first of the list.

lenstronomy/Util/util.py
```diff
@@ -541,3 +541,29 @@ def convert_bool_list(n, k=None):
     else:
         raise ValueError('input list k as %s not compatible' % k)
     return bool_list
+
+def compute_freeform_alpha(source_x, source_y, x_image, y_image, lensModel, kwargs_list):
+
+    """
+
+    :param source_x: x source position [arcsec]
```
sibirrer: parameter description do not match input arguments

lenstronomy/Util/util.py
```diff
@@ -541,3 +541,29 @@ def convert_bool_list(n, k=None):
     else:
         raise ValueError('input list k as %s not compatible' % k)
     return bool_list
+
+def compute_freeform_alpha(source_x, source_y, x_image, y_image, lensModel, kwargs_list):
```
sibirrer: the function name is a bit misleading in my opinion. It computes a specific deflection angle to be added on top of an otherwise constrained lens model such that a certain image position matches a certain source position.

I don't think this is really an utility function as it is but rather a more 'advanced' option that may go into the lens model_extension module. Also, the lensModel instance requires to have the 'FREEFORM' model to be the first in the list.

sibirrer: I don't see why you necessarily want to apply this to the 'FREEFORM' model. I'd imaging you apply this to a parameterized simple model (e.g. constant shift, curved arc, both implemented in lenstronomy). The 'FREEFORM' model is basically ill-defined when it is not evaluated where it is intended to or effectively a constant shift, so why not just using a constant shift model and if needed higher orders adding shear and convergence or a CURVED_ARC description?

PR #143, opened 2020-01-07
Best fit values updated after MCMC

aymgal:
When running only a MCMC with no PSO beforehand, no "best fit" kwargs were set to the fitting sequence, making inaccessible the `best_fit()` method. This PR aims to fix that, by finding the best MCMC sample from logL values, and computing corresponding best fit kwargs (the `kwargs_result`).


sibirrer:
Thank you Aymeric! This makes a lot of sense to have a best fit parameter updated with the MCMC sampling.

PR #150, opened 2020-02-10
bug fix when using TDCosmography.velocity_dispersion_dimension_less with MGE 

martin-millon:
Errors arise when calling `velocity_dispersion_dimension_less`, using numerical velocity dispersion. 
The function `KinematicAPI.velocity_dispersion_numerical` actually needs `kwargs_mge_mass` and `kwargs_mge_light` and a rough estimate of the Einstein radius otherwise, it will raise an error.  

The user still needs to call  `KinematicAPI.kinematics_modeling_settings` before `KinematicAPI.velocity_dispersion_numerical` to avoid the error


sibirrer:
Thanks. I had locally already a similar fix but adopt yours. I'll add a test for the mge settings later. One note: for some kinematics calculation, theta_E and r_eff are not required to be known (and thus computed). The way you changed it is save but may for some options not be required.

PR #154, opened 2020-02-20
Convenient feature to delete INTERPOL cache

aymgal:


lenstronomy/LightModel/Profiles/interpolation.py
```diff
@@ -94,3 +94,7 @@ def coord2image_pixel(ra, dec, center_x, center_y, phi_G, scale):
         y_ = dec_ / scale
         x, y = util.rotate(x_, y_, phi_G)
         return x, y
+
+    def delete_cache(self):
+        """delete the cached interpolated image"""
+        del self._image_interp
```
sibirrer: the variable self._image_interp may not exist and this function will raise an error in this case. The definition should check whether _image_interp exists and only delete the variable in this case.

aymgal: Done

sibirrer:
Thanks!

PR #156, opened 2020-02-25
Ellipticity convention

sibirrer:
Makes ellipticity conventions more transparent, adopts same conventions for NIE as SPEMD(gamma=2)

### DIFF
--- a/lenstronomy/LensModel/Profiles/nie.py
+++ b/lenstronomy/LensModel/Profiles/nie.py
@@ -8,7 +8,8 @@ from lenstronomy.LensModel.Profiles.base_profile import LensProfileBase
 
 class NIE(LensProfileBase):
     """
-
+    Non-singular isothermal ellipsoid
+    kappa = theta_E/2 [s2IE + r2(1 − e *cos(2*phi)]−1/2
     """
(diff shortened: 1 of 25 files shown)


sibirrer:
@jiwoncpark  I updated the NIE conventions (also with better documentation). Can you check whether this works? You can check out the ellipticity_convention branch to do so. Thank you!

lenstronomy/LensModel/Profiles/nie.py
```diff
@@ -15,49 +15,58 @@ class NIE(LensProfileBase):
...
 
-        :param theta_E:
-        :param e1:
-        :param e2:
-        :return:
+        :param theta_E: Einstein radius
+        :param e1: eccentricity component
+        :param e2: eccentricity component
+        :param s_scale: smoothing scale
+        :return: b, s, q, phi_G
```
jiwoncpark: A general description of `b, s` would help here, e.g. `b` is the critical radius and `s` is the smoothing scale, both loaded with appropriate ellipticity factors. You can refer readers to the expressions after Equation 8 in Keeton and Kochanek 1998,

sibirrer: Thank you Ji Won! I made the minor addition and merge it to the master branch.

PR #158, opened 2020-02-27
Add  as

ajshajib:
Add `set_param_value` as `fitting_sequence`

### DIFF
--- a/lenstronomy/Workflow/fitting_sequence.py
+++ b/lenstronomy/Workflow/fitting_sequence.py
@@ -441,37 +441,15 @@ class FittingSequence(object):
         """
         current_params = self._updateManager.parameter_state
 
-        for item in lens:
-            index = item[0]
-            keys = item[1]
-            values = item[2]
+        for items, kwargs_key in zip([lens, source, lens_light, ps],
+            ['kwargs_lens', 'kwargs_source', 'kwargs_lens_light', 'kwargs_ps']):
+            for item in items:
+                index = item[0]
+                keys = item[1]
+                values = item[2]
 
-            for key, value in zip(keys, values):
-                current_params['kwargs_lens'][index][key] = value
-
-        for item in source:
-            index = item[0]
-            keys = item[1]
-            values = item[2]
-
-            for key, value in zip(keys, values):
-                current_params['kwargs_source'][index][key] = value
-
-        for item in lens_light:
-            index = item[0]
-            keys = item[1]
-            values = item[2]
-
-            for key, value in zip(keys, values):
-                current_params['kwargs_lens_light'][index][key] = value
-
-        for item in ps:
-            index = item[0]
-            keys = item[1]
-            values = item[2]
-
-            for key, value in zip(keys, values):
-                current_params['kwargs_ps'][index][key] = value
+                for key, value in zip(keys, values):
+                    current_params[kwargs_key][index][key] = value
 
         self.update_state(current_params)
         return 0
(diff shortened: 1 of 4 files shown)


lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -422,6 +425,57 @@ def update_settings(self, kwargs_model={}, kwargs_constraints={}, kwargs_likelih
...
+        :param lens_light: [[i_model, ['param1', 'param2',...], [...]]
+        :type lens_light:
+        :param ps: [[i_model, ['param1', 'param2',...], [...]]
+        :type ps:
+        :return: 0, the value of the param is overwritten
+        :rtype:
+        """
+        current_params = self._updateManager.parameter_state
+
+        for item in lens:
```
sibirrer: there is quite some repetition for the different model components. Maybe you can write one function that you then call several times for the different keyword argument list?

lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -422,6 +425,57 @@ def update_settings(self, kwargs_model={}, kwargs_constraints={}, kwargs_likelih
         self._updateManager.update_limits(change_source_lower_limit, change_source_upper_limit)
         return 0
 
+    def set_param_value(self, lens=[], source=[], lens_light=[], ps=[]):
```
sibirrer: This function is not tested yet

ajshajib: The added test lines would eventually call this function, so this function is "covered". Do you want a separate test function specific to this function?

sibirrer: ah, I see. It would be slightly better if the tests were in the UpdateManager test class. I know I also not always follow this but the tests should be executed at the level of the routine if possible.

ajshajib: I only added the method to `FittingSequence`, not `UpdateManger`.

sibirrer: ah, I would put it in the UpdateManager class, as it is of this kind

ajshajib: I am confused. In this method, the parameter update is executed through `FittingSequence.update_state()` method, which I see as an already established channel to talk to `UpdateManger`. I am not calling any `UpdateManger` method, only the `current_state` of the parameters are being read out from `UpdateManger.param_state()`. I am not sure if the modification within this method would directly change the the values for `UpdateManger._kwargs_temp`, but to be secure I can put a `deepcopy` around it.

sibirrer: The idea is that all the parameter sorting/adding/changing/deletions is managed in UpdateManager. So you can write the same function within UpdateManager and then have a one-liner in the FittingSequence class, like the update_state() routine. In the future I even consider inherit UpdateManager in FittingSequence. I just want to have the FittingSequence class as lightweight as possible.

ajshajib: Alright, got it now.

sibirrer:
Great, thanks! I'll remove the import copy statement after the merge

ajshajib:
Thanks, forgot about the rogue `copy` statement.  Oh hey, that was my 100th commit. :tada:

sibirrer:
Awesome, congratulations! Happy to receive more commits and PR's in the future :)

PR #163, opened 2020-03-09
add downhill simplex optimization to fitting sequence

dangilman:
It finds a local minima really quickly, and seems to speed up the fitting process running right after a relatively short PSO

### DIFF
--- a/lenstronomy/Sampling/likelihood.py
+++ b/lenstronomy/Sampling/likelihood.py
@@ -247,7 +247,7 @@ class LikelihoodModule(object):
     def __call__(self, a):
         return self.logL(a)
 
-    def negativeChi2(self, a):
+    def negativelogL(self, a):
         logL = self.logL(a)
         return -logL
 
--- a/lenstronomy/Sampling/likelihood.py
+++ b/lenstronomy/Sampling/likelihood.py
@@ -247,6 +247,10 @@ class LikelihoodModule(object):
     def __call__(self, a):
         return self.logL(a)
 
+    def negativeChi2(self, a):
+        logL = self.logL(a)
+        return -logL
+
     def likelihood(self, a):
         return self.logL(a)
 
--- a/lenstronomy/Sampling/likelihood.py
+++ b/lenstronomy/Sampling/likelihood.py
@@ -247,9 +247,5 @@ class LikelihoodModule(object):
     def __call__(self, a):
         return self.logL(a)
 
-    def negativelogL(self, a):
-        logL = self.logL(a)
-        return -logL
-
     def likelihood(self, a):
         return self.logL(a)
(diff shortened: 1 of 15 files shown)


lenstronomy/Sampling/sampler.py
```diff
@@ -26,8 +27,35 @@ def __init__(self, likelihoodModule):
...
+    def simplex(self, init_pos, n_iterations, method, print_key='SIMPLEX'):
+        """
+        returns the best fit for the lens model using the optimization routine specified by method
+        """
+        print('Performing the optimization using algorithm:', method)
+        time_start = time.time()
+        result = minimize(self.chain.negativelogL, x0=init_pos, method=method,
+                          options={'maxiter': n_iterations, 'disp': True})
+
+        chi2 = -self.chain.negativelogL(result['x'])
```
sibirrer: this is not chi2 but -log likelihood (there is a factor of two difference), so please rename.

sibirrer: why do you need a nagativelogL function to be called? It duplicates code in the Likelihood module which I think is not necessary, just call logL I recommend.

lenstronomy/Sampling/sampler.py
```diff
@@ -26,8 +27,35 @@ def __init__(self, likelihoodModule):
         self.chain = likelihoodModule
         self.lower_limit, self.upper_limit = self.chain.param_limits
 
+    def simplex(self, init_pos, n_iterations, method, print_key='SIMPLEX'):
```
sibirrer: some input parameters are not in the docstring.

test/test_Sampling/test_likelihood.py
```diff
@@ -118,6 +118,9 @@ def test_logL(self):
         num_data_evaluate = self.Likelihood.num_data
         npt.assert_almost_equal(logL/num_data_evaluate, -1/2., decimal=1)
 
+        negative_logL = self.Likelihood.negativelogL(args)
```
sibirrer: same as above: do you need this feature in the Likelihood module?

lenstronomy/Sampling/likelihood.py
```diff
@@ -247,5 +247,9 @@ def effective_num_data_points(self, **kwargs):
     def __call__(self, a):
         return self.logL(a)
 
+    def negativelogL(self, a):
```
sibirrer: do you need this feature? It duplicates quite a bit existing functions (I know there are already a lot, but I try to minimize them).

dangilman: the scipy routine has to minimize something, so passing in the likelihood or the log-likelihood wouldn't work since minimize this quantity makes the fit worse. I replaced the extra function in likelihood.py with a lambda function inside the simplex routine

sibirrer: I see, thanks!

Issue #164, opened 2020-03-12
Make dynesty sampler an optional install


ajshajib:
Just like pyMultinest and pyPolychord packages are optional, dynesty can also be made optional. In this way, users who only want to use emcee, for example, won't need to install another software that they don't need. Similarly, any software that depends on lenstronomy but is not going to use dynesty doesn't have to include dynesty to requirements.

PR #165, opened 2020-03-14
make dynasty sampler optional

sibirrer:
tracks issue #164 
First implementation to make the dynesty sampler optional. We might even consider moving the sampler requirement to a requirement_dev.tex such that it will be installed for testing but not directly when installing lenstronomy.

### DIFF
--- a/lenstronomy/Sampling/Samplers/dynesty_sampler.py
+++ b/lenstronomy/Sampling/Samplers/dynesty_sampler.py
@@ -39,7 +39,7 @@ class DynestySampler(NestedSampler):
             from schwimmbad import MPIPool
             import sys
 
-            pool = MPIPool()  # use_dill=True not supported
+            pool = MPIPool(use_dill=True)  # use_dill=True not supported for some versions of schwimmbad
             if not pool.is_master():
                 pool.wait()
                 sys.exit(0)
--- a/lenstronomy/Sampling/Samplers/dynesty_sampler.py
+++ b/lenstronomy/Sampling/Samplers/dynesty_sampler.py
@@ -1,7 +1,5 @@
 __author__ = 'aymgal'
 
-import os
-import shutil
 import numpy as np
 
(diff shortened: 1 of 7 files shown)


lenstronomy/Sampling/Samplers/dynesty_sampler.py
```diff
@@ -41,19 +39,19 @@ def __init__(self, likelihood_module, prior_type='uniform',
             from schwimmbad import MPIPool
             import sys
 
-            pool = MPIPool(use_dill=True)
+            pool = MPIPool()  # use_dill=True not supported
```
ajshajib: Back in September, when I tried to use dynesty with MPI, I could only make it work with dill. I am not sure if things have changed since.

sibirrer: On my local install 'use_dill' was not a supported input argument but happy to move it back

ajshajib: Right, that's because the owner of schwimmbad hasn't released a new pypy version that would include the `use_dill` option. In a recent commit, I set schwimmbad to be directly installed from github with this change implemented. I think, your installed version is from pypy. We can request schwimmbad owner to release a new version on pypy if that makes things easier for lenstronomy users.

sibirrer: I see, so the requirement for schwimmbad is set correctly linking to a specific GitHub commit ID.

ajshajib: Yes, that's the commit ID after my PR to add dill got merged.

sibirrer:
@ajshajib I will merge this PR unless you have further comments.

ajshajib:
> @ajshajib I will merge this PR unless you have further comments.

Yes, please do so, looks good to me.

Issue #166, opened 2020-03-16
MCMC sampler does not work with MPI=False and threadCount > 1


aymgal:
The `schwimmbad.choose_pool` method, when `mpi=False` and `threadCount` > 1 returns a MultiPool object that has no `is_master` attribute, hence the following error :
```python

  File "/Users/aymericg/Documents/EPFL/PhD_LASTRO/Code/Forked/lenstronomy_forked/lenstronomy/Workflow/fitting_sequence.py", line 94, in fit_sequence
    mcmc_output = self.mcmc(**kwargs)
  File "/Users/aymericg/Documents/EPFL/PhD_LASTRO/Code/Forked/lenstronomy_forked/lenstronomy/Workflow/fitting_sequence.py", line 230, in mcmc
    threadCount=threadCount, progress=progress)
  File "/Users/aymericg/Documents/EPFL/PhD_LASTRO/Code/Forked/lenstronomy_forked/lenstronomy/Sampling/sampler.py", line 149, in mcmc_emcee
    if pool.is_master():
AttributeError: 'MultiPool' object has no attribute 'is_master'

```

In addition,  I see that the new schwimmbad way of dealing with pools calls the `multiprocessing` module instead of the `multiprocess`, as it was done before in lenstronomy, and which is more robust because of pickling issues in `multiprocessing`. @ajshajib , could you please look into that since you introduced this change ?

I did not test for the PSO, but it's likely to give the same result.

In the meantime, I will temporarily make my own fork of schwimmbad to fix that, but it's not viable to link lenstronomy to this fork.

sibirrer:
@aymgal do you satisfy the requirements as indicated in requirements.txt? @ajshajib added a temporary fix to a specific commit of schwimmbad. Just checking not to duplicate efforts in this direction.

aymgal:
Yes I do, I took the last github version of schwmimbad as required. The schwimmbad multithreading pool has no `is_master` attribute.

ajshajib:
Would it be preferable if we open an issue for schwimmbad to add an `is_master` method for the `MPIPool` class? After all, the philosophy of schwimmbad is to make switching pools between single-processor, multithreads, and MPI seamless.

The pickling issue between multiprocess and multiprocessing can be resolved by using `use_dill=True` in `choose_pool`.  multiprocessing is a default Python 3 package, whereas multiprocess just forks multiprocessing to use dilll. It's the same author(s) for both multiprocessing and dill.

[...]

PR #167, opened 2020-03-16
Multithreading mpi

sibirrer:
addresses issue #166


aymgal:
> You should check for `if processes == 1 or mpi`, because `schwimmbad.choose_pool()` prioritizes `mpi=True` over `processes > 1`. So, in case a user sends `mpi=True, processes > 1`, then there would be pickling problem as you are taking out `use_dill` inside the `else` clause.

You're right, thanks for pointing this out !

sibirrer:
Thank you @aymgal , @ajshajib ! @ajshajib I added you as a reviewer to make sure the MPI option is ok for your usage. If you are happy you can push the button to merge - thank you!

PR #170, opened 2020-03-20
Anisotropy dev

sibirrer:
minor updates with removed redundancies


sibirrer:
Coverage only reduced because of the removal of code. No addition of untested code.

PR #173, opened 2020-03-31
Multithreading mpi

sibirrer:
This PR should solve the issue of multi-threading with pickle. Instead of using multiprocessing, we are using multiprocess, which is using dill instead. This change also requires to modify the MultiPool class of schwimmbad. We instead patched this class according to the fork https://github.com/aymgal/schwimmbad/commit/6d597efd7471c40c7026c55caeb72ea7fbfdb01d and added it directly to lenstronomy to circumvent dependencies on specific forks.


requirements.txt
```diff
@@ -12,6 +12,7 @@ dynesty
 pymultinest
 pypolychord
 nestcheck
--e git://github.com/adrn/schwimmbad.git@master#egg=schwimmbad
+schwimmbad==0.3.0
```
ajshajib: You will still need latest commit of Schwimmbad from github for MPI use.

sibirrer:
Thank you guys

PR #178, opened 2020-05-12
Add support for backup MCMC state and use it as starting point

aymgal:
This feature propagates the emcee "backend" feature to the MCMC sampler routine. This allows the user to provide a name of a backup file that keeps track of the MCMC sampling (_including burn-in iterations_), and can be used as initial state of a future MCMC run.

Two new parameters were added to the MCMC fitting step:
- `backup_filename` (defaults to `None`): name of the HDF5 file used to save the chains during sampling.
- `start_from_backup` (defaults to `False`): if `True`, starts from the state saved in the backup file. If `n_burn` is set, this will discard samples from the very beginning of the samples backed up (there was no burn-in applied before saving the previous samples).


sibirrer:
Thanks a lot @aymgal ! This looks great!

PR #179, opened 2020-05-12
fix tnfw deflection angle at x == 1

dangilman:
there were some strange lines in the TNFW function that were redundant and resulted in the wrong deflection angle at x == 1

### DIFF
--- a/test/test_LensModel/test_Optimizer/test_multi_plane.py
+++ b/test/test_LensModel/test_Optimizer/test_multi_plane.py
@@ -135,24 +135,24 @@ class TestMultiPlaneOptimizer(object):
 
     def test_fixed_routines(self):
 
-        # sie = FixedPowerLaw_Shear(['SIE','SHEAR'],self.kwargs_lens_simple,self.x_pos_simple,self.y_pos_simple,
-        #                           constrain_params={'shear':[0.06, 0.01]})
+        sie = FixedPowerLaw_Shear(['SPEMD','SHEAR'],self.kwargs_lens_simple,self.x_pos_simple,self.y_pos_simple,
+                                  constrain_params={'shear':[0.06, 0.01]})
 
-        # powerlaw = VariablePowerLaw_Shear(['PEMD','SHEAR'],self.kwargs_lens_simple,self.x_pos_simple,self.y_pos_simple,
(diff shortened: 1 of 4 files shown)


dangilman:
added the changes you requested and changed SPEMD -> PEMD in the optimizer class

test/test_LensModel/test_Optimizer/test_multi_plane.py
```diff
@@ -135,10 +135,10 @@ def test_params(self):
 
     def test_fixed_routines(self):
 
-        sie = FixedPowerLaw_Shear(['SPEMD','SHEAR'],self.kwargs_lens_simple,self.x_pos_simple,self.y_pos_simple,
+        sie = FixedPowerLaw_Shear(['PEMD','SHEAR'],self.kwargs_lens_simple,self.x_pos_simple,self.y_pos_simple,
```
sibirrer, line 138: can you change this model to 'SIE' or 'SPEP'? The reason is that the PEMD (former SPEMD) requires fastell4py which is not part of the testing (returns zeros). Sorry that I pick this up now.

dangilman:
I can't swap out lens models because the class expects certain kwargs_lens arguments, but I can just take out the test that uses fastell

sibirrer:
thanks! Now, there is less test coverage in the optimizer :(

dangilman:
Is there a way to include test functions for profiles that require fastell? Not sure how else to proceed 

- Daniel

sibirrer:
> Is there a way to include test functions for profiles that require fastell? Not sure how else to proceed - Daniel

You can put in PEMD models in the testing and they will not raise, they just don't do what they are supposed to do (you will not predict 4 images etc). You don't need to test the specific lens model but rather whether the optimizer does what it supposed to do. A SPEP model e.g. is effectively almost the same (same parameters but ellipticity in the potential) and all the solver tests on my side are validated with this model instead.

dangilman:
ok well as long as the PEMP doesn't raise an error the tests should pass and it should be ok right?

sibirrer:
The tests are still lower than previously...

dangilman:
how about now? I've been meaning to re-write the whole Optimizer class, but right now I don't have time to do that

sibirrer:
thank you! Sorry for the hassle and the persistence.

PR #181, opened 2020-05-15
elliptical power law (EPL), Tessore & Metcalf 2015

ntessore:
Implements the (singular) elliptical power law (EPL) profile in the form given by [Tessore & Metcalf (2015)](https://arxiv.org/pdf/1507.01819.pdf).

The critical calculation is outsourced to `scipy.special.hyp2f1`. This could be sped up quite significantly using the technique from Section 4 of the paper, in particular if array operations are used.

I have not really looked into the other power law models, so you might want to align the parametrisation before considering merging this. Our power law exponent `t` is the exponent of the 2D density, and differs by unity from models that parametrise the 3D exponent `gamma`.

### DIFF
--- a/lenstronomy/LensModel/profile_list_base.py
+++ b/lenstronomy/LensModel/profile_list_base.py
@@ -110,6 +110,12 @@ class ProfileListBase(object):
         elif lens_type == 'SPEMD':
             from lenstronomy.LensModel.Profiles.spemd import SPEMD
             return SPEMD()
+        elif lens_type == 'EPL':
+            from lenstronomy.LensModel.Profiles.epl import EPL
+            return EPL()
+        elif lens_type == 'EPL_SIMPLE':
+            from lenstronomy.LensModel.Profiles.epl import EPLMajorAxis
+            return EPLMajorAxis()
         elif lens_type == 'NFW':
             from lenstronomy.LensModel.Profiles.nfw import NFW
             return NFW()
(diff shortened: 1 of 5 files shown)


test/test_LensModel/test_Profiles/test_epl.py
```diff
@@ -0,0 +1,71 @@
...
+    def setup(self):
+        from lenstronomy.LensModel.Profiles.epl import EPL
+        self.EPL = EPL()
+        from lenstronomy.LensModel.Profiles.nie import NIE
+        self.NIE = NIE()
+
+    def test_function(self):
+        phi_E = 1.
+        t = 1.
+        q = 0.999
```
sibirrer, line 23: can you also give a test for a lower axis ratio as done for the derivatives?

lenstronomy/LensModel/profile_list_base.py
```diff
@@ -110,6 +110,12 @@ def _import_class(lens_type, custom_class, z_lens=None, z_source=None):
         elif lens_type == 'SPEMD':
             from lenstronomy.LensModel.Profiles.spemd import SPEMD
             return SPEMD()
+        elif lens_type == 'EPL':
+            from lenstronomy.LensModel.Profiles.epl import EPL
+            return EPL()
+        elif lens_type == 'EPL_SIMPLE':
```
sibirrer, line 116: the name 'simple' might be a bit misleading. You can call it' EPL_MAJORAXIS'  or we don't even need to give the front-end user axes to it.

sibirrer:
Great! Thank you very much!

PR #183, opened 2020-05-29
Penalty in TD likelihood for predicted images inconsistent with observations 

aymgal:
When using the point source option `additional_images_list` for allowing the model to predict more images than explicitly modelled, while sampling `D_dt` in the likelihood, I got an issue when comparing model and measured time delays. For instance, the PSO can initially sample a wrong lens model, that will predict a smaller number of delays, inconsistently with measured delays.

This PR proposes a way to solve it: when the "additional image" option is used, and number of predicted delays is smaller than number of measured delays, a penalty (-1e15) is added to the likelihood.

### DIFF
--- a/lenstronomy/PointSource/point_source.py
+++ b/lenstronomy/PointSource/point_source.py
@@ -43,6 +43,7 @@ class PointSource(object):
         self._fixed_magnification_list = fixed_magnification_list
         if additional_images_list is None:
             additional_images_list = [False] * len(point_source_type_list)
+        self.additional_images_list = additional_images_list
         for i, model in enumerate(point_source_type_list):
             if model == 'UNLENSED':
                 from lenstronomy.PointSource.point_source_types import Unlensed
--- a/lenstronomy/Sampling/Likelihoods/time_delay_likelihood.py
+++ b/lenstronomy/Sampling/Likelihoods/time_delay_likelihood.py
@@ -24,6 +24,7 @@ class TimeDelayLikelihood(object):
         self._delays_errors = np.array(time_delays_uncertainties)
         self._lensModel = lens_model_class
         self._pointSource = point_source_class
+        self._ps_model_index = 0   # index of ps model to consider in the point_source_model_list
 
     def logL(self, kwargs_lens, kwargs_ps, kwargs_cosmo):
         """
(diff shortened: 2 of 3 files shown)


sibirrer:
Hi @aymgal , thanks for pointing this out and proposing a fix. There is an additional subtlety in using `additional_image_list` in combination with the time delay likelihood: As the image positions get re-computed (and over-written) by the lens equation solver, the ordering of the images may not necessarily coincide with the initial ordering of the images and their provided relative delays. The lens equation solver by default orders the images according to their arrival time.

I would guess that if there are fewer images predicted than present (and originally modeled), the penalty of the lens equation solver and/or the matched source position may also jump in.

In that sense, it might be better for the time-delay likelihood to evaluate the position of the model parameters prior of the lens equation solver (at least for `LENSED_POSITION`). What do you think of this fix? It would require minor changes in the `PointSource` module to have this functionality.

The way Anowar did it for 0408 was not using Ddt and the time-delay likelihood in the modeling and post-process it with the image positions of the model.

aymgal:
Yes I see the issue here. If I understand correctly, you suggest to check the image positions, _prior_ to the likelihood evaluation, to ensure the ordering remained the same, consistently to time delays. I am thinking about taking the `min_distance` value as a way to check if an image position (taken from `kwargs_ps`) matches one of the predicted positions (taken from the lens solver), and apply some sort of re-ordering based on that. Do you agree with this ?

I'm aware that Anowar did not use TD to constrain the lens model, but I would like to use it this time, since it should help breaking degeneracies that can arise when there are several main deflectors (speedup the convergence as well).

sibirrer:
I would suggest even something simpler, 'just' using the arguments in `kwargs_ps` directly (as done without the additional images constraints). In case the lens equation is not satisfied, the image model and other punishing factors related to the lens equation solution do act anyways. I think this is the most save way to guarantee consistency with other approaches not using additional image constraints. I can code it up but there might occur some conflicts with this PR, if you are ok with it.

aymgal:
Ok. So simply ignoring additional predicted images other than those explicitly modelled...
Since it seems you have a much clearer idea of how to implement it, no problem on my side if you code it up! You can close the PR if you want.

sibirrer:
I just made a push to master. If this fixes your issue, you can close this PR, thanks.

aymgal:
Thanks Simon!

PR #192, opened 2020-07-24
remove unnecessary code

dangilman:


dangilman:
Sure, how do I test that a specific line raises a ValueError? 

- Daniel

sibirrer:
> Sure, how do I test that a specific line raises a ValueError? - Daniel
> […](#)

there are specific raise test functions. You can look for examples in tests of other classes, such as in test/test_Data/test_imaging_data.py as an example.

sibirrer:
Thanks! So the Optimizer module is at 96.15% test coverage. If you can bring it above 97% that'd be good. You can see what lines are tested here:
https://coveralls.io/builds/32307679
Sorry to be obnoxious here, as usage of lenstronomy increases, I want to make sure there is as good a testing as possible.

dangilman:
I removed some untested lines that were inaccessible anyways (I don't remember adding them myself)

sibirrer:
> I removed some untested lines that were inaccessible anyways (I don't remember adding them myself)

so the set_static() and set_dynamic() calls can speed up lens equation solving under fixed parameters of the mass profiles (as it stores some re-ocurant quantities in cache for some lens models). I added it as an option in this code to allow to make you use it.

sibirrer:
thanks! up to you to plug back in the set_static and set_dynamic calls. Otherwise I am happy to approve the PR. Just let me know.

dangilman:
Yeah you can leave them out

PR #194, opened 2020-07-27
Add ObservationConfig folder with LSST settings

mueland:
(and correct a few typos)


lenstronomy/SimulationAPI/ObservationConfig/LSST.py
```diff
@@ -0,0 +1,29 @@
+"""Provisional LSST instrument and observational settings.
+See Optics and Observation Conditions spreadsheet at
+https://docs.google.com/spreadsheets/d/1pMUB_OOZWwXON2dd5oP8PekhCT5MBBZJO1HV7IMZg4Y/edit?usp=sharing for list of
+sources. """
+
+camera = {'read_noise': 10,  # will be <10
+               'pixel_scale': 0.2,
+               'ccd_gain': 1}  # gain is 2.3 electrons/ADU, but zero points below are calibrated with gain = 1
```
sibirrer, line 8: for the latest lenstronomy version, the gain should not play a role when providing all the noise and brightness quantities in e-/s, so you can set 2.3 here and should not expect differences in the simulation output when specifying the output as e- (new default, previously ADUs)

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
```diff
@@ -0,0 +1,29 @@
...
+
```
sibirrer, line 9: Can you specify that these aspects (in particular number of exposures) are for a certain period (e.g. full 10 years)? This may help the user be aware that the number of exposures might change when they simulate a shorter survey time.

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
mueland: Great, I had forgotten counts in photoelectrons is now the default for zero points!

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
mueland: Sure! Should I also specify other args more (e.g. units), or leave this for SingleBand class declaration?

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
```diff
@@ -5,25 +5,40 @@
...
                    'magnitude_zero_point': 27.79,
                    'num_exposures': 460,
                    'seeing': 0.71,
                    'psf_type': 'GAUSSIAN'}
+
+"""
+exposure_time: exposure time per image (in seconds)
+sky_brightness: sky brightness (in magnitude per square arcseconds in units of electrons)
+magnitude_zero_point: magnitude in which 1 count (e-) per second per arcsecond square is registered
+num_exposures: number of exposures that are combined
```
jiwoncpark, line 40: Maybe move the comment on survey duration (from line 18) here? In the future, we might want the user to specify the field (WFD, DDF) or the duration (year x or the whole 10 years), and internally query the correct `num_exposures`. I think having WFD for 10 years for now is fine, but just to make it clearer!

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
jiwoncpark: We should specify the units (`data_count_unit`). Given that `read_noise` is given in electrons, it should be set as `'e-'`.

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
```diff
@@ -5,25 +5,40 @@
...
 
-i_band_obs = {'exposure_time': 30.,
+i_band_obs = {'exposure_time': 15.,
                    'sky_brightness': 20.48,
                    'magnitude_zero_point': 27.79,
                    'num_exposures': 460,
                    'seeing': 0.71,
                    'psf_type': 'GAUSSIAN'}
+
+"""
```
jiwoncpark, line 36: Just a thought because you already have a docstring, it might be good to wrap this in an LSST class. I'm thinking about future cases where the user might want to configure an LSST-like PSF that's `PIXEL` rather than `GAUSSIAN` or experiment with different years of coadd (varying `num_exposures`). If you have a class structure, the user can input those when instantiating the LSST object and the class can spit out the right dictionary for them.

test/test_SimulationAPI/test_LSST.py
```diff
@@ -0,0 +1,40 @@
...
+        assert self.g_band._read_noise == LSST.camera['read_noise']
+
+        assert self.g_band._exposure_time == LSST.g_band_obs['exposure_time']
+        assert self.g_band._sky_brightness_ == LSST.g_band_obs['sky_brightness']
+        assert self.g_band._num_exposures == LSST.g_band_obs['num_exposures']
+        assert self.g_band._seeing == LSST.g_band_obs['seeing']
+        assert self.g_band._psf_type == LSST.g_band_obs['psf_type']
+        assert self.g_band._magnitude_zero_point == LSST.g_band_obs['magnitude_zero_point']
+
+    # add r and i band tests?
```
jiwoncpark, line 36: You can format the `test_g_band` test in a loop so that there's little code duplication for `r` and `i` (and less room for typos!).

test/test_SimulationAPI/test_LSST.py
```diff
@@ -0,0 +1,40 @@
...
+        assert self.g_band.ccd_gain == LSST.camera['ccd_gain']
+        assert self.g_band.pixel_scale == LSST.camera['pixel_scale']
+        assert self.g_band._read_noise == LSST.camera['read_noise']
+
+        assert self.g_band._exposure_time == LSST.g_band_obs['exposure_time']
+        assert self.g_band._sky_brightness_ == LSST.g_band_obs['sky_brightness']
+        assert self.g_band._num_exposures == LSST.g_band_obs['num_exposures']
+        assert self.g_band._seeing == LSST.g_band_obs['seeing']
+        assert self.g_band._psf_type == LSST.g_band_obs['psf_type']
+        assert self.g_band._magnitude_zero_point == LSST.g_band_obs['magnitude_zero_point']
```
jiwoncpark, line 34: I also recommend using `self.assertEqual` or `self.assertAlmostEqual` (or the analogs in `np.testing`) so that you can include the error message if the assert statement fails (something like `seeing doesn't match`). If you're looping over an array like `['ccd_gain', 'pixel_scale', ...]` to do the tests, you can string format the tests very easily!

mueland: Sure, I'll adjust that so I can add error messages! Re looping, I was considering that before, but I didn't end up doing it because the kwargs vary slightly in order and naming between different dictionaries. Would you loop through two arrays concurrently, then - one following dict naming conventions, one following SingleBand conventions?

jiwoncpark: Hmm I didn't quite understand the "ordering" part (because dictionaries are unordered), but you can create a dictionary that maps your dict naming convention (keys) to the SingleBand conventions (values) and iterate through that dictionary. Note that you can get an object attribute using a string of the attribute name via, e.g. `getattr(self, '_seeing')`

mueland: I guess I was thinking about iterating through a sorted dict when I referenced ordering before - but mapping my dict to single band conventions is a good idea. I got it to work with two side-by-side arrays but it's not very readable, so I'll change it. Thanks for the attribute tip, I was just trying to figure that out!

jiwoncpark: Yes! In general, dictionaries are better for readability than iterating through two arrays using the same index :-)

jiwoncpark: Note for your case the instance should be `self.g_band` not `self`!

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
mueland: In SingleBand class declaration, data_count_unit is defined as "string, unit of the data (not noise properties - see other definitions),
         'e-': (electrons assumed to be IID),
         'ADU': (analog-to-digital unit)"
I interpreted this as meaning that the data_count_unit does not correspond to the units of read_noise?

jiwoncpark: Oops, you are right! Please disregard my above comment. `data_count_unit` is the unit that'll end up in the final image. And to answer your earlier question, we should leave this up to the user when they instantiate `SingleBand` (or any other children class).

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
mueland: Good idea! You are just referring to wrapping the observational settings in an LSST class, right? In addition to parameters like psf_type and years of coadd, would the constructor also take band?

jiwoncpark: Hmmmm, right. I need to think more about how to fold in the multi-band image rendering in Baobab, but yes, can we please also take the band? If I wanted to simulate gri images of a single system, I'd want to instantiate 3 different objects of `LSST`

mueland: Great, so to confirm - I'm leaving camera settings out of the class? Should `LSST` inherit the `Observation` class?

jiwoncpark: Maybe `LSST` can have the properties `camera` and `obs` (the dictionaries) so you can bring in the camera settings into the class as well. It can inherit from both `Instrument` and `Observation` OR you can inherit from `SingleBand` and allow (force?) the user to update the `data_count_unit` in `SingleBand` (by slightly modifying `SingleBand`). The former might be simpler for now...

sibirrer: > Maybe `LSST` can have the properties `camera` and `obs` (the dictionaries) so you can bring in the camera settings into the class as well. It can inherit from both `Instrument` and `Observation` OR you can inherit from `SingleBand` and allow (force?) the user to update the `data_count_unit` in `SingleBand` (by slightly modifying `SingleBand`). The former might be simpler for now...

@jiwoncpark 
I don't think inheritance is required. The sole aim of such a class is to provide the constructing elements for the SimAPI object, otherwise we are mixing up too many class roles.

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
```diff
@@ -35,11 +27,50 @@
...
+    when coadd_years = 10: num_exposures is baseline num of visits over 10 years (x2 since 2x15s exposures per visit)
+:keyword seeing: Full-Width-at-Half-Maximum (FWHM) of PSF
+:keyword psf_type: string, type of PSF ('GAUSSIAN' and 'PIXEL' supported)
 """
 
+class LSST(Instrument, Observation):
+    """
+    class contains LSST instrument and observation configurations
+    """
+    def __init__(self, band='g', psf_type='GAUSSIAN', coadd_years=10):
```
sibirrer, line 43: I like the design of this class! It however does not need to inherit `Instrument` and `Observation`. The only think the class needs to return are the keyword argument settings that allow to construct those through `SimAPI`, otherwise we are mixing different roles of classes. This class is only providing the settings themselves.

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
```diff
@@ -35,11 +27,50 @@
...
+               'pixel_scale': 0.2,
+               'ccd_gain': 2.3,
+          }
+        """
+        :keyword read_noise: std of noise generated by read-out (in units of electrons)
+        :keyword pixel_scale: scale (in arcseconds) of pixels
+        :keyword ccd_gain: electrons/ADU (analog-to-digital unit).
+        """
+
+    # TODO: add method that creates SingleBand object?
```
sibirrer, line 76: effectively you can add one single routine to provide the SimAPI settings as a merge of `camera` and `obs`

mueland: Yes! Should I leave that to be done externally so this class remains lightweight, or just include a simple method here that merges the two?

sibirrer: > Yes! Should I leave that to be done externally so this class remains lightweight, or just include a simple method here that merges the two?

maybe having a simple method that merges the two would be great. This means less code outside.

mueland: Sure! Just returning the merged dicts, leaving the instance of SimAPI to be created externally?

Btw, when you said earlier that "the current class will most likely crash as the inheritance can not be provided with the current init() statement", was that because if you create a class using inheritance it needs to have the arguments from the base class(es)?

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
jiwoncpark: Agreed, and as far as I'm concerned I just need two dictionaries.

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
sibirrer: > Btw, when you said earlier that "the current class will most likely crash as the inheritance can not be provided with the current init() statement", was that because if you create a class using inheritance it needs to have the arguments from the base class(es)?

Yes, if not otherwise specifically instanced (don't worry about that)

lenstronomy/SimulationAPI/ObservationConfig/LSST.py
```diff
@@ -0,0 +1,81 @@
...
+        elif band == 'i':
+            self.obs = i_band_obs
+        else:
+            raise ValueError("band %s not supported! Choose 'g', 'r', or 'i'." % band)
+
+        if psf_type != 'GAUSSIAN':
+            raise ValueError("psf_type %s not supported!" % psf_type)
+
+        if coadd_years != 10:
+            raise ValueError(" %s coadd_years not supported! "
```
sibirrer, line 62: this is the only line of code that is not tested yet in this PR

test/test_SimulationAPI/test_LSST.py
```diff
@@ -0,0 +1,66 @@
...
+        self.assertEqual(explicit_g.obs, default.obs)
+
+        with self.assertRaises(ValueError):
+            bad_band = LSST(band='u')
+
+        with self.assertRaises(ValueError):
+            bad_psf = LSST(psf_type='blah')
+
+        with self.assertRaises(ValueError):
+            bad_coadd_years = LSST(psf_type=100)
```
sibirrer, line 47: here you probably wanted to test the coda_year argument, not the pdf_type

mueland: oops, nice catch!

sibirrer:
Awesome! Thanks @jiwoncpark for the review and congratulations @mueland for your first accepted PR :)

PR #198, opened 2020-08-08
Update DES and LSST configs

mueland:
Adjusted magnitude zero points for DES (now corresponding to counts in e-), and added u, z and y bands to LSST configs.


sibirrer:
Thank you @jiwoncpark ! Let me know if things don't work with the changes.

PR #200, opened 2020-08-15
Increase LSST coadd_years support

mueland:
Minor change expanding support for coadd_years in LSST configs from 10 to 1-10

### DIFF
--- a/lenstronomy/SimulationAPI/ObservationConfig/LSST.py
+++ b/lenstronomy/SimulationAPI/ObservationConfig/LSST.py
@@ -68,7 +68,7 @@ class LSST(object):
 
         :param band: string, 'u', 'g', 'r', 'i', 'z' or 'y' supported. Determines obs dictionary.
         :param psf_type: string, type of PSF ('GAUSSIAN' supported).
-        :param coadd_years: int, number of years corresponding to num_exposures in obs dict. Currently supported: 10.
+        :param coadd_years: int, number of years corresponding to num_exposures in obs dict. Currently supported: 1-10.
         """
         if band.isalpha():
             band = band.lower()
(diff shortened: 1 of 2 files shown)


lenstronomy/SimulationAPI/ObservationConfig/LSST.py
```diff
@@ -90,9 +90,11 @@ def __init__(self, band='g', psf_type='GAUSSIAN', coadd_years=10):
         if psf_type != 'GAUSSIAN':
             raise ValueError("psf_type %s not supported!" % psf_type)
 
-        if coadd_years != 10:
-            raise ValueError(" %s coadd_years not supported! "
-                             "You may manually adjust num_exposures in obs dict if required." % coadd_years)
+        if coadd_years > 10 or coadd_years < 1:
+            raise ValueError(" %s coadd_years not supported! Choose an integer between 1 and 10." % coadd_years)
+        elif coadd_years != 10:
+            self.obs['num_exposures'] = coadd_years*self.obs['num_exposures']//10
```
sibirrer, line 96: can you specify in the LSST config file that the number of exposures is for the 10 years survey to make sure that it is well documented to understand this calculation here?

mueland: Currently in the documentation for num_exposures I note: `when coadd_years = 10: num_exposures is baseline num of visits over 10 years (x2 since 2x15s exposures per visit)`

Happy to make this more explicit if you want!

sibirrer: ah, I see. It's ok, I was looking further up in the file. It's fine.

PR #201, opened 2020-08-20
Add Multipole LensModel

LyneVdV:


lenstronomy/LensModel/Profiles/multipole.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+import numpy as np
+
+import lenstronomy.Util.param_util as param_util
+from lenstronomy.LensModel.Profiles.base_profile import LensProfileBase
+
+class Multipole(LensProfileBase):
+    """
+    This class contains a multipole contribution (for 1 component with m>=2)
```
sibirrer, line 11: are there some literature references for this family of profiles?

lenstronomy/LensModel/Profiles/multipole.py
```diff
@@ -0,0 +1,38 @@
...
+    m : int, multipole order
+    a_m : float, multipole strength
+    phi_m : float, multipole orientation in radian
+    """
+    param_names = ['m', 'a_m', 'phi_m', 'center_x', 'center_y']
+    lower_limit_default = {'m': 2,'a_m':0, 'phi_m':-np.pi, 'center_x': -100, 'center_y': -100}
+    upper_limit_default = {'m': 100,'a_m':100, 'phi_m':np.pi, 'center_x': 100, 'center_y': 100}
+
+    def function(self, x, y, m, a_m, phi_m, center_x=0, center_y=0):
```
sibirrer, line 20: for documentations and such that it appears in the readthedocs html format, can you add the doc strings to each definition, not just on top of the class?

lenstronomy/LensModel/Profiles/multipole.py
```diff
@@ -0,0 +1,38 @@
...
+        f_ = r*a_m /(1-m**2) *np.cos(m*(phi-phi_m))
+        return f_
+    def derivatives(self,x,y, m, a_m, phi_m, center_x=0, center_y=0):
+        r, phi = param_util.cart2polar(x, y, center_x=center_x, center_y=center_y)
+        f_x = np.cos(phi)*a_m/(1-m**2) *np.cos(m*(phi-phi_m)) + np.sin(phi)*m*a_m/(1-m**2)*np.sin(m*(phi-phi_m))
+        f_y = np.sin(phi)*a_m/(1-m**2) *np.cos(m*(phi-phi_m)) - np.cos(phi)*m*a_m/(1-m**2)*np.sin(m*(phi-phi_m))
+        return f_x,f_y
+    def hessian(self,x,y, m, a_m, phi_m, center_x=0, center_y=0):
+        r, phi = param_util.cart2polar(x, y, center_x=center_x, center_y=center_y)
+        if isinstance(r, int) or isinstance(r, float):
```
sibirrer, line 31: these lines of code can be simplified (should do it for other parts of lenstronomy too) with
r = np.minimum(r, 0.000001)
This works for both, arrays and single numbers

lenstronomy/LensModel/profile_list_base.py
```diff
@@ -224,6 +224,9 @@ def _import_class(lens_type, custom_class, z_lens=None, z_source=None):
         elif lens_type == 'NumericalAlpha':
             from lenstronomy.LensModel.Profiles.numerical_deflections import NumericalAlpha
             return NumericalAlpha(custom_class)
+        elif lens_type == 'MULTIPOLE':
```
sibirrer, line 227: can you add a test to check that this model is being initiated in the LensModel class? You may simply add it to test_single_plane.py test_init()

test/test_LensModel/test_Profiles/test_multipole.py
```diff
@@ -0,0 +1,92 @@
+__author__ = 'lynevdv'
+
+
+from lenstronomy.LensModel.Profiles.multipole import Multipole
+
+import numpy as np
+import pytest
+import numpy.testing as npt
+
+class TestMultipole(object):
```
sibirrer, line 10: thanks for adding the tests! In the current form, the tests are executed to reproduce a numerical value but this does not allow to assess wether these are the correct ones and self-consistent between potential, deflection angle, and hessian.
There is a specific testing class in test_numerical_lens_differentials.py which allows for testing the numerical differentials to match the analytical outputs. Can you add the Multipole to this tests (you can follow the examples of other lens models). I suggest adding these tests for at least two different multipole moments m. Thank you!

lenstronomy/LensModel/Profiles/multipole.py
LyneVdV: yes ! Xu et al. 2013 appendix B3. I will add it !

lenstronomy/LensModel/Profiles/multipole.py
LyneVdV: yes, good idea !

lenstronomy/LensModel/Profiles/multipole.py
LyneVdV: shouldn't it be np.maximum(r,0.000001) ?
I didn't know about those, I will replace the corresponding lines !

sibirrer: > shouldn't it be np.maximum(r,0.000001) ?
> I didn't know about those, I will replace the corresponding lines !
ah yes, np.maximum() :)

lenstronomy/LensModel/profile_list_base.py
LyneVdV: yes !

test/test_LensModel/test_Profiles/test_multipole.py
LyneVdV: yes !

sibirrer:
Thanks! I will make some small changes in the doc strings for the right conventions, i.e. :param param_name: ... and a :return: statement.

PR #202, opened 2020-08-24
update

dangilman:
Hi, seems like it makes sense to have the power law profiles all use consistent names for parameters (e.g. "gamma" for the slope instead of "t"). This would make it easier to test the profiles by switching back and forth without having to change the keyword arguments in my code. 

Also, on this code version for some reason most of the NIE tests failed even though I didn't touch those files...

### DIFF
--- a/lenstronomy/LensModel/Profiles/epl.py
+++ b/lenstronomy/LensModel/Profiles/epl.py
@@ -10,22 +10,22 @@ from scipy.special import hyp2f1
 class EPL(LensProfileBase):
     """
     Elliptical Power Law
-    kappa = (2-t)/2*(b/r)^t
+    kappa = (2-gamma)/2*(b/r)^gamma
     """
-    param_names = ['theta_E', 'e1', 'e2', 't', 'center_x', 'center_y']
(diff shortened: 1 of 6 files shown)


lenstronomy/LensModel/Profiles/epl.py
```diff
@@ -10,22 +10,22 @@
 class EPL(LensProfileBase):
     """
     Elliptical Power Law
-    kappa = (2-t)/2*(b/r)^t
+    kappa = (2-gamma)/2*(b/r)^gamma
```
sibirrer, line 13: this is not the same gamma definition as for the PEMD and gamma=1 is the isothermal profile (because t is defined in projection).

dangilman:
Ahh I see, I didn’t realize they were different definitions. Nevermind! 

- Daniel

sibirrer:
should not take much to change it for the entire profile and it's a good thing (I think). You can also add one layer and rename the current EPL model such that both versions are accessible. With the 't' parameterization it is consistent with Tagore et al. and with gamma with the PEMD.

Issue #203, opened 2020-08-26
HST config to run without separate kernel_point_source


sibirrer:
https://github.com/sibirrer/lenstronomy/blob/257119cfb3079e922a4b63347a23b46dcad161c2/lenstronomy/SimulationAPI/ObservationConfig/HST.py#L13

The HST data config only runs with particular settings to add a 'kernel_point_source'. Either this default setting needs to be 'GAUSSIAN' to allow for usage or a pixelated kernel needs to be injected as well (or a raise statement specifying what needs to be done). @jiwoncpark @mueland 
What solution do you prefer? I personally prefer having the default as GAUSSIAN with the option for pixelated kernels.

jiwoncpark:
Ah, this makes sense. For the PSF FWHM, would you be comfortable with having an approximate undrizzled value? From what I understand, the final drizzled PSF size can vary significantly depending on the drizzling method and base templates. Windhorst et al 2011 (source @mueland cites in `HST.py`) gives a pre-drizzled value (I think) of 0.15" for the WFC3/IR F160W filter:
![image](https://user-images.githubusercontent.com/15166916/91338721-5b8eeb80-e7a3-11ea-8095-753f0a29e408.png)

sibirrer:
@jiwoncpark the current pixel scale is 0.08" in the config, this means drizzled. So having a drizzled PSF FWHM estimate might be more appropriate for the given settings. Perhaps we can comment directly in the code from where we use this values such that the user can gage it. This is approximate after all.

sibirrer:
PR #204 solves this issue

PR #204, opened 2020-08-27
Default to Gaussian PSF with seeing set equal to approximate pixel scale from drizzled PSF

jiwoncpark:
Addressing #203 with some inline comments


sibirrer:
Thank you! Looks great!

PR #211, opened 2020-09-22
regularize EPL at origin

dangilman:
I just implemented the property I want in the deflection angles alpha(0, 0) = 0 by replacing nan values with zero. The function can only be nan at the origin so it works

### DIFF
--- a/lenstronomy/LensModel/Profiles/epl.py
+++ b/lenstronomy/LensModel/Profiles/epl.py
@@ -186,15 +186,6 @@ class EPLMajorAxis(LensProfileBase):
 
         super(EPLMajorAxis, self).__init__()
 
-    @staticmethod
-    def _regularize(value):
-
-        out = np.nan_to_num(value, posinf=1e+3, neginf=-1e+3)
-        if isinstance(value, float) or isinstance(value, int):
-            out = float(out)
-
-        return out
-
     def function(self, x, y, b, t, q):
         """
         returns the lensing potential
@@ -224,8 +215,8 @@ class EPLMajorAxis(LensProfileBase):
         alpha = 2/(1+q)*(b/R)**t*R_omega
 
         # return real and imaginary part
-        alpha_real = self._regularize(alpha.real)
-        alpha_imag = self._regularize(alpha.imag)
+        alpha_real = np.nan_to_num(alpha.real)
+        alpha_imag = np.nan_to_num(alpha.imag)
 
         return alpha_real, alpha_imag
 
@@ -241,7 +232,7 @@ class EPLMajorAxis(LensProfileBase):
 
         # convergence, eq. (2)
         kappa = (2 - t)/2*(b/R)**t
-        kappa = self._regularize(kappa)
+        kappa = np.nan_to_num(kappa)
 
         # deflection via method
         alpha_x, alpha_y = self.derivatives(x, y, b, t, q)
@@ -249,8 +240,8 @@ class EPLMajorAxis(LensProfileBase):
         # shear, eq. (17), corrected version from arXiv/corrigendum
         gamma_1 = (1-t)*(alpha_x*cos - alpha_y*sin)/r - kappa*cos2
         gamma_2 = (1-t)*(alpha_y*cos + alpha_x*sin)/r - kappa*sin2
-        gamma_1 = self._regularize(gamma_1)
-        gamma_2 = self._regularize(gamma_2)
+        gamma_1 = np.nan_to_num(gamma_1)
+        gamma_2 = np.nan_to_num(gamma_2)
 
         # second derivatives from convergence and shear
         f_xx = kappa + gamma_1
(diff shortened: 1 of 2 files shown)


lenstronomy/LensModel/Profiles/epl.py
```diff
@@ -183,8 +183,27 @@ class EPLMajorAxis(LensProfileBase):
...
     def __init__(self):
+
         super(EPLMajorAxis, self).__init__()
 
+    @staticmethod
+    def _regularize(alpha):
+
+        alpha_real, alpha_imag = alpha.real, alpha.imag
+
+        if isinstance(alpha_real, float) or isinstance(alpha_real, int):
```
sibirrer, line 194: can you simplify this with this functionality in numpy? https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html

lenstronomy/LensModel/Profiles/epl.py
```diff
@@ -214,7 +233,9 @@ def derivatives(self, x, y, b, t, q):
         alpha = 2/(1+q)*(b/R)**t*R_omega
 
         # return real and imaginary part
-        return alpha.real, alpha.imag
+        alpha_real, alpha_imag = self._regularize(alpha)
```
sibirrer, line 236: isn't this also an issue for derivatives and Hessians?

sibirrer (changes requested):
Could you simplify things with `numpy.nan_to_num`? I don't see a specific test case that checks the case of turning nan's into zero. Thanks!

dangilman:
I added the numpy function. Passing in (x, y) = (0, 0) creates np.nan values inside the class that get output as zeros, but I added another test function that tests the new regularize function explicitly. And it is also an issue for hessian, but hessian is computed by calling derivatives so it is taken care of

dangilman:
nevermind, was looking at the wrong hessian class.... let me fix it for the hessian also

dangilman:
not really sure what to do with infinities... I just default them to plus/minus 1000

lenstronomy/LensModel/Profiles/epl.py
```diff
@@ -183,8 +183,18 @@ class EPLMajorAxis(LensProfileBase):
...
 
     def __init__(self):
+
         super(EPLMajorAxis, self).__init__()
 
+    @staticmethod
+    def _regularize(value):
+
+        out = np.nan_to_num(value, posinf=1e+3, neginf=-1e+3)
+        if isinstance(value, float) or isinstance(value, int):
```
sibirrer, line 193: do you need this part?

lenstronomy/LensModel/Profiles/epl.py
```diff
@@ -183,8 +183,18 @@ class EPLMajorAxis(LensProfileBase):
     param_names = ['b', 't', 'q', 'center_x', 'center_y']
 
     def __init__(self):
+
         super(EPLMajorAxis, self).__init__()
 
+    @staticmethod
+    def _regularize(value):
+
+        out = np.nan_to_num(value, posinf=1e+3, neginf=-1e+3)
```
sibirrer, line 192: posinf and neginf do not need to be specified I think.

test/test_LensModel/test_Profiles/test_epl.py
```diff
@@ -100,6 +100,42 @@ def test_static(self):
...
+        npt.assert_almost_equal(f_yy, 0.)
+
+        x = 0.3
+        y = 0.1
+        R = np.sqrt(x ** 2 + y ** 2)
+        f_x, f_y = self.EPL.derivatives(x, y, phi_E, e1, e2, gamma)
+        npt.assert_almost_equal(f_x, phi_E * x/R)
+        npt.assert_almost_equal(f_y, phi_E * y/R)
+
+        value = np.nan
```
sibirrer, line 130: I was more thinking of putting in (0,0) and see whether you get regular behavior out

sibirrer (changes requested):
I don't think you need the array and float/int distinction with the new routine. This will simplify things quite a bit.

test/test_LensModel/test_Profiles/test_epl.py
dangilman: I do that on line 111

lenstronomy/LensModel/Profiles/epl.py
dangilman: I'll take them out

lenstronomy/LensModel/Profiles/epl.py
dangilman: It serves only to keep the input and output data types the same. If you supply a float, without that check you get back a numpy array. I can take it out if you want

sibirrer: I had guessed that the nan_to_num keeps the same in and output format. It's worth testing and if confirmed, you don't need this.

sibirrer (approved):

sibirrer:
Thank you!

PR #212, opened 2020-10-02
Flag to prevent second linear inversion and verbosity in image reconstruction class

aymgal:


sibirrer (approved):
Looks good, thanks you!

PR #213, opened 2020-10-09
Propagate missing shapelets parameter to Parameters class

aymgal:
This is to allow the use of shapelets in the mass within e.g. the FittingSequence.


sibirrer (approved):
Thank you for adding this! I have not used this feature since my work on RXJ1131 in 2017 and did not keep track of this parameter since.

PR #214, opened 2020-10-13
Optional shorter APIs

JelleAalbers:
This is a proposal to add an optional shorter API to lenstronomy, activated through the new `lenstronomy.short()` function:
```python
import lenstronomy as ls
ls.short()

lens_model = ls.LensModel.lens_model.LensModel(...)
source_model = ls.LightModel.light_model.LightModel(...)
image_model = ls.ImSim.image_model.ImageModel(...)
...
```

After you do `ls.short()`, tab-activated autocomplete will also work in jupyter, e.g. after you type `ls.` in the example above.

For more convenience, at the risk of losing clarity, there is also laconic mode:
```python
import lenstronomy as ls
ls.laconic()

lens_model = ls.LensModel_(...)
source_model = ls.LightModel_(...)
image_model = ls.ImageModel(...)
```
The underscores after `LensModel` and `LightModel` are needed to distinguish the classes from modules of the same name. You can also use the 'short' convention (`ls.LensModel.lens_model.LensModel`) in laconic mode.

Code written for `short` also runs under `laconic`; code written using the classic import-everything-explicitly API works under both.

Motivation
----------------
Lenstronomy has many submodules. To use a function or class, you have to know where it is defined, then import it from there. This is super clean and explicit, but a bit inconvenient for new users -- was `make_grid` defined in `lenstronomy.Util.image_util` or `lenstronomy.Util.util`? It also leads to baroque import headers, e.g.

```python
from lenstronomy.LightModel.light_model import LightModel
from lenstronomy.ImSim.image_model import ImageModel
import lenstronomy.Util.image_util as image_util
import lenstronomy.Util.util as util
from lenstronomy.Cosmo.lens_cosmo import LensCosmo
...
```

Other big python packages usually [import their own components in `__init__.py`](https://github.com/numpy/numpy/blob/master/numpy/__init__.py#L145), so you can write e.g. `np.asmatrix` without `from numpy.matrixlib.defmatrix import asmatrix`. 

We could do the same in lenstronomy, and maybe this PR is a stepping stone towards that. 

However, there might be a concern: `__init__.py` is loaded whenever anything from a package is imported. Perhaps someone is running lenstronomy without astropy (or some other dependency) installed, and is using only the parts of lenstronomy that do not import astropy. If we went the `__init__.py` route, lenstronomy would break for that user. This PR avoids that problem, at the cost of the extra `ls.short()` line.

If you're not worried about this, I'd be happy to change this PR to have short or laconic mode active by defaul, i.e. without the `ls.short()` call.

Details
---------
This adds:
  * `__all__` lists to lenstronomy files, listing the lenstronomy symbols defined in that module. This is a standard python package feature, useful even if you hate the short/laconic APIs. If a user does `from ... import *`, it ensures only symbols from `__all__` are exported into the user's namespace, and not local variables, numpy, os, etc. `__all__` can also be exploited by documentation generators and other tools.
    * For files with one function/class, the `__all__` is defined manually at the top of the file. For files with many functions/classes, an `@export` decorator is used. It's easier to remember to add `@export` to a function than to add its name to `__all__` far away at the top of a file.
  * The `shortcut` and `laconic` functions. These make the modules (shortcut) or just everything (laconic) available under lenstronomy's namespace, which is empty by default. The `__all__` lists are used to figure out which symbols to make accessible in laconic mode (so you don't have `ls.np`, `ls.os`, etc.).

Note:
  * The old import API still works fine. Nothing extra is loaded if you don't use short or laconic mode.
  * Symbols used twice in lenstronomy are skipped for laconic mode. You'll have to type out `ls.Data.psf.PSF` or `ls.GalKin.psf.PSF`; there is no ambiguous `ls.PSF`.
    * Most clashes are between lens and light models of the same name. Full list: ` PSF  Chameleon DoubleChameleon TripleChameleon Gaussian Hernquist Interpol NIE PJaffe PJaffe_Ellipse Sersic   ParticleSwarmOptimizer radial_profile`.
    * Four key classes clash with top-level modules (`Cosmo` `LensModel` `LightModel` and `PointSource`).  In laconic and short mode, `ls.LensModel` points to the module, and the class is accessible through its full path (`ls.LensModel.lens_model.LensModel`). In laconic mode, the class is also accessible through its name with an extra underscore (`ls.LensModel_`).

Footnotes
-------------

Also removed some unused imports:
  * Coordinates1D in Data.pixel_grid
  * param_util and mask_util in LensModel.lens_model_extensions
  * polar2cart in LensModel.Optimizer.penalities
  * numpy in LightModel.Profiles.profile_base
  * scipy.special in LensModel.Profile.sis
  
Minor remarks:
  * The loose multi-line string in GalKin.aperture would not be found by any auto documentation tool. Maybe you want to combine it with the Aperture class or __init__ docstring? There might have been other, similar occurrences.
  * In `GalKin.aperature_types`, I currently just `@export`'ed everything, but maybe you don't need the `shell_select`, `shell_ifu_select` etc. since these are accessible through the aperture type classes instead. You might want to fine-tune exports elsewhere too if the laconic API becomes popular in the future.

### DIFF
--- a/lenstronomy/Util/package_util.py
+++ b/lenstronomy/Util/package_util.py
@@ -1,0 +1,83 @@
+
+def exporter(export_self=False):
+    """Export utility, modified from https://stackoverflow.com/a/41895194
+    Returns export decorator, __all__ list
+    """
+    all_ = []
+    if export_self:
+        all_.append('exporter')
+
+    def decorator(obj):
+        all_.append(obj.__name__)
+        return obj
+
+    return decorator, all_
+
+
+export, __all__ = exporter(export_self=True)
+
+
+@export
+def laconic():
+    """Activate laconic / super-shortcut mode.
+
+    Usage:
+        import lenstronomy as ls
+        ls.laconic()
+
+        lens_model = ls.LensModel(...)
+        source_model = ls.LightModel(...)
+        image_model = ls.ImageModel(...)
+    """
+    short(_laconic=True)
+
+
+@export
+def short(_laconic=False):
+    """Activate shortcut mode.
+
+    Usage:
+        import lenstronomy as ls
+        ls.short()
+
+        lens_model = ls.LensModel.lens_model.LensModel(...)
+        source_model = ls.LightModel.light_model.LightModel(...)
+        image_model = ls.ImSim.image_model.ImageModel(...)
+    """
+    import pkgutil
+    import lenstronomy
+
+    to_add = dict()
+    all_modules = dict()
+
+    for loader, module_name, is_pkg in pkgutil.walk_packages(lenstronomy.__path__):
+        # Load the module
+        module = all_modules[module_name] = \
+            loader.find_module(module_name).load_module(module_name)
+
+        if '.' in module_name:
+            # Submodule, e.g. Data.psf
+            # Monkeypatch the parent module to make it accessible
+            fragments = module_name.split('.')
+            parent_module_name, child_name = '.'.join(fragments[:-1]), fragments[-1]
+            setattr(all_modules[parent_module_name], child_name, module)
+        else:
+            # Top-level module, e.g. Data: add as lenstronomy attribute
+            to_add[module_name] = module
+
+        if _laconic:
+            # If the module defines an __all__, load its symbols as well.
+            # (unlike import *, we do not just load everything if __all__ is missing)
+            if hasattr(module, '__all__'):
+                for symbol in module.__all__:
+                    if symbol in to_add:
+                        # Name clash! Do not add either this symbol
+                        # or the one it clashed with.
+                        del to_add[symbol]
+                    else:
+                        to_add[symbol] = getattr(module, symbol)
+
+    # Make symbols accessible under lenstronomy.[x]
+    for key, value in to_add.items():
+        setattr(lenstronomy, key, value)
+    lenstronomy.__all__ = to_add.keys()
--- a/lenstronomy/Util/package_util.py
+++ b/lenstronomy/Util/package_util.py
@@ -1,3 +1,5 @@
+import types
+
 
 def exporter(export_self=False):
     """Export utility, modified from https://stackoverflow.com/a/41895194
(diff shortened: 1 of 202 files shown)


lenstronomy/Util/package_util.py
```diff
@@ -0,0 +1,83 @@
...
+        return obj
+
+    return decorator, all_
+
+
+export, __all__ = exporter(export_self=True)
+
+
+@export
+def laconic():
```
sibirrer, line 21: This function does not have a unit test.

lenstronomy/Util/package_util.py
```diff
@@ -0,0 +1,83 @@
...
+
+        lens_model = ls.LensModel(...)
+        source_model = ls.LightModel(...)
+        image_model = ls.ImageModel(...)
+    """
+    short(_laconic=True)
+
+
+@export
+def short(_laconic=False):
```
sibirrer, line 36: This function does not have a unit test.

sibirrer (changes requested):
Thanks @JelleAalbers ! This is a nice addition. I indeed would not like to have it being hard coded in the `__init__.py` file as some imports are not required by many users (such as nested sampler etc) and to keep the installation light-weighted. Having the full import statements in the notebooks and documentation provides (in my view) good access to the actual source code for the user (with the price of tedious import statements). I have two requests before merging: 1. add unit tests for the newly added features (you might also need to add `pkgutil` to the requirements). 2. Please add your name to the contributors list in the `AUTHORS.rst` file

JelleAalbers:
Thanks for the review Simon; glad you like the idea! I added unit tests now. `pkgutil` is a standard library package (since python 2.3), so we can't add it to requirements.txt without causing problems (see e.g. https://github.com/scikit-multilearn/scikit-multilearn/issues/72 and https://pytosquatting.overtag.dk/).

I made one more change to laconic mode. Previously, key classes like `LensModel` and `LightModel` were not put into the lenstronomy namespace by laconic, since their 'slot' was taken by modules with the same name. You could only access them through the 'short' convention, e.g. `lenstronomy.LensModel.lens_model.LensModel`. Now, you can _also_ access them as `lenstronomy.LensModel_`. I updated the example above to reflect this. It's not very pretty, but I guess the name clashes are here to stay, and if you can't access the main classes through a shortcut, there wouldn't be much point to laconic mode.

sibirrer (approved):

sibirrer:
I see, as the folder has the same name as the class within the folder that makes it ambiguous. Thank you very much!

PR #215, opened 2020-10-28
Fix multiband image reconstruction for all bands_compute options

aymgal:
When `bands_compute` was e.g. `[False, True]` for a 2-bands setup, there were index issues because the `image_linear_solve` method does not always return lists with same length as the number of bands. The fix is that for bands that are not "computed", now a `None` is put in the list instead. It also allows to simplify the code in the constructor of `MultiBandImageReconstruction`.
Please ask if the `None` choice is not the most suitable.


sibirrer:
Thanks @aymgal ! I think this convention with returning None's but keeping the list length is more intuitive. I checked the code and it looks good. Somehow travis-ci is not working...I'll check on that

sibirrer (approved):
Test ratio only dropped due to removal of code lines.

PR #216, opened 2020-11-17
Prevent copy of a possibly large array in memory

aymgal:
For instance, when using a high number of shapelet basis functions to model a large number of data pixels, the `nan_to_num` function implicated the copy of a large array in memory. This PR intends to limit the number of array copies.

### DIFF
--- a/lenstronomy/ImSim/image_linear_solve.py
+++ b/lenstronomy/ImSim/image_linear_solve.py
@@ -273,20 +273,20 @@ class ImageLinearFit(ImageModel):
             image = source_light_response[i]
             image *= extinction
             image = self.ImageNumerics.re_size_convolve(image, unconvolved=unconvolved)
-            A[n, :] = self.image2array_masked(image)
+            A[n, :] = np.nan_to_num(self.image2array_masked(image), copy=False)
             n += 1
         # response of lens light profile
         for i in range(0, n_lens_light):
             image = lens_light_response[i]
             image = self.ImageNumerics.re_size_convolve(image, unconvolved=unconvolved)
-            A[n, :] = self.image2array_masked(image)
+            A[n, :] = np.nan_to_num(self.image2array_masked(image), copy=False)
             n += 1
         # response of point sources
         for i in range(0, n_points):
             image = self.ImageNumerics.point_source_rendering(ra_pos[i], dec_pos[i], amp[i])
-            A[n, :] = self.image2array_masked(image)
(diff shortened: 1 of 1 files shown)


lenstronomy/ImSim/image_linear_solve.py
```diff
@@ -273,20 +273,20 @@ def _linear_response_matrix(self, kwargs_lens, kwargs_source, kwargs_lens_light,
             image = source_light_response[i]
             image *= extinction
             image = self.ImageNumerics.re_size_convolve(image, unconvolved=unconvolved)
-            A[n, :] = self.image2array_masked(image)
+            A[n, :] = np.nan_to_num(self.image2array_masked(image), copy=False)
```
sibirrer, line 276: is this nan_to_num() statement replacing another statement? I don't see another file changed in lenstronomy, or is it in slitronomy?

sibirrer (commented):
I see your point in carefully usage of copying and the use of nan_to_num(), thanks for identification of this bottle neck for large shapeless basis sets.
In the current PR you effectively add a nan_to_num statement without a replacement of a statement that does copying, or do I miss something?

lenstronomy/ImSim/image_linear_solve.py
```diff
@@ -273,20 +273,20 @@ def _linear_response_matrix(self, kwargs_lens, kwargs_source, kwargs_lens_light,
...
             n += 1
         # response of point sources
         for i in range(0, n_points):
             image = self.ImageNumerics.point_source_rendering(ra_pos[i], dec_pos[i], amp[i])
-            A[n, :] = self.image2array_masked(image)
+            A[n, :] = np.nan_to_num(self.image2array_masked(image), copy=False)
             n += 1
-        return np.nan_to_num(A)
```
aymgal, line 289: I removed this line, and put the nan_to_num for each `image` instead, so the functions is applied to smaller arrays as well.

lenstronomy/ImSim/image_linear_solve.py
aymgal: No it is not related ton SLITronomy (see the line below)

lenstronomy/ImSim/image_linear_solve.py
sibirrer: ah I see, thanks. I was not aware of the copy issue with nan_to_num(), I might assess these function calls also in other places.

PR #220, opened 2020-12-14
Fix for non-square source images in Interpol LightModel

JelleAalbers:
I got an error trying to use the Interpol LightModel with a non-square image as the `image` kwarg: `ValueError: x dimension of z must have same number of elements as x`, from inside scipy's `RectBivariateSpline`.

The cause is at https://github.com/sibirrer/lenstronomy/blob/master/lenstronomy/LightModel/Profiles/interpolation.py#L63, where the `x_grid` and `y_grid` arguments are switched. The first argument to `RectBivariateSpline` has to match the first dimension in the image array, whether one calls it x or y. 

This:
  * changes Interpol's unit test to use a (20, 5) instead of a (10, 10) image, which should be a strictly harder test case;
  * switches x_grid and y_grid arguments in the line above. Without this fix, the new test fails.

This only specifies a valid gridsize to the interpolator, nothing is changed in the image handling. In particular, the x and y are still 'flipped' when calling the interpolator in the line directly below https://github.com/sibirrer/lenstronomy/blob/master/lenstronomy/LightModel/Profiles/interpolation.py#L64. Since it is not changed, I did not think about this or check if it is correct -- but I assume this 'flip' is intentional, to transpose between computer science and physics display conventions? 

Just let me know if I misunderstood something or if you'd prefer a different solution.

### DIFF
--- a/lenstronomy/Data/psf.py
+++ b/lenstronomy/Data/psf.py
@@ -67,7 +67,7 @@ class PSF(object):
     def kernel_point_source(self):
         if not hasattr(self, '_kernel_point_source'):
             if self.psf_type == 'GAUSSIAN':
-                kernel_numPix = self._truncation * self._fwhm / self._pixel_size
+                kernel_numPix = round(self._truncation * self._fwhm / self._pixel_size)
                 if kernel_numPix % 2 == 0:
                     kernel_numPix += 1
                 self._kernel_point_source = kernel_util.kernel_gaussian(kernel_numPix, self._pixel_size, self._fwhm)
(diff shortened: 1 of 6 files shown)


sibirrer (changes requested):
Thanks @JelleAalbers ! Currently indeed, there was implicitly a rectangular grid required. For the tests, I suggest to extend the tests in https://github.com/sibirrer/lenstronomy/blob/master/test/test_LightModel/test_Profiles/test_interpolation.py 

There is at least one test that shifts the coordinates in a specific direction requiring that the 'center_x' shift corresponds to the shift in the first coordinate entry. Repeating this test with a non-square grid might be good. The zero point of the rotation angle itself was not properly defined indeed (meaning arbitrary), so this change is compatible. The current routine util.make_grid() is rectangular. You can either code up a non-rectangular definition or chop out part of the grid to make it non-rectangular.

Let me know if you have questions and thank you very much!

sibirrer (approved):

sibirrer:
Thanks @JelleAalbers !

JelleAalbers:
Thanks Simon! Just for the record, the final changes I added:
  * Run the function tests in `test_LightModel/test_Profiles/test_interpolation.py` on both a square and non-square grid.
  * Non-square grid support for `make_grid` + tests. You can now specify e.g. `numPix=(20, 10)`.
  * Specifying a non-round float for numPix is now an error (12 and 12.0 would be OK, 12.0000003 is not). 
  * Fix one instance in `Data/psf.py` where non-round numPixes were occurring and causing test failures. `round`ing got the tests to pass, so I assume/hope this is the intended behavior..
  * Add a few comments.

PR #225, opened 2021-01-04
New lens model : elliptical density slice

LyneVdV:
This lens model has a constant surface density inside an ellipse and has 0. outside.
This model is based on Schramm 1994 (A toolbox for general elliptical gravitational lenses) : the general idea of the paper is to discretize a surface density of a galaxy into a sum of elliptical slices of denstity with different spatial extensions and orientations. The potential and deflections of one slice are known analytically (see Schramm 1994). This new lens model, called ElliSLICE, gathers the lensing quantites associated to one elliptical slice with constant surface density. 
It can be used to create complex models of galaxies with isophotal twists, change in ellipticity etc by assembling several elliptical slices with different intrinsic parameters. It can also be used in a fitting process but I have not tried it.
I tried to be explicit in the class definition and in the code. I also created and run different tests. Let me know if I have to change anything !


lenstronomy/LensModel/Profiles/elliptical_density_slice.py
```diff
@@ -0,0 +1,261 @@
...
+__all__ = ['ElliSLICE']
+
+class ElliSLICE (LensProfileBase):
+    """
+    This class computes the lensing quantities for an elliptical slice of constant density.
+    Based on Schramm 1994 https://ui.adsabs.harvard.edu/abs/1994A%26A...284...44S/abstract
+
+    Computes the lensing quantities of an elliptical slice with semi major axis 'a' and
+    semi minor axis 'b', centered on 'center_x' and 'center_y', oriented with an angle 'psi'
+    in radian, and with constant surface mass density 'sigma_0'
```
sibirrer, line 17: Can you add a mathematical formula directly to this documentation? I know it's not been done for all the old models but I would like to change that in the long run. One example could be found here: https://github.com/sibirrer/lenstronomy/blob/master/lenstronomy/LensModel/Profiles/p_jaffe.py

lenstronomy/LensModel/Profiles/elliptical_density_slice.py
```diff
@@ -0,0 +1,261 @@
...
+            zb_plus_eps = z_plus_eps.conjugate()
+            I_out_minus = 2 * a * b / f2 * (zb_minus_eps * e2ipsi - eipsi * self.sign(zb_minus_eps * eipsi)
+                                            * c.sqrt(zb_minus_eps ** 2 * e2ipsi - f2)) * sig_0
+            I_out_plus = 2 * a * b / f2 * (zb_plus_eps * e2ipsi - eipsi * self.sign(zb_plus_eps * eipsi)
+                                           * c.sqrt(zb_plus_eps ** 2 * e2ipsi - f2)) * sig_0
+            I_out_mid = 2 * a * b / f2 * (zb * e2ipsi - eipsi * self.sign(zb * eipsi)
+                                          * c.sqrt(zb ** 2 * e2ipsi - f2)) * sig_0
+            I_out_real = np.median([I_out_minus.real,I_out_plus.real,I_out_mid.real])
+            I_out_imag = np.median([I_out_minus.imag, I_out_plus.imag, I_out_mid.imag])
+        else :
```
sibirrer, line 178: these lines of code are note testes. See: https://github.com/sibirrer/lenstronomy/blob/master/lenstronomy/LensModel/Profiles/p_jaffe.py

test/test_LensModel/test_Profiles/test_elliptical_density_slice.py
```diff
@@ -0,0 +1,81 @@
...
+
+from lenstronomy.LensModel.Profiles.elliptical_density_slice import ElliSLICE
+
+import numpy as np
+import pytest
+import numpy.testing as npt
+
+class TestElliSLICE(object):
+    """
+    tests tthe elliptical slice lens model
```
sibirrer, line 12: minor typo :)

test/test_LensModel/test_Profiles/test_elliptical_density_slice.py
```diff
@@ -0,0 +1,81 @@
...
+        x = np.array([0])
+        y = np.array([0])
+        values = self.ElliSLICE.function(x, y, a, b, psi, sigma_0)
+        npt.assert_almost_equal(values[0], 4.054651081,decimal=5)
+
+        x = np.array([np.sqrt(3), np.sqrt(3)+0.000000001, np.sqrt(3)-0.000000001])
+        y = np.array([1, 1.000000001, 0.999999999])
+        values = self.ElliSLICE.function(x, y,  a, b, psi, sigma_0)
+        npt.assert_almost_equal(values[0], values[1], decimal=5)
+        npt.assert_almost_equal(values[1], values[2], decimal=5)
```
sibirrer, line 35: Is there a way to test a correspondence with a different profile (such as a point source outside of the elliptical shell)? Just wondering, and if so, might be good to demonstrate it in the tests.

test/test_LensModel/test_numeric_lens_differentials.py
```diff
@@ -305,6 +305,10 @@ def test_multipole(self):
         kwargs = {'m': 3, 'a_m': 0.07, 'phi_m': 0., 'center_x': -0.01, 'center_y': -0.5}
         lens_model = ['MULTIPOLE']
         self.assert_differentials(lens_model, kwargs, potential=True)
+    def test_elli_slice(self):
```
sibirrer, line 308: can you add a spacing between the test functions?

test/test_LensModel/test_numeric_lens_differentials.py
```diff
@@ -305,6 +305,10 @@ def test_multipole(self):
...
+        kwargs = {'a':1., 'b':2., 'psi':30*np.pi/180., 'sigma_0':5., 'center_x':1., 'center_y':2.}
```
sibirrer, line 309: the differential test is made at a default (1,1) location (maybe I need to change that more broadly in general). For your case, the differentials are only tested inside of the ellipse. Can you add a second scenario where the differential is performed outside? This is also reflected in the test coverage.

sibirrer (changes requested):
Hi Lyne, thank you very much for the PR and this addition! Overall looks very good. I have some requests for testing and documentation. Other than that happy to merge. Let me know if you have questions.

lenstronomy/LensModel/Profiles/elliptical_density_slice.py
LyneVdV: Really nice idea to have a mathematical formula for all models !

lenstronomy/LensModel/Profiles/elliptical_density_slice.py
LyneVdV: Solved !

test/test_LensModel/test_Profiles/test_elliptical_density_slice.py
LyneVdV: Not exactly : outside the elliptical slice, the potential symmetry is confocal and is getting rounder and rounder as you move away from the slice and thus _tend_ to mimic a point mass model at large distance. So there is no exact correspondence to test. If you really want to, I can try to find the conversion factor between my model and the point mass model and see at which distance I have to go before having a correspondence at x decimals. It will be a fine tuning depending on the slice parameters so I don't think it is really needed here.

test/test_LensModel/test_numeric_lens_differentials.py
LyneVdV: You have already broaden it with x = np.linspace(start=0.1, stop=5.5, num=10) and y = np.zeros_like(x) ! I changed the parameters and added a line so that it can go inside and outside the slice and also on and off the ellipse axis.

LyneVdV:
Hi Simon ! 
I implemented the changes and even found a typo that was slowing down the code.
For the comparison with the point_mass model, let me know what you think.
Have a nice day !

test/test_LensModel/test_Profiles/test_elliptical_density_slice.py
sibirrer: I think it's fine not having it in the code. The equation in the documentation should be sufficient for people if there is need to find conversion factors, and can be implemented when required.

sibirrer:
Thank you very much! Looks all good and thanks for the updated documentation! There is one minor thing: There are two lines not yet tested: https://coveralls.io/builds/36090719/source?filename=lenstronomy/LensModel/Profiles/elliptical_density_slice.py
They are for single-valued input of the potential and deflection angles for outside the ellipse. Sorry to be petty here, just want to get to 100% test coverage of as many modules as possible

LyneVdV:
Done ! Don't worry, your are not being petty, just conscientious ;)

sibirrer:
Perfect, thank you! :)

PR #227, opened 2021-01-05
Better lens equation solving

ewoudwempe:
This adds a jit compiler to the local minima finder. Sped up lens equation solving for me (quite significantly if I recall correctly). 
Also simplifies the indices that are checked (to all pixels with distance <= sqrt(5)), instead of some asymmetrical cross-like structure with gaps). Probably this could be optimized a bit further with a lot of testing. I'm curious if the tests pass, for me it has been quite reliable.

### DIFF
--- a/lenstronomy/Util/util.py
+++ b/lenstronomy/Util/util.py
@@ -7,6 +7,7 @@ this file contains standard routines
 import numpy as np
 import mpmath
 import itertools
+from numba import njit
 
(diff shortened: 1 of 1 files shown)


sibirrer:
Congrats, tests are passing! Test coverage only decreased due to @jit coverage. This is one of the more impactful changes in the recent development and I'll keep an eye on how performance and reliability is changing. Thank you very much!

PR #228, opened 2021-01-05
Changed the truncated normal to a proper truncnorm distribution.

ewoudwempe:
Instead of sampling from a Gaussian, and then clipping the values out of
bounds, sample from a truncated Gaussian from the start. This avoids an
effective delta-peak in the PDF at the bounds.

### DIFF
--- a/lenstronomy/Util/sampling_util.py
+++ b/lenstronomy/Util/sampling_util.py
@@ -1,7 +1,7 @@
 __author__ = 'aymgal'
 
 import numpy as np
-from scipy import special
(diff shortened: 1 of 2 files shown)


sibirrer:
The tests are failing due to a test of a removed function (no surprise :) ). You can either re-insert the function or remove the test of the function. There is no use of this function and I am not concerned of removing it.

sibirrer:
Perfect! Minor decrease of coverage is due to removal of redundant code and not addition of untested code. Thank you!

PR #229, opened 2021-01-05
Better caustics

ewoudwempe:
This improves the caustics calculation by removing a matplotlib call (that produced an empty plot) and replacing it with a scikit-image based contour finder.
Also, the inverse magnification is used for the contour finding, because 1/mag is approximately linear near critical curves, while the magnification diverges. That smoothens out things a lot, at least when I tested things for some SIE & SPEMD lenses. Maybe this is not as great for some other lens models, if the critical curve does not have a diverging magnification (know any good examples?), so perhaps an option to turn this on or off is good?

### DIFF
--- a/lenstronomy/LensModel/lens_model_extensions.py
+++ b/lenstronomy/LensModel/lens_model_extensions.py
@@ -1,5 +1,6 @@
 import numpy as np
 import lenstronomy.Util.util as util
+from skimage.measure import find_contours
 
 __all__ = ['LensModelExtensions']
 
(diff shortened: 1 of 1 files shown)


sibirrer:
The current test in lenstronomy is very specific to the former routine (which is bad testing). I'll change the test function and push it to master. I hope, if you then merge the current implementation of the caustics with the new test function that it will pass.

sibirrer:
I updated the tests (see commit linked to this PR). Can you pull and push on your side a small update to trigger travis-CI if it does not automatically re-run the tests? Thanks!

ewoudwempe:
Done, thanks for making the new test!

sibirrer:
Perfect! Test coverage decrease only due to fewer lines of code required

ewoudwempe:
Minor nitpick: the test you made also passed for the previous caustics calculation. When I did the test myself, it passed with 170 minimum absolute magnification for the old one (larger than the required 100), and after the modification it passed with minimum absolute magnification of 113386, so there is a good improvement at least :)

sibirrer:
Ha! That is a substantial improvement! I just updated the test requirements to make sure they are matched by anyone who wants to change the code and reference to this PR :)

PR #230, opened 2021-01-05
Allow specifiying constraints as a dictionary if the keys are not equal

ewoudwempe:
Useful e.g. to set the shear center at the lens center, not previously possible because 'ra_0' is not the same key as 'center_x', and the constraint function only set the same values.
If there was already a standard way of doing this, or if this belongs somewhere else, feel free to point me in the right direction and close this pull request.

### DIFF
--- a/lenstronomy/Sampling/parameters.py
+++ b/lenstronomy/Sampling/parameters.py
@@ -439,12 +439,19 @@ class Param(object):
         :param kwargs_list_1: list of keyword arguments
         :param kwargs_list_2: list of keyword arguments
         :param joint_setting_list: [[i_1, k_2, ['param_name1', 'param_name2', ...]], [...], ...]
+                                     or: [[i_1, k_2, {'param_old1': 'param_new1', 'ra_0': 'center_x'}], [...], ...]
         :return: updated kwargs_list_2 with arguments from kwargs_list_1 as defined in joint_setting_list
         """
         for setting in joint_setting_list:
             i_1, k_2, param_list = setting
-            for param_name in param_list:
-                kwargs_list_2[k_2][param_name] = kwargs_list_1[i_1][param_name]
+            if type(param_list) == list:
+                for param_name in param_list:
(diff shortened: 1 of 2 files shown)


lenstronomy/Sampling/parameters.py
```diff
@@ -439,12 +439,19 @@ def _update_joint_param(kwargs_list_1, kwargs_list_2, joint_setting_list):
...
-            for param_name in param_list:
-                kwargs_list_2[k_2][param_name] = kwargs_list_1[i_1][param_name]
+            if type(param_list) == list:
+                for param_name in param_list:
+                    kwargs_list_2[k_2][param_name] = kwargs_list_1[i_1][param_name]
+            elif type(param_list) == dict:
+                for param_to, param_from in param_list.items():
+                    kwargs_list_2[k_2][param_to] = kwargs_list_1[i_1][param_from]
+            else:
+                raise TypeError("Bad format for settings list")
```
sibirrer, line 454: Can you also let the setting be printed such that the user sees it? You can do it with "here is some text and string of the parameter is %s " %(param_list)

sibirrer (changes requested):
This is generally a nice feature and extends the capabilities of merging two parameters with different names! 

Side note: For the shear components however, I would not recommend this. The shear distortion is fully defined by gamma1 and gamma2. The center of the Taylor expansion effectively adds a translation which is not observable and does not change the time-delays neither. So for the shear, I would fix ra_0 and dec_0 to an 'arbitrary' value and not try to sample them (nor link it to another parameter.

Can you add a simple test scenario of this new convention here: https://github.com/sibirrer/lenstronomy/blob/master/test/test_Sampling/test_parameters.py and check that the parameters are over-written in this new convention?

Thank you very much!

sibirrer:
Perfect! The raise statement I can test later on when cleaning up other raise tests too. Other than that all lines of code are tested.

PR #231, opened 2021-01-13
new routine for finite source size magnification computations 

dangilman:
I added a new method in LensModelExtensions that computes magnifications with finite source sizes much faster than the existing magnification_finite routine

### DIFF
--- a/lenstronomy/LensModel/lens_model_extensions.py
+++ b/lenstronomy/LensModel/lens_model_extensions.py
@@ -113,7 +113,10 @@ class LensModelExtensions(object):
 
                 diff = abs(new_magnification - magnification_current)/magnification_current
 
-                if diff < tol and new_magnification > minimum_magnification:
+                # the sqrt(2) will allow this algorithm to eventully fill up the entire rectangular aperture
+                if r_max > np.sqrt(2) * grid_radius_arcsec:
+                    break
+                elif diff < tol and new_magnification > minimum_magnification:
                     break
                 else:
                     r_min += step
--- a/lenstronomy/LensModel/lens_model_extensions.py
+++ b/lenstronomy/LensModel/lens_model_extensions.py
@@ -3,6 +3,7 @@ import lenstronomy.Util.util as util
 from skimage.measure import find_contours
 from lenstronomy.Cosmo.background import Background
 from lenstronomy.LightModel.light_model import LightModel
+from astropy.cosmology import default_cosmology
 
 __all__ = ['LensModelExtensions']
 
@@ -12,8 +13,8 @@ class LensModelExtensions(object):
     class with extension routines not part of the LensModel core routines
     """
     def __init__(self, lensModel):
-        """
 
+        """
         :param lensModel: instance of the LensModel() class, or with same functionalities.
         In particular, the following definitions are required to execute all functionalities presented in this class:
         def ray_shooting()
@@ -25,31 +26,39 @@ class LensModelExtensions(object):
         """
         self._lensModel = lensModel
 
-    def magnification_finite_adaptive(self, x_pos, y_pos, source_x, source_y, kwargs_lens,
-                                      source_fwhm_pasec, z_source,
-                                      astropy=None, grid_resolution=None, grid_radius_arcsec=None, tol=0.001,
-                                      grid_res_scale=0.0004, grid_size_scale=0.004, step_size=0.05):
+    def magnification_finite_adaptive(self, x_image, y_image, source_x, source_y, kwargs_lens,
+                                      source_fwhm_parsec, z_source,
+                                      cosmo=None, grid_resolution=None, grid_radius_arcsec=None, axis_ratio=0.5,
+                                      tol=0.001, grid_res_scale=0.0004, grid_size_scale=0.005, step_size=0.05):
         """
         This method computes image magnifications with a finite-size background source assuming a Gaussian
-        source light profile. This method can be much faster that magnification_finite for lens models with many
-        deflectors. This is because most pixels in a rectangular window around a lensed image will contain zero flux.
-        Rather than ray tracing through a rectangular grid, this routine ray traces through a circular
-        aperture that becomes progressively larger until all of the light is captured and the magnification converges
-        to a fixed value.
+        source light profile. It can be much faster that magnification_finite for lens models with many
+        deflectors and a relatively compact source. This is because most pixels in a rectangular window around a lensed
+        image of a compact source will contain zero flux, and therefore don't contribute to the image brightness.
+
+        Rather than ray tracing through a rectangular grid, this routine accelerates the computation of image
+        magnifications with finite-size sources by ray tracing through an elliptical aperture oriented such that
+        it resembles the surface brightness of the lensed image itself. The aperture size is initially quite small,
+        and increases in size until the flux inside of it (and hence the magnification) converges. The orientation of
+        the elliptical aperture is computed from the magnification tensor at the image coordinate.
+
+        If for whatever reason you prefer a circular aperture to the elliptical approximation using the hessian eigenvectors,
+        you can just set axis_ratio = 1.
 
         The default settings for the grid resolution and ray tracing window size work well for sources with fwhm between
         0.5 - 100 pc.
 
-        :param x_pos: a list or array of x coordinates
-        :param y_pos: a list or array of y coordinates
+        :param x_image: a list or array of x coordinates [units arcsec]
+        :param y_image: a list or array of y coordinates [units arcsec]
         :param kwargs_lens: keyword arguments for the lens model
-        :param source_fwhm_pasec: the size of the background source in parsecs
+        :param source_fwhm_parsec: the size of the background source [units parsec]
         :param z_source: the source redshift
-        :param astropy: (optional) an instance of astropy; if not specified, a default cosmology will be used
+        :param cosmo: (optional) an instance of astropy.cosmology; if not specified, a default cosmology will be used
         :param grid_resolution: the grid resolution in units arcsec/pixel; if not specified, an appropriate value will
         be estimated from the source size
         :param grid_radius_arcsec: (optional) the size of the ray tracing region; if not specified, an appropriate value
         will be estimated from the source size
+        :param axis_ratio: the axis ratio of the ellipse used for ray tracing
         :param tol: tolerance for convergence in the magnification
         :param grid_res_scale: sets the grid resolution
         :param grid_size_scale: determines the size of the ray tracing window
@@ -57,23 +66,22 @@ class LensModelExtensions(object):
         :return: an array of image magnifications
         """
 
-        if astropy is None:
-            background_cosmology = Background()
-            astropy = background_cosmology.cosmo
+        if cosmo is None:
+            cosmo = default_cosmology.get()
 
-        # These default settings determined by guess and check, seem adequate for sources with size 0.1 - 100 pc
+        # These default settings determined by guess and check seem adequate for sources with size 0.1 - 100 pc
         if grid_resolution is None:
             ref = 10.
             power = 1
-            grid_resolution = grid_res_scale * (source_fwhm_pasec / ref) ** power
+            grid_resolution = grid_res_scale * (source_fwhm_parsec / ref) ** power
         if grid_radius_arcsec is None:
             size_0 = 1.
             power = 1.
-            grid_radius_arcsec = grid_size_scale * (source_fwhm_pasec / size_0) ** power
+            grid_radius_arcsec = grid_size_scale * (source_fwhm_parsec / size_0) ** power
 
-        pc_per_arcsec = 1000 / astropy.arcsec_per_kpc_proper(z_source).value
+        pc_per_arcsec = 1000 / cosmo.arcsec_per_kpc_proper(z_source).value
         # factor of 2.355 for FWHM to variance
-        source_sigma = source_fwhm_pasec / pc_per_arcsec / 2.355
+        source_sigma = source_fwhm_parsec / pc_per_arcsec / 2.355
         kwargs_source = [{'amp': 1., 'center_x': source_x, 'center_y': source_y, 'sigma': source_sigma}]
         source_model = LightModel(['GAUSSIAN'])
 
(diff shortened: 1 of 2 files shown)


lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -23,6 +25,143 @@ def hessian()
...
+                    r_min += step
+                    r_max += step
+                    magnification_current = new_magnification
+
+            magnifications.append(new_magnification)
+
+        return np.array(magnifications)
+
+    @staticmethod
+    def _iterate_adaptive(flux_array, x_image, y_image, grid_x, grid_y, grid_r, r_min, r_max,
```
sibirrer, line 131: the naming of this function is a bit misleading. This routine does not do any 'iterative' or 'adaptive' calls

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -23,6 +25,143 @@ def hessian()
...
+                          lensModel, kwargs_lens, source_model, kwargs_source):
+        """
+        performs a signle iteration of the ray tracing computations performed in magnification_finite_adaptive
```
sibirrer, line 134: here I would specify better what this function does (you don't necessarily need to mention the iterative calling of this function in another definition, as this is part of the other definition)

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -23,6 +25,143 @@ def hessian()
...
+
+        The default settings for the grid resolution and ray tracing window size work well for sources with fwhm between
+        0.5 - 100 pc.
+
+        :param x_pos: a list or array of x coordinates
+        :param y_pos: a list or array of y coordinates
+        :param kwargs_lens: keyword arguments for the lens model
+        :param source_fwhm_pasec: the size of the background source in parsecs
+        :param z_source: the source redshift
+        :param astropy: (optional) an instance of astropy; if not specified, a default cosmology will be used
```
sibirrer, line 48: to be consistent with the rest of lenstronomy, I would name this variable 'cosmo', in fact it should be an astropy.cosmology instance, and not an astropy instance

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -23,6 +25,143 @@ def hessian()
...
+        will be estimated from the source size
+        :param tol: tolerance for convergence in the magnification
+        :param grid_res_scale: sets the grid resolution
+        :param grid_size_scale: determines the size of the ray tracing window
+        :param step_size: sets the increment for the successively larger ray tracing windows
+        :return: an array of image magnifications
+        """
+
+        if astropy is None:
+            background_cosmology = Background()
```
sibirrer, line 61: I would directly make the instance through astropy rather than going through the Background() class to extract the cosmology from there. Makes it more compatible with astropy directly.

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -23,6 +25,143 @@ def hessian()
...
+        source light profile. This method can be much faster that magnification_finite for lens models with many
+        deflectors. This is because most pixels in a rectangular window around a lensed image will contain zero flux.
+        Rather than ray tracing through a rectangular grid, this routine ray traces through a circular
+        aperture that becomes progressively larger until all of the light is captured and the magnification converges
+        to a fixed value.
+
+        The default settings for the grid resolution and ray tracing window size work well for sources with fwhm between
+        0.5 - 100 pc.
+
+        :param x_pos: a list or array of x coordinates
```
sibirrer, line 43: in this function, you are using units (translating angle to physical scales with astropy). Can you add the units (arc seconds or pc) in the input requirements for the different parameters?

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -23,6 +25,143 @@ def hessian()
...
+            grid_x, grid_y = np.meshgrid(_grid_x, _grid_y)
+            grid_r = np.hypot(grid_x, grid_y).ravel()
+            grid_x, grid_y = grid_x.ravel(), grid_y.ravel()
+
+            flux_array = np.zeros_like(grid_x)
+
+            step = step_size * grid_radius_arcsec
+            r_min = 0.
+            r_max = r_min + step
+            flux_array = self._iterate_adaptive(flux_array, xi, yi, grid_x, grid_y, grid_r,
```
sibirrer, line 97: these lines of code are kind of duplicated in the while loop. Is it possible that also the first calculation is already part of the while loop to make it more compact?

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -23,6 +25,143 @@ def hessian()
         """
         self._lensModel = lensModel
 
+    def magnification_finite_adaptive(self, x_pos, y_pos, source_x, source_y, kwargs_lens,
+                                      source_fwhm_pasec, z_source,
+                                      astropy=None, grid_resolution=None, grid_radius_arcsec=None, tol=0.001,
+                                      grid_res_scale=0.0004, grid_size_scale=0.004, step_size=0.05):
+        """
+        This method computes image magnifications with a finite-size background source assuming a Gaussian
```
sibirrer, line 33: can you also describe a bit more the iterative procedure in this scheme?

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -23,6 +25,143 @@ def hessian()
...
+        source light profile. This method can be much faster that magnification_finite for lens models with many
```
sibirrer, line 34: Here you can also specify that this effectively a speed-up in general when the ray-tracing is the bottle neck and most source positions are zeros.

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -23,6 +25,143 @@ def hessian()
...
+            grid_radius_arcsec = grid_size_scale * (source_fwhm_pasec / size_0) ** power
+
+        pc_per_arcsec = 1000 / astropy.arcsec_per_kpc_proper(z_source).value
+        # factor of 2.355 for FWHM to variance
+        source_sigma = source_fwhm_pasec / pc_per_arcsec / 2.355
+        kwargs_source = [{'amp': 1., 'center_x': source_x, 'center_y': source_y, 'sigma': source_sigma}]
+        source_model = LightModel(['GAUSSIAN'])
+
+        npix = int(2 * grid_radius_arcsec / grid_resolution)
+        _grid_x = np.linspace(-grid_radius_arcsec, grid_radius_arcsec, npix)
```
sibirrer, line 81: could there be a further speed-up if the grid is not rectangular and already tight to the magnification tensor of the lensing feature? Or do you want to deliberately not have it be that way to have better non-local distortion control? Might be also worth mentioning in the description (sorry to make you write a lot of text)

test/test_LensModel/test_lens_model_extensions.py
```diff
@@ -75,12 +76,63 @@ def test_elliptical_ray_trace(self):
...
+                       {'gamma1': 0.01, 'gamma2': 0.03}]
+
+        lensmodel = LensModel(lens_model_list)
+        extension = LensModelExtensions(lensmodel)
+        solver = LensEquationSolver(lensmodel)
+        source_x, source_y = 0.07, 0.03
+        x_image, y_image = solver.findBrightImage(source_x, source_y, kwargs_lens)
+
+        source_fwhm_parsec = 60.
+        background = Background()
```
sibirrer, line 98: can you make the test with a specific astropy.cosmology instance? In this form, if the default cosmology is changing (not in control of lenstronomy) this tests could lead to different results.

test/test_LensModel/test_lens_model_extensions.py
```diff
@@ -75,12 +76,63 @@ def test_elliptical_ray_trace(self):
...
+
+        mag_square_grid = extension.magnification_finite(x_image, y_image, kwargs_lens, source_sigma=source_sigma,
+                                                         grid_number=5000, window_size=0.45)
+
+        mag_adaptive_grid = extension.magnification_finite_adaptive(x_image, y_image, source_x, source_y, kwargs_lens, source_fwhm_parsec,
+                                                                    z_source)
+
+        mag_square_grid *= max(mag_square_grid) ** -1
+        mag_adaptive_grid *= max(mag_adaptive_grid) ** -1
+        npt.assert_almost_equal(mag_square_grid, mag_adaptive_grid, 3)
```
sibirrer, line 111: here you test against magnification_finite(), which is great! Can you also test against the expected magnification, given the knowledge of the (local) magnification and the intrinsic source brightness? (I know it's not perfect due to the extended size of the source, but perhaps what you can test is with a constant hessian lens model (there you know the answer of the constant magnification over an extended range)

test/test_LensModel/test_lens_model_extensions.py
```diff
@@ -75,12 +76,63 @@ def test_elliptical_ray_trace(self):
...
+        pc_per_arcsec = 1000 / astropy.arcsec_per_kpc_proper(z_source).value
+        source_sigma = source_fwhm_parsec / pc_per_arcsec / 2.355
+
+        mag_square_grid = extension.magnification_finite(x_image, y_image, kwargs_lens, source_sigma=source_sigma,
+                                                         grid_number=5000, window_size=0.45)
+
+        mag_adaptive_grid = extension.magnification_finite_adaptive(x_image, y_image, source_x, source_y, kwargs_lens, source_fwhm_parsec,
+                                                                    z_source)
+
+        mag_square_grid *= max(mag_square_grid) ** -1
```
sibirrer, line 109: testing against magnification_finite does not require you to re-scale with the magnification. Perhaps you want to test this in relative precision rather than in absolute one (modulo magnification).

test/test_LensModel/test_lens_model_extensions.py
```diff
@@ -75,12 +76,63 @@ def test_elliptical_ray_trace(self):
...
+        grid_r = np.hypot(grid_x, grid_y)
+
+        source_model = LightModel(['GAUSSIAN'])
+        kwargs_source = [{'amp': 1., 'center_x': source_x, 'center_y': source_y, 'sigma': source_sigma}]
+
+        r_min = 0.
+        r_max = source_sigma*0.9
+        flux_array = extension._iterate_adaptive(flux_array, x_image, y_image, grid_x, grid_y, grid_r, r_min, r_max,
+                          lensmodel, kwargs_lens, source_model, kwargs_source)
+        npt.assert_equal(True, flux_array[0] > 0.)
```
sibirrer, line 126: is there a more specific expected outcome of this?

sibirrer (changes requested):
Thanks for making this PR! This is a nice addition and well documented. I added some requests for further documentation and minor changes of the design and tests. Overall, this function matches the type of functionality in the LensModelExtension routines.

lenstronomy/LensModel/lens_model_extensions.py
dangilman: I initially intended to do that but it's a bit more complicated and I didn't feel like coding it up at the time. A circular aperture is also 'safe' in the sense that it should capture highly-curved images that might extended beyond an ellipse estimated from the magnification tensor

test/test_LensModel/test_lens_model_extensions.py
dangilman: I'm not really sure what you mean

lenstronomy/LensModel/lens_model_extensions.py
dangilman: when you say add them to the input requirements, do you mean remove the astropy instance creation from the code and instead explicitly specify a value for parsec_per_arcsec in the function call?

test/test_LensModel/test_lens_model_extensions.py
sibirrer: a test for the absolute magnification/brightness relative to the (analytically known) expected brightness derived from intrinsic brightness x magnification

lenstronomy/LensModel/lens_model_extensions.py
sibirrer: ok, that's fine. But it's useful still to clarify that in the top-level documentation such that if someone else picks it up, there is space for extensions

lenstronomy/LensModel/lens_model_extensions.py
sibirrer: no, I mean that the units of the deflections and angles must be arc seconds, such that the conventions are met. Just add [in arc seconds] or so to the input parameter where this is relevant.

test/test_LensModel/test_lens_model_extensions.py
dangilman: so basically a test w.r.t. lensmodel.magnification?

lenstronomy/LensModel/lens_model_extensions.py
dangilman: ok, if I were to include it now, would it basically amount to rotating the ellipse such that the major axis is parallel to the eigenvector with the largest eigenvalue?

sibirrer: yes (probably), but if you are not using it, it would require you to do also quite some tests to make sure it's working. If you make a comment/TODO/idea that'll do.

test/test_LensModel/test_lens_model_extensions.py
sibirrer: yes, correct!

lenstronomy/LensModel/lens_model_extensions.py
dangilman: I think I got it working, I make some plots of the full ray tracing through the rectangular window the ray tracing through an ellipse with major axis parallel to largest eigenvector 

image

dangilman: image

image

image

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -1,6 +1,9 @@
 import numpy as np
 import lenstronomy.Util.util as util
 from skimage.measure import find_contours
+from lenstronomy.Cosmo.background import Background
```
sibirrer, line 4: do you need that still?

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -1,6 +1,9 @@
...
+from lenstronomy.LightModel.light_model import LightModel
+from astropy.cosmology import default_cosmology
```
sibirrer, line 6: perhaps to make the imports a bit more lightweight on top, can you import this directly just before using the cosmology? It's a class instance so it might be used multiple times (at least in theory)

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -23,6 +26,149 @@ def hessian()
...
+        :param y_image: a list or array of y coordinates [units arcsec]
+        :param kwargs_lens: keyword arguments for the lens model
+        :param source_fwhm_parsec: the size of the background source [units parsec]
+        :param z_source: the source redshift
+        :param cosmo: (optional) an instance of astropy.cosmology; if not specified, a default cosmology will be used
+        :param grid_resolution: the grid resolution in units arcsec/pixel; if not specified, an appropriate value will
+        be estimated from the source size
+        :param grid_radius_arcsec: (optional) the size of the ray tracing region; if not specified, an appropriate value
+        will be estimated from the source size
+        :param axis_ratio: the axis ratio of the ellipse used for ray tracing
```
sibirrer, line 61: do you also want to have the option such that the axis ratio is the tangential/radial eigenvalue? This might further optimize (if local value is reflective to the non-local behavior)

lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -23,6 +26,149 @@ def hessian()
...
+        for i, (xi, yi) in enumerate(zip(x_image, y_image)):
+
+            w1, w2, v11, v12, v21, v22 = self.hessian_eigenvectors(xi, yi, kwargs_lens)
+            if abs(w1) > abs(w2):
+                v = np.array([v11, v12])
+            else:
+                v = np.array([v21, v22])
+            rotation_angle = np.arctan(v[1] / v[0]) - np.pi / 2
+            grid_x, grid_y = util.rotate(grid_x_0, grid_y_0, rotation_angle)
+            grid_r = np.hypot(grid_x, grid_y / axis_ratio).ravel()
```
sibirrer, line 106: the axis ratio might be also used from the eigenvalues w1 and w2 (same comment as above)

test/test_LensModel/test_lens_model_extensions.py
```diff
@@ -6,15 +6,18 @@
 from lenstronomy.LensModel.lens_model_extensions import LensModelExtensions
 from lenstronomy.LensModel.Solver.lens_equation_solver import LensEquationSolver
 from lenstronomy.LensModel.lens_model import LensModel
+from lenstronomy.Cosmo.background import Background
```
sibirrer, line 9: do you need that import still?

test/test_LensModel/test_lens_model_extensions.py
```diff
@@ -75,12 +77,76 @@ def test_elliptical_ray_trace(self):
...
+                                                                    source_fwhm_parsec,
+                                                                    z_source, cosmo=None, tol=0.0001)
+        # tests the r_max > sqrt(2) * grid_radius stop criterion
+        _ = extension.magnification_finite_adaptive(x_image, y_image, source_x, source_y, kwargs_lens,
+                                                    source_fwhm_parsec,
+                                                    z_source, cosmo=None, tol=0.0001, step_size=1000)
+
+        mag_point_source = abs(lensmodel.magnification(x_image, y_image, kwargs_lens))
+
+        npt.assert_almost_equal(mag_square_grid/mag_adaptive_grid, np.ones_like(mag_square_grid), 2)
```
sibirrer, line 120: this test is to the percent level. Does the test fail at the per mill level? Sub-percent precision might be necessary for very precise flux measurements and it's better it fulfills this requirement.

sibirrer (changes requested):
two more minor requests and then I think we are good to go. Thanks a lot for putting your time into this!

dangilman:
Hi,

I think I addressed all your comments in the version I just pushed. 

About the precision, the magnifications are all off by a constant factor, which means that the smaller ray tracing area in the new routine misses a small fraction of the total integrated flux. This can be fixed by making the default grid size larger, but that in turn slows down the algorithm a bit.  

I made a compromise by increasing the default grid size such that the magnifications are accurate to within half a percent. Since the discrepancy is a constant factor, the flux ratios are more precise, accurate to a quarter percent with the default setting. 

- Daniel

sibirrer (approved):
Thank you Daniel! This looks all good to me! One note on the cosmology dependence that came to my mind: The LensModel instance might also be cosmology dependent and in the way this routine is coded, the user might accidentally use two different cosmologies. So in the long run either asking the user to pass the angular source size (meaning putting more responsibility directly in the hands of the user) or having a more generalized 'default cosmology' within the lenstronomy environment might be useful. Just some thoughts.

PR #232, opened 2021-01-14
access cosmology specified in LensModel in magnification_finite_adaptive

dangilman:
I just added that attempts to access the cosmo attribute of _lensModel, and only reverts to the default astropy cosmology if the _lensModel class does not have a .cosmo attribute for whatever reason

### DIFF
--- a/lenstronomy/LensModel/lens_model_extensions.py
+++ b/lenstronomy/LensModel/lens_model_extensions.py
@@ -72,13 +72,8 @@ class LensModelExtensions(object):
         """
 
         if cosmo is None:
-            # try to access a cosmology specified in the lens model first
-            try:
-                cosmo = self._lensModel.cosmo
-            except:
-                from astropy.cosmology import default_cosmology
-                cosmo = default_cosmology.get()
-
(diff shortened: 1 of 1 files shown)


lenstronomy/LensModel/lens_model_extensions.py
```diff
@@ -72,8 +72,12 @@ def magnification_finite_adaptive(self, x_image, y_image, source_x, source_y, kw
         """
 
         if cosmo is None:
-            from astropy.cosmology import default_cosmology
-            cosmo = default_cosmology.get()
+            # try to access a cosmology specified in the lens model first
+            try:
```
sibirrer, line 76: you actually don't need the try and error. If you look at the LensModel class, this instance is always generated

sibirrer (changes requested):
minor change to avoid try and error when not necessary (also avoids further testing/test coverage decrease in this PR)

dangilman:
coverage went down only because I removed some lines that were tested

sibirrer:
Thanks! Test coverage decrease due to fewer lines of code

PR #233, opened 2021-01-17
add auto grid size and resolution computation to util 

dangilman:


lenstronomy/Util/util.py
```diff
@@ -529,6 +529,33 @@ def neighborSelect(a, x, y):
                     values.append(a[i])
     return np.array(x_mins), np.array(y_mins), np.array(values)
 
+def auto_raytracing_grid_size(source_fwhm_parcsec, grid_size_scale=0.01, power=1.):
```
sibirrer, line 532: this definition is for me a bit too specific to be part of the general util functions (there are other functions currently in there that might also need to move somewhere else). As these routines are very specific to the usage with the magnification_finite_adaptive() call, perhaps having these functions in a 'adaptive_ray_tracing_util' file or even close to the extension file in the LensModel module would be better suited.

sibirrer: these functions are not separately being tested without the usage within the adaptive magnification calculation. Can you make test functions specific to this definitions?

sibirrer (changes requested):
Looks overall good. I would like to see the added functions being put in a different file as the util.py has already quite a long list of functions and the ones you added are very specific to the adaptive ray-tracing.

dangilman:
Hi,

Ah oops I wrote individual test functions for the new routines but forgot to include them in the pull request. 

If you want I can move the new functions into LensModelExtensions, although I would like to access them externally so I don’t want to make them private. I could create a whole new util file if you want, but it would only have those two functions in it…. 

- Daniel

sibirrer:
> Hi, Ah oops I wrote individual test functions for the new routines but forgot to include them in the pull request. If you want I can move the new functions into LensModelExtensions, although I would like to access them externally so I don’t want to make them private. I could create a whole new util file if you want, but it would only have those two functions in it…. - Daniel
> […](#)

If you can make a new file, either in Util or in LensModel and put those routines in there, that might be good. Then you can access them from outside and have less danger that I move them around :)

dangilman:
Ok I did it 

- Daniel

dangilman:
Hi,

When I pulled the latest lenstronomy version I got a new error when running it on the University of Toronto cluster. The error shows up only when I submit through the cluster job portal, not when I run locally on the my laptop terminal, or through the terminal in the login portal on the cluster. The error message is 

RuntimeError: cannot cache function 'rotate': no locator available for file '/gpfs/fs1/home/b/bovy/gilmanda/Code/lenstronomy/lenstronomy/Util/util

I don’t know exactly what the error means, but I do know that it has to do with the @jit() decorators in util. As a temporary workaround I removed the @jit() decorators on my fork of lenstronomy, and the code runs fine. Just thought I’d let you know since other people might run into the same issue 

- Daniel

sibirrer:
thanks for letting me know! Perhaps we need a different @jit environment for cluster settings. This might indeed be also an issue for other people. I look into this

PR #239, opened 2021-02-25
Multi plane cosmology interpolation

sibirrer:
This PR allows to initiate and call an interpolated cosmology back end to satrapy.cosmology with the aim for a faster LensModel() class instantiation with hundreds of lensing planes, when the options are chosen.

### DIFF
--- a/lenstronomy/LensModel/multi_plane.py
+++ b/lenstronomy/LensModel/multi_plane.py
@@ -16,7 +16,8 @@ class MultiPlane(object):
     """
 
     def __init__(self, z_source, lens_model_list, lens_redshift_list, cosmo=None, numerical_alpha_class=None,
-                 observed_convention_index=None, ignore_observed_positions=False, z_source_convention=None):
+                 observed_convention_index=None, ignore_observed_positions=False, z_source_convention=None,
+                 cosmo_interp=False, z_interp_stop=None, num_z_interp=100):
         """
 
         :param z_source: source redshift for default computation of reduced lensing quantities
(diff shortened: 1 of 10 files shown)


lenstronomy/LensModel/multi_plane.py
```diff
@@ -36,10 +37,15 @@ def __init__(self, z_source, lens_model_list, lens_redshift_list, cosmo=None, nu
```
swagnercarena, line 37: it may be useful to copy a slightly modified definition of the parameters 

> cosmo_interp=False, z_interp_stop=None, num_z_interp=100

from LensModel here

lenstronomy/LensModel/multi_plane_base.py
```diff
@@ -16,7 +16,7 @@ class MultiPlaneBase(ProfileListBase):
     """
 
     def __init__(self, lens_model_list, lens_redshift_list, z_source_convention, cosmo=None,
-                 numerical_alpha_class=None):
+                 numerical_alpha_class=None, cosmo_interp=True, z_interp_stop=None, num_z_interp=100):
```
swagnercarena, line 19: Why does the base class default to cosmo_interp=True but the MultiPlane class does not?

sibirrer: I couldn't decide what the default should be. I though choosing the default to not change anything with previous usage is better for backwards compatibility.

swagnercarena (commented):
Everything looks good. I added some more modifications to improve speed.

swagnercarena (approved):

sibirrer:
can you add your name and github username on the AUTHORS.rst file in alphabetic order? Then the PR is ready to go :)

sibirrer:
Awesome! Thanks a lot!

Issue #241, opened 2021-03-02
Numba for people who don't want to install it


ewoudwempe:
Edit: ah, nevermind, I think you reverted that commit in the end, I should read the commits a bit better, apologies! Still, a better solution is probably a good idea?
#240 reverted some, in my opinion, quite helpful numba decorators, especially the neighborselect decorator sped up the lens equation solver quite a bit for me (see #227). Can I just add them back, or do we need some mechanism to help people who do not have numba installed? 

I suppose we could just make the jit decorator do nothing if the numba package fails to load, so that all code works if numba is not installed, would that solve your issue @dangilman?

dangilman:
ah, the @jit decorators cause my code to crash when I run on a cluster so I remove them on my branch. I guess I forgot to put them back in when submitting the pull request

ewoudwempe:
Actually, you didn't, I was just a bit confused by the commits :)

sibirrer:
some people are indeed struggling with numba (either the install or the configuration on clusters). Having a better configuration method (currently you would need to over-write on initialization the variables in numba_util.py) and/or a drop-back to pure python in case it fails would be great indeed.

sibirrer:
I close this issue since there is a configuration file that enables to disable numba completelyhttps://github.com/lenstronomy/lenstronomy/blob/main/lenstronomy/Conf/conf_default.yaml

PR #242, opened 2021-03-02
Added a new fast implementation of the EPL profile

ewoudwempe:
This is the same profile as the already existing EPL profile, but my own implementation of it with Numba. Numba code is never pretty, but it is quite a bit faster than the existing PEMD models. Timings:
https://gist.github.com/ewoudwempe/10d47a7433afdb4d750dfad29e8dcd95
Comments? I suppose I'll have to make some tests too...
I could cache the parameter conversions, but it is a very small part of the computation times, so I don't think it's worth it.

PR #243, opened 2021-03-02
Adds a safeguard that falls back to pure python if importing numba fails

ewoudwempe:
or if the environment variable NUMBA_DISABLE_JIT is non-empty.
Should solve #241. Can you confirm this works @dangilman?

### DIFF
--- a/lenstronomy/Util/numba_util.py
+++ b/lenstronomy/Util/numba_util.py
@@ -38,9 +38,11 @@ with open(conf_file) as file:
     fastmath = numba_conf['fastmath']
     error_model = numba_conf['error_model']
 
-#nopython = True
-#cache = True
-#parallel = False
+if numba_enabled:
+    try:
+        import numba
+    except ImportError:
+        numba_enabled = False
 
 __all__ = ['jit']
 
--- a/lenstronomy/Util/numba_util.py
+++ b/lenstronomy/Util/numba_util.py
@@ -2,7 +2,7 @@ import numba
 import lenstronomy
 import os
 import yaml
-
+import numpy as np
 
 """
 From pyautolens:
@@ -12,7 +12,6 @@ If on laptop:
 If on super computer:
 @numba.jit(nopython=True, cache=False, parallel=True)
 """
-
 
 from xdg.BaseDirectory import xdg_config_home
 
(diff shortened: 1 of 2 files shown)


lenstronomy/Util/numba_util.py
```diff
@@ -17,7 +23,11 @@
 
 
 def jit(nopython=nopython, cache=cache, parallel=parallel):
-    def wrapper(func):
-        return numba.jit(func, nopython=nopython, cache=cache, parallel=parallel)
+    if numba_enabled:
```
sibirrer, line 26: Is it possible to change the numba_enabled variable and run a function with @jit decorator twice to test both options? Currently, depending on the environment, the tests are only executed once. Perhaps, to save guard failing of the tests if numba is not enabled, only explicitly test the numba_enabled=False on top of whatever the system option is (which, if installed properly, should be using numba by default)

sibirrer (changes requested):
That's great! One thing that is still missing from what I added in #241 is a better option to configure the numba settings. On a cluster you need different settings and the current way is, well, a bit a hack in that regard (my code).

Let me know if it is not possible to perform a test as mentioned in the review comments. Thanks a lot!

lenstronomy/Util/numba_util.py
ewoudwempe: Sure, you modify the decorator to always run the jitted function as well as the pure function, and then for the testing assert that the two functions return the same value. But then you want to separate between testing and real-world conditions? Because we don't want the everything to run twice if we are not running the test suite. Something like this could be used https://stackoverflow.com/questions/25188119/test-if-code-is-executed-from-within-a-py-test-session

sibirrer: I was more thinking of just executing one single call with a changed environment and then go back to the previous one by manually changing the numba_util.numba_enable variable (deep copy the previous one, setting it =False and then go back to the previous one). This would just ensure that both calls get executed at least once during test time.

ewoudwempe: Decorators immediately execute when defining a function to return a modified function, and although you could create a new wrapper function in the jit decorator, this will not work with Numba, as numba can only deal with jitted functions, and the wrapper is necessarily too general in its type assignments for numba to handle. Actually, my earlier suggestion would therefore also not work... hmm... I'm not sure how to handle that then. To 'switch' between jitted and non-jitted functions, the imported to-be-tested modules need to be reloaded somehow I think.

sibirrer: hmm...I see. So defining configurations needs to be done on a higher level with pre-configured configuration files. Can you move the @jit decorators in PR #242  on the ellipticity2phi_q and the nan_to_zero routines to this PR? Then we have all numba related parts affecting potentially other aspects all together in this PR. Thanks!

ewoudwempe: Done. I put the nan_to_zero in numba_util as it is quite specifically for numba, but if you want I can put it in util.py too. How do you want to solve the testing issue? Maybe add an extra testing module that specifically tests the numba stuff with numba on?

sibirrer: Thanks! For me it's less a testing issue than the ability for the user to change the numba settings pre-run. This might require a separate config file (which needs to be editable by the user and not be part of package itself to not being overwritten when lenstronomy is updated. I added @dangilman to this PR. Perhaps having a config import (in specific path as a yaml file, and if it does not exists, go back to default numba settings?

dangilman: yeah I guess you could make a numba_config.txt file and put it in the Util directory, then add something like this in numba_util.py

import inspect
file_path = inspect.getfile(inspect.currentframe()) + 'numba_config.txt' 
f = open(file_path, 'r')
config = eval(f.read())
if config==0:
    ~do something~
else:
    ~do something else~

dangilman: there is probably a more elegant way....

sibirrer:
I merged the PR with the yaml file configurations. It resulted in a conflict (hopefully minor) due to some duplication. This PR is otherwise ready to go. Thanks a lot for the patience!

ewoudwempe:
Thanks, I updated my PR. Perhaps you can turn off Numba for the coverage run by the way? The new function is implicitly tested in the other commit - good enough?

sibirrer (changes requested):
Everything looks great! I commented in your other PR about docstrings in some of the definitions. These docstrings appear in the html documentation and it would be great if you could add some. Other than that it's great! Thank you very much!

ewoudwempe:
Sorry for the mess, but this one is good now. Apparently you need to not forget to rebase onto master after you do an interactive rebase to move some commits :)

sibirrer:
> Sorry for the mess, but this one is good now. Apparently you need to not forget to rebase onto master after you do an interactive rebase to move some commits :)

I don't see an added documentation to the routines in this PR. I added my comments on the other PR. Sorry for the slight confusion of doing two separate PR's

ewoudwempe:
I don't understand... I did add the documentation to this PR - if you go to files changed tab in github: line 76, 83, 96 contain (admittedly short) docstrings for the added numba nan_to_num functions, and the new @generated_jit decorator, part of commit 9c58f86. Or did you mean a more verbose multi-line docstring with all arguments on separate lines etc?

sibirrer:
I see, I though of doc strings for the individual parameters but the numba reference is sufficient (and probably most accurate). Thanks and sorry for the confusion!

sibirrer (approved):

PR #244, opened 2021-03-02
allow a user to define environmental configurations

sibirrer:
configuration for numba in a yaml file. Current PR is given a default file and a place-holder to allow a different path (we still need to decide where this absolute path could be, suggestions welcome). This PR is intended to allow more flexibility and to complement #243


sibirrer:
Hi @dangilman here a PR for a configuration of numba from a yaml file. The idea is that you can specify and save a user specified configuration for the lenstronomy numba options and if this file exists, it takes those options and overwrites the default configurations. I currently have not specified the user location (could be absolute path to user home and going from there). This might actually work for you if you on the cluster with the right settings. You should have access to the PR branch directly and can edit it too. My idea is to merge this PR first, then go to #243 and then to #242. In principle this PR can be merged prior to specifying the specific path for the user to allow for smoother merge. Let me know what you think!

ewoudwempe:
The 'standard' path for config files on linux systems is $XDG_CONFIG_HOME or ~/.config, so we could use something like $XDG_CONFIG_HOME/lenstronomy/config.yml. For example matplotlib also does this. If that file doesn't exist we could also check the Conf dir itself (needs a .gitignore entry), that is perhaps easier for windows/mac users.

Anyway, isn't an 'enabled' option missing that disables Numba entirely - in case the system has an old version of numba that does not trigger an importerror but which needs to be disabled anyway? And we could maybe make the config file less specific to Numba for future additions, and instead put Numba stuff in a Numba block.

sibirrer:
> The 'standard' path for config files on linux systems is $XDG_CONFIG_HOME or ~/.config, so we could use something like $XDG_CONFIG_HOME/lenstronomy/config.yml. For example matplotlib also does this. If that file doesn't exist we could also check the Conf dir itself (needs a .gitignore entry), that is perhaps easier for windows/mac users.
> 
> Anyway, isn't an 'enabled' option missing that disables Numba entirely - in case the system has an old version of numba that does not trigger an importerror but which needs to be disabled anyway? And we could maybe make the config file less specific to Numba for future additions, and instead put Numba stuff in a Numba block.

Thanks for the hint with xdg! I implemented a version of this. I added the keywords you introduced in your PR #243 but did not wanted to interfere to much with your PR for merging conflicts @ewoudwempe .

@dangilman  I think this PR allows you now to customize and/or disable numba and we can merge it with the main branch to allow @ewoudwempe 's PR's to proceed. Let me know otherwise.

dangilman:
Go for it

PR #248, opened 2021-03-10
Vectorize LightModel.Interpol's interpolation

JelleAalbers:
This uses the grid=False option to scipy [`scipy.interpolate.RectBivariateSpline.__call__`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RectBivariateSpline.__call__.html) to simplify and speed up the `Interpol` light model. 

In simple examples (where lensing or other calculations are not the limiting factor), this can give a significant speedup to image generation. I got a factor ~5 improvement in a simple case with only an SIS and a tNFW lens. The interpolation itself speeds up by about 20x for a 125x125 pixel image.

(I guess the coverage decrease is because this shortens already-tested code?)

### DIFF
--- a/lenstronomy/LightModel/Profiles/interpolation.py
+++ b/lenstronomy/LightModel/Profiles/interpolation.py
@@ -40,16 +40,8 @@ class Interpol(object):
         :param scale: pixel scale (in angular units) of the simulated image
         :return: surface brightness from the model at coordinates (x, y)
         """
-        n = len(np.atleast_1d(x))
         x_, y_ = self.coord2image_pixel(x, y, center_x, center_y, phi_G, scale)
-        if n <= 1 and np.shape(x) == ():
-            f_out = self.image_interp(x_, y_, image)
-            return f_out[0][0] * amp
-        else:
-            f_out = np.zeros(n)
-            for i in range(n):
-                f_out[i] = self.image_interp(x_[i], y_[i], image)
-        return f_out * amp
+        return amp * self.image_interp(x_, y_, image)
 
     def image_interp(self, x, y, image):
         if not hasattr(self, '_image_interp'):
@@ -66,7 +58,7 @@ class Interpol(object):
 
         # y and x must be flipped in call to interpolator
         # (try reversing, the unit tests will fail)
-        return self._image_interp(y, x)
+        return self._image_interp(y, x, grid=False)
 
     def total_flux(self, image, scale, amp=1, center_x=0, center_y=0, phi_G=0):
         """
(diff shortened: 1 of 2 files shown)


lenstronomy/LightModel/Profiles/interpolation.py
```diff
@@ -40,16 +40,8 @@ def function(self, x, y, image=None, amp=1, center_x=0, center_y=0, phi_G=0, sca
...
         x_, y_ = self.coord2image_pixel(x, y, center_x, center_y, phi_G, scale)
-        if n <= 1 and np.shape(x) == ():
-            f_out = self.image_interp(x_, y_, image)
-            return f_out[0][0] * amp
-        else:
-            f_out = np.zeros(n)
-            for i in range(n):
-                f_out[i] = self.image_interp(x_[i], y_[i], image)
-        return f_out * amp
+        return amp * self.image_interp(x_, y_, image)
```
sibirrer, line 44: can you make one test function to show that if the input x, y are single values (not arrays) then the output is also a single value (and not an array of length one)? I think this is the reason I distinguished the two modes between a loop and a single call with output f_out[0][0]

sibirrer (changes requested):

sibirrer:
Thanks @JelleAalbers ! I have one request to check the output format. I am not sure there was a test explicitly testing it in the previous version but I know it was relevant for splitting the output format after the interpolation call.

lenstronomy/LightModel/Profiles/interpolation.py
JelleAalbers: That's covered by https://github.com/sibirrer/lenstronomy/blob/main/test/test_LightModel/test_light_model.py#L56, which also tests Interpol, right? Or would you like an additional test of this in the dedicated interpol testing code?

JelleAalbers: That's covered by https://github.com/sibirrer/lenstronomy/blob/main/test/test_LightModel/test_light_model.py#L56, which also tests Interpol, right? Or would you like an additional test of this in the Interpol testing code?

sibirrer: this function does not test the return of the interpol format (the interpol one is not part of this function call). A separate test within the interpol testing code would be appreciated. It can just be one line in which you evaluate x=0, y=0 and test that the return statement is a single value and not a list. Thanks

JelleAalbers: Sure! There are some single-value tests in test_interpolation already, so I added an assert that these indeed give a float.

sibirrer (approved):

sibirrer:
thanks a lot! Test coverage drop only due to reduction of lines of code

PR #251, opened 2021-04-03
Fix identity checks that trigger SyntaxWarning on python 3.8

JelleAalbers:
This extremely minor PR fixes  two identity comparisons that [trigger a `SyntaxWarning`](https://bugs.python.org/issue34850) at import time, at least on some python versions since 3.8 (in some circumstances).
```
/sdf/home/j/jaalbers/software/lenstronomy/lenstronomy/Sampling/Likelihoods/flux_ratio_likelihood.py:44: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if self._source_type is 'INF':
/sdf/home/j/jaalbers/software/lenstronomy/lenstronomy/Workflow/fitting_sequence.py:229: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if sampler_type is 'EMCEE':
```

### DIFF
--- a/lenstronomy/Sampling/Likelihoods/flux_ratio_likelihood.py
+++ b/lenstronomy/Sampling/Likelihoods/flux_ratio_likelihood.py
@@ -41,7 +41,7 @@ class FluxRatioLikelihood(object):
         :param kwargs_cosmo:
         :return: log likelihood of the measured flux ratios given a model
         """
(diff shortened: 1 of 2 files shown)


sibirrer:
Thanks @JelleAalbers ! This was also noted by others in other parts of the code and I hope the 'is' is replaced by '==' subsequently.

PR #252, opened 2021-04-29
JOSS paper: expand names where necessary

ntessore:
This standardises all names which end up showing up in the draft.


sibirrer (approved):
Thank you! Let's see whether it compiles properly. Otherwise we leave it like that. I was not considering that non-uniform descriptions between different citation keys might be the issue, but perhaps the compilation code is smarter than me :)

PR #255, opened 2021-05-26
Uldm profile (and correlated) update

lucateo:
Hi, 

as promised, I am making a new pull request with some updated features for the Uldm profile. In particular, I allow for the possibility of inserting a slope different from the one you would obtain by the so called "self gravitating soliton" (useful if you want to implement feedback arising from background potential). This is a feature I used in my recent work (2105.10873 for those who are interested) which made use of lenstronomy. 

I also modified the cored_density_mst.py file by inserting in all functions a rescaling with lambda_approx; the thing is, these profiles are used just in the MCMC inference to try to minimize covariances, but the MCMC inference done when you just subtract a perfect mass sheet from the cored profile doesn't give you back the "original" central convergence; if I worked out the math correctly (and I think I did, since with this correction I obtain what I expect), rescaling the functions with lambda_approx should overcome this issue.

Finally, I insert, always in cored_density_mst.py, a fourth profile which refers to Uldm profile. Given the structure of the function defined in that file, I just implement the possibility to use Uldm class in this "subtracted mass sheet degeneracy" way only for the standard soliton with slope = 8.


lenstronomy/LensModel/Profiles/uldm.py
```diff
@@ -15,51 +17,45 @@ class Uldm(LensProfileBase):
 
     .. math::
 
-        \rho = \rho_0 (1 + 0.091(r/r_c)^2)^{-8}
+        \rho = \rho_0 (1 + a(r/r_c)^2)^{-\beta}
```
sibirrer, line 20: can you change the '\' into a '\\' in the math mode? Apparently then it gets recognized as a latex symbol.

sibirrer, line 20: Can you also add the formula in terms of the parameters being used in this profile?

lenstronomy/LensModel/Profiles/uldm.py
```diff
@@ -68,31 +64,39 @@ def function(self, x, y, kappa_0, theta_c, center_x=0, center_y=0):
...
+        if np.isscalar(r) == True:
+            hypgeom = float(kappa_0 /2 * r**2 * 
+                hyp3f2(1, 1, slope - 0.5, 2, 2, -(a_factor_sqrt * r /theta_c )**2))
+        else:
+            hypgeom =  np.array([ kappa_0 /2. * r_i**2. *
+                hyp3f2(1, 1, slope - 0.5, 2, 2, -(a_factor_sqrt * r_i / theta_c)**2.) for r_i in r], dtype=float)
+        return hypgeom
 
-    def alpha_radial(self, r, kappa_0, theta_c):
+    def alpha_radial(self, r, kappa_0, theta_c, slope = 8):
```
sibirrer, line 76: in the input argument, following PEP8, can you remove the spacing in slope=8?

lenstronomy/LensModel/Profiles/cored_density_mst.py
```diff
@@ -48,7 +55,7 @@ def function(self, x, y, lambda_approx, r_core, center_x=0, center_y=0):
         kappa_ext = 1 - lambda_approx
         f_cored_density = self._profile.function(x, y, kappa_ext, r_core, center_x, center_y)
         f_ms = self._convergence.function(x, y, kappa_ext, center_x, center_y)
-        return f_cored_density - f_ms
+        return (f_cored_density - f_ms)/lambda_approx
```
sibirrer, line 58: would it be, perhaps, easier if you re-define kappa_ext instead of dividing the return arguments? This might save some code, i.e. kappa_ext = (1 - lambda_approx) / lambda_approx

sibirrer (changes requested):
This looks great! Thank you! I have some very minor documentation and formatting comments. For some reason, the coveralls coverage report was not triggered. I hope this will be done in the next push, but I don't see any untested lines myself.

lenstronomy/LensModel/Profiles/uldm.py
lucateo: I am not so sure I understand, I tried to see other examples in lenstronomy and the solution I found is to change \ into \\, let me know if I should do something else.

sibirrer (approved):

sibirrer:
Thank you @lucateo ! This is grea! If there are some formatting issues coming up, I can easily change them later too.

PR #264, opened 2021-06-15
SIDM jeans profile

jhod0:
Adds a model for the profile of Self-Interacting Dark Matter halos using the Jeans method modeling, following [Robertson 2021](https://github.com/sibirrer/lenstronomy.git) and [Sagunski 2021](https://github.com/sibirrer/lenstronomy.git).

The model is called `JEANS_ISO`. I have used it on my own machine and in low cross section limit it looks identical to NFW as expected, but I have not fully tested it.

This is a work in progress. TODOs:

- [ ] Fix inconsistent naming
- [ ] Examples
- [ ] Hessian of profile (currently only lensing potential and deflection are implemented)
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lenstronomy/Cosmo/lens_cosmo.py
+++ b/lenstronomy/Cosmo/lens_cosmo.py
@@ -2,10 +2,12 @@ __author__ = 'sibirrer'
 
 # this file contains a class to convert lensing and physical units
 
+from astropy import units as u
 import numpy as np
 import lenstronomy.Util.constants as const
 from lenstronomy.Cosmo.background import Background
 from lenstronomy.Cosmo.nfw_param import NFWParam
+from lenstronomy.LensModel.Profiles.jeans_iso import _jeans_soln
 
 __all__ = ['LensCosmo']
 
@@ -224,22 +226,26 @@ class LensCosmo(object):
 
     def jeans_physical2angle(self, M, c, sigma_v, halo_age):
         """
-        converts the physical mass, concentration, and velocity-weighted cross
-        section of an isothermal SIDM halo halo into lensing parameters.
+        Converts the physical to lensing parameters for the JEANS_ISO
+        self-interacting dark matter model.
+
 
         :param M: mass enclosed 200 rho_crit in units of M_sun (physical units, meaning no little h)
         :param c: NFW concentration parameter (r200/r_s)
         :param sigma_v: velocity-weighted cross section, <\sigma v>, in (cm^2 / g)*(km/s)
         :param halo_age: age of the halo, in Gyr
 
-        :return: Rs_angle (angle at scale radius) (in units of arcsec), alpha_Rs (observed bending angle at the scale radius
+        :return: 5-tuple of Jeans parameters as follows
+                 Rs_angle (angle of scale radius) (in units of arcsec),
+                 alpha (transision radius relative to Rs) (unitless),
+                 beta (Jeans scale radius relative to Rs) (unitless),
+                 gamma (Jeans central density relative to NFW rho0) (unitless),
+                 rho0 (NFW rho0 in angular units)
         """
         rho0, Rs, r200 = self.nfwParam_physical(M, c)
         Rs_angle = Rs / self.dd / const.arcsec
         reduced_rho0 = (rho0 / self.sigma_crit) * self.dd * const.arcsec
 
-        from astropy import units as u
-        from lenstronomy.LensModel.Profiles.jeans_iso import _jeans_soln
         # Need a conversion factor
         factor = (
             1 * u.Unit('g / solMass') * u.Unit('Mpc / cm')**2 * u.Unit('Mpc / km')
--- a/lenstronomy/Cosmo/lens_cosmo.py
+++ b/lenstronomy/Cosmo/lens_cosmo.py
@@ -222,6 +222,39 @@ class LensCosmo(object):
         theta_r200 = r200 / self.dd / const.arcsec
         return theta_r200
 
+    def jeans_physical2angle(self, M, c, sigma_v, halo_age):
+        """
+        converts the physical mass, concentration, and velocity-weighted cross
+        section of an isothermal SIDM halo halo into lensing parameters.
+
+        :param M: mass enclosed 200 rho_crit in units of M_sun (physical units, meaning no little h)
+        :param c: NFW concentration parameter (r200/r_s)
+        :param sigma_v: velocity-weighted cross section, <\sigma v>, in (cm^2 / g)*(km/s)
+        :param halo_age: age of the halo, in Gyr
+
+        :return: Rs_angle (angle at scale radius) (in units of arcsec), alpha_Rs (observed bending angle at the scale radius
+        """
+        rho0, Rs, r200 = self.nfwParam_physical(M, c)
+        Rs_angle = Rs / self.dd / const.arcsec
+        reduced_rho0 = (rho0 / self.sigma_crit) * self.dd * const.arcsec
+
+        from astropy import units as u
+        from lenstronomy.LensModel.Profiles.jeans_iso import _jeans_soln
+        # Need a conversion factor
+        factor = (
+            1 * u.Unit('g / solMass') * u.Unit('Mpc / cm')**2 * u.Unit('Mpc / km')
+            * u.Unit('s / Gyr')
+        ).to(1).value
+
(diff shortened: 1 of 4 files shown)


lenstronomy/Cosmo/lens_cosmo.py
```diff
@@ -222,6 +222,39 @@ def nfw_M_theta_vir(self, M):
         theta_r200 = r200 / self.dd / const.arcsec
         return theta_r200
 
+    def jeans_physical2angle(self, M, c, sigma_v, halo_age):
+        """
+        converts the physical mass, concentration, and velocity-weighted cross
```
sibirrer, line 227: can you also add to which model this corresponds/ can be used for?

lenstronomy/Cosmo/lens_cosmo.py
```diff
@@ -222,6 +222,39 @@ def nfw_M_theta_vir(self, M):
...
+        :param sigma_v: velocity-weighted cross section, <\sigma v>, in (cm^2 / g)*(km/s)
+        :param halo_age: age of the halo, in Gyr
+
+        :return: Rs_angle (angle at scale radius) (in units of arcsec), alpha_Rs (observed bending angle at the scale radius
+        """
+        rho0, Rs, r200 = self.nfwParam_physical(M, c)
+        Rs_angle = Rs / self.dd / const.arcsec
+        reduced_rho0 = (rho0 / self.sigma_crit) * self.dd * const.arcsec
+
+        from astropy import units as u
```
sibirrer, line 241: perhaps some of these imports can be made at the file level to avoid a lot of (new) imports for every function call (if it becomes an issue)

lenstronomy/LensModel/Profiles/jeans_iso.py
```diff
@@ -0,0 +1,515 @@
...
+    return -np.sqrt(2/3)*np.tanh(x/np.sqrt(6))
+
+
+def _ddy_approx(x):
+    return -1/(3*np.cosh(x/np.sqrt(6)))
+
+
+class _JeansSolutionInterp:
+    '''
+    Solves the Jeans method differential equation, and boundary conditions
```
sibirrer, line 36: perhaps some more documentation with context and references can be useful here

lenstronomy/LensModel/Profiles/jeans_iso.py
```diff
@@ -0,0 +1,515 @@
...
+        )
+        deflection = deflection.reshape(pot_shape)
+
+        f_x = (_x/thetas) * deflection
+        f_y = (_y/thetas) * deflection
+
+        # compute f_x, f_y and return
+        return f_x, f_y
+
+    def hessian(self, *args, **kwargs):
```
sibirrer, line 447: if you don't find an analytic solution here, you can also do a numerical differentiation, as done for some other profiles as well

lenstronomy/LensModel/Profiles/jeans_iso.py
```diff
@@ -0,0 +1,515 @@
...
+        """
+        x = radius / Rs
+        if x < alpha:
+            unitless_density = _jeans_soln.enclosed_mass_3d(x / beta)
+        else:
+            raise NotImplementedError
+
+        return rho0 * Rs**3 * unitless_density
+
+    def set_static(self, **kwargs):
```
sibirrer, line 500: you don't need to add these as they are inherited from the base class (unless you want to allow some cache being saved that helps improve function calls for fixed parameters (i.e, could be useful when solving the lens equation)

lenstronomy/LensModel/profile_list_base.py
```diff
@@ -285,6 +285,9 @@ def _import_class(lens_type, custom_class, z_lens=None, z_source=None):
         elif lens_type == 'CORED_DENSITY_ULDM_MST':
             from lenstronomy.LensModel.Profiles.cored_density_mst import CoredDensityMST
             return CoredDensityMST(profile_type='CORED_DENSITY_ULDM')
+        elif lens_type == 'JEANS_ISO':
```
sibirrer, line 288: the name is ok, but it does not indicate that in the outer parts, this is an NFW profile. Perhaps finding a (slightly) different name?

sibirrer (changes requested):
Hi @jhod0 , thanks for this PR!
I think this is a valuable addition! I am waiting for updated documentation and tests for further review. I also ping @dangilman here in case he sees duplication or usefulness for his application as well.

lenstronomy/LensModel/Profiles/jeans_iso.py
jhod0: I had copied these over to remind myself if I wanted to implement them later. There is a lot of numerical integration in the potential/deflection implementations, so I am sure there is room for that kind of optimization!

jhod0:
Thanks for the helpful comments Simon!

dangilman:
Hi @jhod0 and @sibirrer, it looks this pull request is related to adding a lens profile whose density is given by solving the radial Jeans equation out to r1 with the Kaplinghat et al. 2016 prescription. We just wrote a paper doing similar stuff so maybe I have provide some helpful input https://arxiv.org/pdf/2105.05259.pdf. My initial reaction is that you can do this in lenstronomy already if you write the Jeans modeling code and numerically integrate the deflection angles separately. You can store the relevant info in a class and pass it into lenstronomy using the profile NumericalAlpha defined in the numerical_deflections.py file. My other immediate reaction is that solving the Jeans equation and numerically integrating deflection angles each time might be prohibitively slow, but I guess this depends on how many deflection angles you need to compute

jhod0:
Hi @dangilman, Yes that is exactly what this lens profile is for! This was partly inspired by your paper.

What you describe is pretty close to what this does: the `_JeansSolutionInterp` class solves the Jeans equation, stores interpolation tables, and solves the boundary conditions between the NFW and isothermal Jeans parts of the profile. The `SIDMJeans` class does the right integrations and analytical forms when it can to compute the potential and deflection angles. For generating mock images it is not prohibitively slow, I am able to generate the images below in seconds. (They show clusters with the r1 transition radius shown in orange, substructures added at green crosses). More complex things might be prohibitively slow

image

dangilman:
Hi @jhod0, cool sounds good! 

If your project requires deflection angles computed from the exact SIDM density profile rather than an approximation like cored NFW, that would be new and I suppose you need to the numerical integration in addition to the Jeans modeling. Whether or not all of that additional code belongs in lenstronomy versus a separate package I think is up to Simon and you. 

In our last paper about SIDM I tried to package the SIDM stuff separately from lenstronomy, you can check out the repo here: https://github.com/dangilman/SIDMpy. To avoid duplicating too much code, you could use it and add whatever routines you need, if you want.

sibirrer:
> Hi @jhod0, cool sounds good!
> 
> If your project requires deflection angles computed from the exact SIDM density profile rather than an approximation like cored NFW, that would be new and I suppose you need to the numerical integration in addition to the Jeans modeling. Whether or not all of that additional code belongs in lenstronomy versus a separate package I think is up to Simon and you.
> 
> In our last paper about SIDM I tried to package the SIDM stuff separately from lenstronomy, you can check out the repo here: https://github.com/dangilman/SIDMpy. To avoid duplicating too much code, you could use it and add whatever routines you need, if you want.

I am happy to have it included if it is tested and documented and useful for you @jhod0 . Best is to link also to Daniel's repo (and perhaps also the reference) to leave the user a range of options.

jhod0:
Thanks for pointing me to SIDMpy - it looks a lot more general than what I have. I will continue the project I've been doing and judge how useful this would be.

dangilman:
hi @jhod0 we left off here a bit over a year ago. I'm wondering if there are any updates on your end and if you're still planning to add something to lenstronomy through this pull request?

jhod0:
Hi @dangilman , apologies for the very late response. I did give up on this approach, as it was overly complicated and the isothermal solving was prohibitively slow. We can close this PR.

dangilman:
> Hi @dangilman , apologies for the very late response. I did give up on this approach, as it was overly complicated and the isothermal solving was prohibitively slow. We can close this PR.

Hi @jhod0, no problem, and OK, I will close this. For reference, if you're still interested in exploring this topic, I wrote some code that does the SIDM Jeans analysis (although I can't guarantee that it is fast enough for your project needs) https://github.com/dangilman/SIDMpy

PR #269, opened 2021-07-20
Composite profile

dangilman:
general class for a composite lens profile. The two profiles are specified in the profile name by passing in 'COMPOSITE_PROFILE1__PROFILE2', i.e. the profile names are separated by a double underscore. the profiles are created and passed into the composite class, which evaluates profile1 at r<=r_match and profile2 at r>=r_match. I guess this only works for spherical mass distributions. If you think this looks generally OK I can put more time into adding documentation and test functions 

I guess a more appropriate name would be 'CONCATENATE_PROFILE1__PROFILE2'...


sibirrer (changes requested):
Hi Daniel, thanks for making this PR. There are some aspects of this PR that are not related to the added feature you described in the PR documentation and it's probably better to have them in a separate PR.

For the composite profiles, one main issue I see is with the parameter handling for fitting etc, as the format is significantly different. Would a lens model class with a generic hole or cut in the lensing quantities also work and then the superposition of two of such profiles with matched centers and cut radius also work? Obviously if the parameters are not matched properly it results in non-physical lensing properties but in this way the overall structure of the module is preserved

dangilman:
> Hi Daniel, thanks for making this PR. There are some aspects of this PR that are not related to the added feature you described in the PR documentation and it's probably better to have them in a separate PR.
Sorry, looks like was working on a local branch with other changes that then got lumped together with the pull request for the composite profile. I'll get rid of those 

> For the composite profiles, one main issue I see is with the parameter handling for fitting etc, as the format is significantly different. Would a lens model class with a generic hole or cut in the lensing quantities also work and then the superposition of two of such profiles with matched centers and cut radius also work? Obviously if the parameters are not matched properly it results in non-physical lensing properties but in this way the overall structure of the module is preserved

I guess creating two new profiles with lensing properties set to zero in certain regions would be a possibility, but that would be less general, and it would require the creation of several profiles that only have physical significance for a certain combination of profiles. 

The composite profile class as it is currently written would not be compatible with certain parts of lenstronomy, but it would be useful for people who throw a bunch of profiles (i.e. subhalos) with fixed lensing properties into a lens model and want to compute deflection angles

sibirrer:
> I guess creating two new profiles with lensing properties set to zero in certain regions would be a possibility, but that would be less general.

Overall, this would allow the same ability for the user, but with different input formats. So I don't think it's less general.

> 
> The composite profile class as it is currently written would not be compatible with certain parts of lenstronomy, but it would be useful for people who throw a bunch of profiles (i.e. subhalos) with fixed lensing properties into a lens model and want to compute deflection angles. It is left up to the user to make sure the keyword arguments for the concatenated profiles are properly normalized

As the user itself must ensure that the keyword arguments are matching a continuous deflection angle at the boundary, I don't really see a benefit of having a general class compared to a superposition of two models. It might make it for the user even more explicit that one has to worry about continuity.

> 
> If you want, I can abandon the general class and just create an instance for the specific combination of profiles I have in mind

I think for the time being, you can implement your specific choice (even perhaps with a continuous normalization. Sorry for pushing back here as I try to preserve the full applicability of models.

dangilman:
> > I guess creating two new profiles with lensing properties set to zero in certain regions would be a possibility, but that would be less general.
> 
> Overall, this would allow the same ability for the user, but with different input formats. So I don't think it's less general.
> 
> > The composite profile class as it is currently written would not be compatible with certain parts of lenstronomy, but it would be useful for people who throw a bunch of profiles (i.e. subhalos) with fixed lensing properties into a lens model and want to compute deflection angles. It is left up to the user to make sure the keyword arguments for the concatenated profiles are properly normalized
> 
> As the user itself must ensure that the keyword arguments are matching a continuous deflection angle at the boundary, I don't really see a benefit of having a general class compared to a superposition of two models. It might make it for the user even more explicit that one has to worry about continuity.
> 
> > If you want, I can abandon the general class and just create an instance for the specific combination of profiles I have in mind
> 
> I think for the time being, you can implement your specific choice (even perhaps with a continuous normalization. Sorry for pushing back here as I try to preserve the full applicability of models.

Yeah I guess you're right, I'll just switch to a specific combination of models that I want to concatenate

PR #272, opened 2021-07-28
Sampling the log10 of the parameters 

martin-millon:
Hi @sibirrer,

I've added the option on my fork to sample the log of some parameters. It looks efficient when you have poorly constrained parameters and want to explore several orders of magnitude.  

If you think that could be useful to other users and if you are happy with this solution for the implementation, I can extend this to the other parameters (it works only for the lens parameters for the moment) and implement the test. 

There are also other commits to add some options here and there. I can remove this from this PR if necessary. 

Martin.


lenstronomy/Conf/conf_default.yaml
```diff
@@ -12,5 +12,5 @@ numba:
 
 conventions:
 
-  sersic_major_axis: False  # if True, defines the half-light radius of the Sersic light profile along the semi-major axis (which is the Galfit convention)
+  sersic_major_axis: True  # if True, defines the half-light radius of the Sersic light profile along the semi-major axis (which is the Galfit convention)
```
sibirrer, line 15: set =False for PR

lenstronomy/LensModel/lens_param.py
```diff
@@ -35,12 +36,16 @@ def __init__(self, lens_model_list, kwargs_fixed, kwargs_lower=None, kwargs_uppe
 
         self.lower_limit = kwargs_lower
         self.upper_limit = kwargs_upper
+        if kwargs_logsampling is None :
+            kwargs_logsampling = [[] for i in range(len(self.model_list))]
+        self.kwargs_logsampling = kwargs_logsampling
```
sibirrer, line 41: does this need to be a public variable?

lenstronomy/LensModel/lens_param.py
```diff
@@ -77,6 +82,10 @@ def getParams(self, args, i):
                         i += 1
                 else:
                     kwargs[name] = kwargs_fixed[name]
+
+                if name in kwargs_logsampling and not name in kwargs_fixed:
```
sibirrer, line 86: just a question: if you first let a parameter fixed and then let it vary in log space, might this cause problems?

lenstronomy/LensModel/lens_param.py
```diff
@@ -116,7 +126,10 @@ def setParams(self, kwargs_list):
                     #    else:
                     #        pass
                     else:
-                        args.append(kwargs[name])
+                        if name in kwargs_logsampling:
+                            args.append(np.log10(kwargs[name]))
+                        else :
```
sibirrer, line 131: remove space between else and :

lenstronomy/Plots/model_band_plot.py
```diff
@@ -175,7 +175,7 @@ def convergence_plot(self, ax, text='Convergence', v_min=None, v_max=None,
 
     def normalized_residual_plot(self, ax, v_min=-6, v_max=6, font_size=15, text="Normalized Residuals",
                                  colorbar_label=r'(f${}_{\rm model}$ - f${}_{\rm data}$)/$\sigma$',
-                                 no_arrow=False, **kwargs):
+                                 no_arrow=False, color_bar = True, **kwargs):
```
sibirrer, line 178: add description in doc string, remove space around '='

lenstronomy/Sampling/parameters.py
```diff
@@ -127,6 +132,7 @@ def __init__(self, kwargs_model,
         :param source_grid_offset: optional, if True when using a pixel-based modelling (e.g. with STARLETS-like profiles),
         adds two additional sampled parameters describing RA/Dec offsets between data coordinate grid and pixelated source plane coordinate grid.
         :param num_shapelet_lens: number of shapelet coefficients in the 'SHAPELETS_CART' or 'SHAPELETS_POLAR' mass profile.
+        :param log_sampling_lens:
```
sibirrer, line 135: doc string

lenstronomy/Sampling/parameters.py
```diff
@@ -169,6 +175,11 @@ def __init__(self, kwargs_model,
         self._joint_lens_with_light = joint_lens_with_light
         self._joint_lens_with_source_light = joint_lens_with_source_light
         self._joint_source_with_point_source = copy.deepcopy(joint_source_with_point_source)
+
+        self._log_sampling_lens = log_sampling_lens
+        kwargs_logsampling_lens = [[] for i in range(len(self._lens_model_list))]
```
sibirrer, line 180: is this needed here or could you add it within _update_log_sampling? Just to have the __init__ a bit cleaner

lenstronomy/Sampling/parameters.py
```diff
@@ -459,6 +472,22 @@ def _update_joint_param(kwargs_list_1, kwargs_list_2, joint_setting_list):
...
+        :param kwargs_logsampling_lens: list of list of parameters to sample in log10
+        :param log_sampling_lens: [[i_1, ['param_name1', 'param_name2', ...]], [...], ...]
+        :return: updated kwargs_logsampling_lens
+        """
+        for setting in log_sampling_lens:
+            i_1, param_list = setting
+            if type(param_list) == list:
+                kwargs_logsampling_lens[i_1] = param_list
+            else:
+                raise TypeError("Bad format for constraint setting: got %s" % param_list)
```
sibirrer, line 488: Can you improve a bit the error message about how the format should look like?

lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -428,13 +430,15 @@ def update_settings(self, kwargs_model={}, kwargs_constraints={}, kwargs_likelih
...
         :param cosmo_remove_fixed: ['param1', 'param2',...]
+        :param change_lens_lower_limit: [[i_model, ['param_name', ...], [value1, value2, ...]]]
         :return: 0, the settings are overwritten for the next fitting step to come
         """
         self._updateManager.update_options(kwargs_model, kwargs_constraints, kwargs_likelihood)
         self._updateManager.update_fixed(lens_add_fixed, source_add_fixed, lens_light_add_fixed,
                                          ps_add_fixed, cosmo_add_fixed, lens_remove_fixed, source_remove_fixed,
                                          lens_light_remove_fixed, ps_remove_fixed, cosmo_remove_fixed)
-        self._updateManager.update_limits(change_source_lower_limit, change_source_upper_limit)
+        self._updateManager.update_limits(change_source_lower_limit, change_source_upper_limit,change_lens_lower_limit,
```
sibirrer, line 440: add space after coma

sibirrer (changes requested):
Looks good overall and no major changes need to be done to be merged. You can leave the support currently for the lens model and we implement it as the feature gets requested/needed.

Can you add some test(s) making use of the feature as well? Thank you!

sibirrer:
The tests fail due to your different setting for the R_sersic convention. Perhaps you can extend the tests such that they ask for the option being used and then assert two different values. Then the tests should pass regardless of the convention used locally.

lenstronomy/LensModel/lens_param.py
martin-millon: I think, in this case, you will need to use both the 'remove_fixed' option and the 'log_sampling_lens' option to let this parameter vary in logspace. It makes sense to me that this should be two separate operations. I will make some tests to check if everything is OK.

lenstronomy/LensModel/lens_param.py
martin-millon: This is called outside the class by the method Param.print_setting(). kwargs_fixed is public so I think kwargs_logsampling should be public as well, but I can change

lenstronomy/Conf/conf_default.yaml
martin-millon: OK, will change for the PR

lenstronomy/LensModel/lens_param.py
martin-millon: corrected

lenstronomy/Plots/model_band_plot.py
martin-millon: corrected

lenstronomy/Sampling/parameters.py
martin-millon: corrected

lenstronomy/Sampling/parameters.py
martin-millon: I coded it that way to avoid having the method _update_log_sampling reset the kwargs_logsampling, in a similar way than _update_joint_param() or _fix_joint_param(). Can be changed if you think that's necessary.

lenstronomy/Sampling/parameters.py
martin-millon: Corrected

lenstronomy/Workflow/fitting_sequence.py
martin-millon: corrected

lenstronomy/LensModel/lens_param.py
martin-millon: Ah yes, you are right, it will depend on the order you are making the changes... I'll correct this.

martin-millon: OK, I think this is fine since the parameters in the arg only if not in kwargs_fixed. The log10 is applied only in the "args" and we keep always the 'kwargs' in direct space (also valid for the limit, etc.... ).

I have tried this sequence, and I did see any problems : 

`    ['update_settings',
     {
         'lens_add_fixed': [[0, ['theta_E']],
                               ]
     }
     ],

    ['update_settings',
     {
         'kwargs_constraints': {'log_sampling_lens': [
                 [0, ['theta_E']], 
             ]

         },
     }
     ],

    ['PSO',
     {
         'n_particles': n_particles,
         'n_iterations': n_iterations,
         'sigma_scale': 0.1,
     }
     ],

    ['update_settings',
     {
         'lens_remove_fixed': [[0, ['theta_E']],
                            ]
     }
    ],

    ['PSO',
     {
         'n_particles': n_particles,
         'n_iterations': n_iterations,
         'sigma_scale': 1.0,
     }
     ],`

sibirrer: ah, yes log10 are only in the args and not kwargs, so this should be fine. Thanks for checking!

lenstronomy/Sampling/parameters.py
sibirrer: perhaps a short comment in the code for what is done here and why might help for future development and orientation. You can keep it there, thanks

lenstronomy/Sampling/parameters.py
```diff
@@ -459,6 +472,23 @@ def _update_joint_param(kwargs_list_1, kwargs_list_2, joint_setting_list):
...
+        :param kwargs_logsampling_lens: list of list of parameters to sample in log10
+        :param log_sampling_lens: [[i_1, ['param_name1', 'param_name2', ...]], [...], ...]
+        :return: updated kwargs_logsampling_lens
+        """
+        for setting in log_sampling_lens:
+            i_1, param_list = setting
+            if type(param_list) == list:
+                kwargs_logsampling_lens[i_1] = param_list
+            else:
+                raise TypeError(
```
sibirrer, line 488: can you test the raise? I know, the parameters.py file does not have these tests for the previous routines but I would like to enforce complete testing for all the PRs, thank you!

sibirrer (changes requested):
Thank you very much! I have only one minor request for testing a newly introduced raise statement, and a documentation/comment of what's happening with `kwargs_logsampling_lens`
Simon

sibirrer:
Thank you!

PR #276, opened 2021-08-02
Added NIE and CHAMELEON to list of viable 3d light models

mattgomer:
Added 'NIE', 'CHAMELEON', 'DOUBLE_CHAMELEON', and 'TRIPLE_CHAMELEON' to LightModel.light_model_base.py list of models which have viable 3d light distributions. Galkin should now work with these models without using MGE.


sibirrer (approved):
Great thanks! Do you want to add your name to the AUTHORS.rst file before merging?

PR #283, opened 2021-08-31
Number of images check and allow for covariance matrix

ewoudwempe:
Checks if the num. of images is the same for the model and the data of
flux ratios, and if not returns a very low likelihood.
And allows one to pass a covariance matrix instead of Gaussian errors.

I'll add some more documentation and some tests too...


sibirrer (changes requested):
Can you add brief test functions for the different input types of the errors and for the condition of unequal length?

PR #286, opened 2021-09-02
Added an analytical lens equation solver and caustics routine for EPL…

ewoudwempe:
Quite a large PR, still needs quite some work before it's ready.
The advantages of using it are better accuracies, and better performance. This is illustrated in the following notebook:
https://gist.github.com/ewoudwempe/f2eefc2b3e31294c34c0e85c2c56b049


lenstronomy/Analysis/epl_shear_solver.py
```diff
@@ -0,0 +1,384 @@
+__author__ = 'ewoudwempe'
+
```
sibirrer, line 2: this file is better suited in lenstronomy.LensModel.Solver

lenstronomy/Analysis/epl_shear_solver.py
```diff
@@ -0,0 +1,384 @@
...
+from lenstronomy.Util.param_util import ellipticity2phi_q, phi_q2_ellipticity, shear_cartesian2polar, shear_polar2cartesian
+from scipy.special import hyp2f1
+
+@jit()
+def alpha_epl_shear(x, y, b, q, t=1, gamma1=0, gamma2=0, Omega=None):
+    """ The complex deflection of EPL+SHEAR """
+    return alpha(x, y, b, q, t=t, Omega=Omega) + (gamma1 * x + gamma2*y) +1j * (gamma2*x -gamma1*y)
+
+@jit()
+def lenseqtop_calcs(args, phi):
```
sibirrer, line 17: re-name function to better indicate what it does

lenstronomy/Util/epl_util.py
```diff
@@ -0,0 +1,134 @@
+__author__ = 'ewoudwempe'
+
```
sibirrer, line 2: as this utilities are specific to the EPL model, it's probably better to have these functions in a separate Util folder within the LensModel module.

lenstronomy/Util/epl_util.py
```diff
@@ -0,0 +1,134 @@
...
+@jit()
+def solvequadeq(a,b,c):
+    # https://en.wikipedia.org/wiki/Loss_of_significance
+    sD=(b**2-4*a*c)**0.5
+    x1=(-b-np.sign(b)*sD)/(2*a)
+    x2=2*c/(-b-np.sign(b)*sD)
+    return np.where(b!=0,np.where(a!=0,x1,-c/b),-(-c/a)**0.5), \
+            np.where(b!=0,np.where(a!=0,x2,-c/b+1e-8),+(-c/a)**0.5)
+
+def brentq_nojit(f, xa, xb, xtol=2e-14, rtol=16*np.finfo(float).eps, maxiter=100, args=(), verbose=False):
```
sibirrer, line 46: is there a function within scipy or something similar that does the same thing? Just to avoid duplication where not necessary

sibirrer (changes requested):
This looks great! I suggest you move some files in different folders, as indicated. Perhaps you can make some function names a bit more intuitive, but perhaps added documentation will also do it. Other than that, documentation and testing is missing. I am particularly curious of tests matching for the same parameters for analytical vs numerical. Thanks a lot!

lenstronomy/Util/epl_util.py
ewoudwempe: Well, this is almost a copy of scipy's version of scipy.optimize.brentq, with a few changes to make it numba-compatible. Directly using the scipy version is unfortunately not possible while keeping everything compiled. This was the easiest way (I think several other libraries also ship their own version of it). There may be some way to avoid it, but that would require some substantial effort I think...

sibirrer: I see, so perhaps referencing in the documentation from where it comes from and why this needs to be there (numba) helps in the understanding and potential future development

lenstronomy/LensModel/Solver/epl_shear_solver.py
```diff
@@ -0,0 +1,313 @@
...
+    int_points = [p1]
+    geom = geomlinspace(1e-4,0.1,Nmeas_extra)
+
+    thpl = np.sort(np.concatenate((np.linspace(0.,np.pi, Nmeas),
+                                   *[i%np.pi-geom for i in int_points],
+                                   *[i%np.pi+geom for i in int_points],
+                                   )))
+    the = getphi(thpl, (b, t, y1, y2, q, gamma1, gamma2))
+    thetas = np.concatenate((the, the + np.pi))
+    if make_diagplot:
```
sibirrer, line 169: this part is neither tested nor documented. You may leave it there but then I suggest testing and documenting it

lenstronomy/LensModel/Solver/epl_shear_solver.py
```diff
@@ -0,0 +1,313 @@
...
+    p *= rotfact
+    many = np.asarray(pos[0]).ndim > 0
+    res = solvelenseq_majoraxis((b, t, p.real, p.imag, q,
+                                 gamma.real, gamma.imag), Nmeas=Nmeas, Nmeas_extra=Nmeas_extra,
+                                make_diagplot=make_diagplot)
+    if not many:
+        xsol, ysol = res
+        x = np.array([(xs+1j*ys)/rotfact+cen for xs, ys in zip(xsol, ysol)])
+        return x.real, x.imag
+    else:
```
sibirrer, line 232: this part is not tested. Perhaps there is a way to do it when also choosing another test example

lenstronomy/LensModel/Util/epl_util.py
```diff
@@ -0,0 +1,135 @@
+__author__ = 'ewoudwempe'
+
+import numpy as np
+from lenstronomy.Util.numba_util import jit
+
```
sibirrer, line 5: this file may benefit of better doc strings description the input and output parameters

lenstronomy/LensModel/Util/epl_util.py
```diff
@@ -0,0 +1,135 @@
...
+def solvequadeq(a,b,c):
+    """Solves a quadratic equation. Care is taken for the numerics"""
+    # https://en.wikipedia.org/wiki/Loss_of_significance
+    sD=(b**2-4*a*c)**0.5
+    x1=(-b-np.sign(b)*sD)/(2*a)
+    x2=2*c/(-b-np.sign(b)*sD)
+    return np.where(b!=0,np.where(a!=0,x1,-c/b),-(-c/a)**0.5), \
+            np.where(b!=0,np.where(a!=0,x2,-c/b+1e-8),+(-c/a)**0.5)
+
+def brentq_nojit(f, xa, xb, xtol=2e-14, rtol=16*np.finfo(float).eps, maxiter=100, args=(), verbose=False):
```
sibirrer, line 47: perhaps you can run a test directly on this functions (as it is not @jit the coverage report can tell a bit more). You can e.g. test it against scipy

sibirrer (changes requested):
This looks great! I have some minor documentation and testing requests only

lenstronomy/LensModel/Solver/epl_shear_solver.py
```diff
@@ -0,0 +1,291 @@
+__author__ = 'ewoudwempe'
+
+import numpy as np
+from lenstronomy.LensModel.Util.epl_util import min_approx, pol_to_cart, cart_to_pol, cdot, ps, rotmat, solvequadeq, brentq_inline
+from lenstronomy.Util.image_util import findOverlap
+from lenstronomy.LensModel.Profiles.epl_numba import alpha, omega
+from lenstronomy.Util.numba_util import jit
+from lenstronomy.Util.param_util import ellipticity2phi_q, phi_q2_ellipticity, shear_cartesian2polar, shear_polar2cartesian
+
```
sibirrer, line 9: ah a minor thing: do you want to keep some of those definitions private (meaning not needed/wanted to have them accessed from other files/classes? If so, it would be good re-naming those starting with a '_'. That can just tell when refactoring is happening, which functions need to be preserved.

sibirrer (approved):
Looks all good. Just wondering whether some definitions could be set as private only meant to be accessed in a single file/class

ewoudwempe:
Good idea, done :)

PR #287, opened 2021-09-02
Nfw vir trunc

sibirrer:
definitions used were r200 rather than r_vir thus naming conventions changed.
Perhaps also move file and make it not cosmology dependent


sibirrer:
This PR is not required anymore as these changes were already made in other commits

PR #298, opened 2021-11-12
Epl match pemd

sibirrer:
Improved documentation and matching of EPL and PEMD model


sibirrer:
test coverage drop only due to removal of unnecessary code blocks

Issue #188, opened 2020-06-19
multiprocess is not a required package for lenstronomy


swagnercarena:
In version 1.5.0 of Lenstronomy, multiprocess is now used. pip installing lenstronomy does not require the package multiprocess to be installed, causing an error if the fitting sequence code is called. 

https://github.com/sibirrer/lenstronomy/blob/b43b5478c9fc9d40e0d2116190960ec4d8524601/lenstronomy/Sampling/Pool/multiprocessing.py#L15

sibirrer:
multiprocess>=0.70.8
is specified in the requirements.txt file. Is the installation not reading it? Travis-ci is installing it and can execute the import statement.

swagnercarena:
It did not install multiprocess when I ran pip install lenstronomy --upgrade. I also tried installing a fresh version of lenstronomy in a new conda environment and ran into the same issue (no multiprocess).

sibirrer:
I would need more details to reproduce your behavior. Is another version of multiprocess<0.70.8 pre-installed? Did you start from an empty environment? Python version as specified by lenstronomy tests? I can only consider this as a lenstronomy issue when starting from an empty environment within the python versions specified.

swagnercarena:
I did start from an empty environment (without multiprocess installed at all) and ran into the same issue. I was using python version 3.7.6. It looks like in your setup.py file there isn't any reference to the requirements.txt file. The only requirements specified there are scipy, numpy, and configparser. The travis.yaml file instead does specifically run the install using pip install -r requirements.txt, which installs multiprocess (along with a number of other dependencies). Maybe this is a misunderstanding on my part, but I believe if the user just runs:
` pip install lenstronomy --user `
as is suggested in the readme, they will be missing packages like multiprocess.

sibirrer:
Ah, I see. The current convention (they way I interpret it, is only have the 'base' requirements in the install_requirements and other parts that might not been used by all users as 'optional' in the requirements.tex. See e.g. https://packaging.python.org/discussions/install-requires-vs-requirements/
I may consider adding more packages to the install_requirements but also don't want to over-load it.

Issue #140, opened 2019-12-09
mpmath not a requirement?


bnord:
mpmath wasn't installed when I pip installed lenstronomy.

sibirrer:
Thanks for the note!. This was an issue in an earlier version (pre 1.x) of lenstronomy. In the current version, mpmath is part of the requirements that gets installed with pip. Please re-open if you have issues with the latest version.

PR #299, opened 2021-11-15
Postpone imports

JelleAalbers:
This moves imports of packages that are not lenstronomy dependencies inside the functions where they are used. Lenstronomy already does this, e.g. for [fastell4py here](https://github.com/sibirrer/lenstronomy/blob/cad2f7a3f580090fa2d4eb6d882727be3b137969/lenstronomy/LensModel/Profiles/spemd.py#L62) and for [astropy here](https://github.com/sibirrer/lenstronomy/blob/6db785667ff099fa8338e972b66253b2901b2827/test/test_LensModel/test_MultiPlane/test_multi_plane.py#L60). This PR merely does the same at 7 other points, of which 3 are for `mpmath`.

This will make lenstronomy easier to use. For example, if you use a function from [lens_model_extensions.py](https://github.com/sibirrer/lenstronomy/blob/main/lenstronomy/LensModel/lens_model_extensions.py), you currently get a `ModuleNotFoundError` unless `scikit-image` is installed, even though only one function in that file uses `scikit-image` (`critical_curve_caustics`).

Moreover, this makes the short and laconic APIs from #214 actually useful. Currently even I am wary of them as `ls.short()` and `ls.laconic()` require you to have many funky packages installed. After this, you can use these APIs in a clean environment with only lenstronomy and astropy. All packages that still have top-level imports in lenstronomy after this PR (like `numpy`, `pyyaml` and `matplotlib`) are dependencies of either lenstronomy or astropy; if you install both, any file in lenstronomy can at least be imported. (With one exception, see below.)

Two special cases:
  * `LensModel.Profiles.gauss_decomposition` used a base class from [future](https://pypi.org/project/future/) to paper over python 2/3 metaclass differences. Since lenstronomy dropped python 2 support last year, this is no longer needed. Other ways to avoid the top-level `future` import would probably be messy.
  * `Sampling.Pool.multiprocessing` is a deep internal module (vendored from Schwimmbad according to the comments) that has a seemingly unavoidable top-level import of [multiprocessing](https://pypi.org/project/multiprocessing/). Rather than changing anything here, I excluded this file from the short and laconic APIs, as users are unlikely to use it directly (as opposed to through `lenstronomy.Pool.pool.choose_pool`). People who are using it directly can just keep importing it explicitly.

Related issues: https://github.com/sibirrer/lenstronomy/issues/140, https://github.com/sibirrer/lenstronomy/issues/188

### DIFF
--- a/lenstronomy/LensModel/Profiles/uldm.py
+++ b/lenstronomy/LensModel/Profiles/uldm.py
@@ -4,7 +4,6 @@ __author__ = 'lucateo'
 import numpy as np
 import scipy.interpolate as interp
 from scipy.special import gamma, hyp2f1
-from mpmath import hyp3f2
 from lenstronomy.LensModel.Profiles.base_profile import LensProfileBase
 import lenstronomy.Util.constants as const
 __all__ = ['Uldm']
@@ -62,6 +61,8 @@ class Uldm(LensProfileBase):
         :param center_y: center of halo (in angular units)
         :return: lensing potential (in arcsec^2)
         """
+        from mpmath import hyp3f2
+
         x_ = x - center_x
         y_ = y - center_y
         r = np.sqrt(x_** 2 + y_** 2)
--- a/lenstronomy/GalKin/velocity_util.py
+++ b/lenstronomy/GalKin/velocity_util.py
@@ -1,6 +1,5 @@
 __author__ = 'sibirrer'
 
-import mpmath as mp
 import numpy as np
 
 from lenstronomy.Util.package_util import exporter
@@ -12,6 +11,8 @@ def hyp_2F1(a, b, c, z):
     """
     http://docs.sympy.org/0.7.1/modules/mpmath/functions/hypergeometric.html
     """
+    import mpmath as mp
+
     return mp.hyp2f1(a, b, c, z)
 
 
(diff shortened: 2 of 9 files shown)


lenstronomy/LensModel/Profiles/uldm.py
```diff
@@ -62,6 +61,8 @@ def function(self, x, y, kappa_0, theta_c, center_x=0, center_y=0, slope=8):
         :param center_y: center of halo (in angular units)
         :return: lensing potential (in arcsec^2)
         """
+        from mpmath import hyp3f2
```
sibirrer, line 64: Just a question about speed: This function call may be repeated many times during a fitting sequence. Does this import statement create a lot of overhead?

sibirrer (commented):
Thanks for the PR! This is a very nice improvement. I had one question/concern and am otherwise just waiting for the tests to go through.

lenstronomy/LensModel/Profiles/uldm.py
JelleAalbers: Thanks for the lightning-fast response Simon! I think speed shouldn't be an issue, as python globally caches imported modules (in `sys.modules`). So repeated imports only add a sub-us overhead, roughly the size of a function call. Here's a test with a huge package, tensorflow:

Screenshot from 2021-11-15 12-18-08

sibirrer: Thanks for the quick test! You convinced me :)

PR #301, opened 2021-11-22
Astropy v5 compatibility

sibirrer:
makes lenstronomy fully compatible with astropy v5 and v4 by choosing different underlying function calls


sibirrer:
test coverage only dropping due to astropy<5 code compatibility

PR #305, opened 2022-01-22
Hotfix/background rms array

ajshajib:
I changed one line to make `background_rms` to be treated as an array instead of a scalar. This is necessary to avoid an error, when a simulated image is created using `SimulationAPI` and a map is provided as `exposure_time` instead of a single value.


sibirrer (approved):
Looks good, thanks for the PR!

PR #309, opened 2022-03-07
changing ellipticity definition for the potential

sibirrer:
changing ellipticity definition for the potential in NFW_ELLIPSE, PJffeEllipse and SERSIC_ELLIPSE_POT models


lenstronomy/LensModel/Profiles/p_jaffe_ellipse.py
```diff
@@ -79,8 +68,8 @@ def mass_3d_lens(self, r, sigma0, Ra, Rs, e1=0, e2=0):
         :param sigma0:
         :param Ra:
         :param Rs:
-        :param q:
-        :param phi_G:
+        :param e1:
```
mattgomer, line 71: Are these e1 and e2 parameters even used in this function?

lenstronomy/LensModel/Profiles/p_jaffe_ellipse.py
```diff
@@ -25,14 +25,7 @@ def function(self, x, y, sigma0, Ra, Rs, e1, e2, center_x=0, center_y=0):
...
         """
-        phi_G, q = param_util.ellipticity2phi_q(e1, e2)
-        x_shift = x - center_x
-        y_shift = y - center_y
-        cos_phi = np.cos(phi_G)
-        sin_phi = np.sin(phi_G)
-        e = min(abs(1. - q), 0.99)
-        x_ = (cos_phi*x_shift+sin_phi*y_shift)*np.sqrt(1 - e)
-        y_ = (-sin_phi*x_shift+cos_phi*y_shift)*np.sqrt(1 + e)
+        x_, y_ = param_util.transform_e1e2_square_average(x, y, e1, e2, center_x, center_y)
```
mattgomer, line 28: I'm not very familiar with p_jaffe. Is ellipticity added in potential like the other cases? If it is added in mass, there may be an inconsistency.

lenstronomy/Util/param_util.py
```diff
@@ -146,7 +146,19 @@ def transform_e1e2_square_average(x, y, e1, e2, center_x, center_y):
...
     return x_, y_
+
+
+def q2e(q):
+    """
+
+    :param q:
+    :return:
+    """
+    e = (1 - q ** 2) / (1 + q ** 2)
```
mattgomer, line 162: Maybe the numerator needs and abs(). Not sure why abs(1-q) was used before. Are there ever cases with q>1?

sibirrer: indeed, that can be done for more stability

lenstronomy/LensModel/Profiles/p_jaffe_ellipse.py
sibirrer: yes, it's in potential. I added a detailed description of the profile

lenstronomy/LensModel/Profiles/p_jaffe_ellipse.py
sibirrer: it is added because the input function does come (potentially) with this parameter

sibirrer:
@mattgomer I updated lenstronomy and fixed some tests. The test coverage reduced because the removal of redundant code lines. Let me know if it is ok or not! Thank you!

mattgomer (approved):
Looks good, provided the hard-coded values in the test files are correct.

PR #315, opened 2022-04-11
Add feature to sample distance ratios in multi-plane case

ajshajib:
@sibirrer, I have added a new function to update `T_ij` values based on sampled parameters `a` and `b`. Can you kindly look into what would be the proper way to pass the sampled values to this class instance? 

A few other points I am not sure about and will require pointers from you:

- The `T_ij` values only seem to go upto the last lens plane? Where is the dependence on `T_ij` between the last lens plane and the source plane accounted for in the code?
- On a similar note, I think, the distances `D_is` between i-th lens plane and the source plane should also matter. The sampled `b` values would inform these distances. I didn't see where these distances are contributing to the calculation.
- For a multi-source plane case, where are the deflection angles rescaled based on the source redshifts or distances?


sibirrer:
Thank you @ajshajib ! The dependence o the last lens plane and the source plane is currently [here](https://github.com/sibirrer/lenstronomy/blob/4a1be552ba95d48c20d4687c00c024e2ca398f69/lenstronomy/ImSim/image2source_mapping.py#L63) separate from the LensModel module since it relates source and lens together.

For the multi-source plane lensing, the reduced deflectors are between the specific lens redshift and the [z_source_convention parameter](https://github.com/sibirrer/lenstronomy/blob/4a1be552ba95d48c20d4687c00c024e2ca398f69/lenstronomy/LensModel/lens_model.py#L37) . If this parameter is not explicitly set, then it is z_source in the lens model class

I like your proposal. I am not sure whether the top-level interface should go through the image2source_mapping class or to re-structure the class itself. Feel free to propose a solution for it as long as it's backwards compatible, if possible

sibirrer:
Thank you @ajshajib ! The dependence o the last lens plane and the source plane is currently [here](https://github.com/sibirrer/lenstronomy/blob/4a1be552ba95d48c20d4687c00c024e2ca398f69/lenstronomy/ImSim/image2source_mapping.py#L63) separate from the LensModel module since it relates source and lens together.

For the multi-source plane lensing, the reduced deflectors are between the specific lens redshift and the [z_source_convention parameter](https://github.com/sibirrer/lenstronomy/blob/4a1be552ba95d48c20d4687c00c024e2ca398f69/lenstronomy/LensModel/lens_model.py#L37) . If this parameter is not explicitly set, then it is z_source in the lens model class

I like your proposal. I am not sure whether the top-level interface should go through the image2source_mapping class or to re-structure the class itself. Feel free to propose a solution for it as long as it's backwards compatible, if possible

ajshajib:
@sibirrer, In the `Image2SourceMapping` [class](https://github.com/sibirrer/lenstronomy/blob/4a1be552ba95d48c20d4687c00c024e2ca398f69/lenstronomy/ImSim/image2source_mapping.py#L7), `def image2source()` doesn't use the `T_ij_start` and `T_ij_end` variables, but the other functions `def image_flux_split()` and `def image_flux_joint()` do. Is this consistent/intended?

sibirrer:
> @sibirrer, In the `Image2SourceMapping` [class](https://github.com/sibirrer/lenstronomy/blob/4a1be552ba95d48c20d4687c00c024e2ca398f69/lenstronomy/ImSim/image2source_mapping.py#L7), `def image2source()` doesn't use the `T_ij_start` and `T_ij_end` variables, but the other functions `def image_flux_split()` and `def image_flux_joint()` do. Is this consistent/intended?

Indeed, you can put in the T_ij_start and T_ij_end there as well. It's just not as relevant as there is only one start and stop, but for the other definitions there can potentially b several.

sibirrer:
thanks @ajshajib . It looks like a lot of files changed  in this PR due to the pull from the current main branch through your branch. Would it be useful if I pull from the main branch into this development branch or you could directly make a PR to the main branch?

ajshajib:
I can directly make a PR to the main branch, when it's ready. I am currently working on this, and I just needed to merge the main branch into it so that `astropy` versions etc. don't throw a fit in my computer. 🙂 I will ping you when it's ready for your review or need feedbacks.

PR #323, opened 2022-07-08
Adding zeus as an alternative MCMC sampler

nataliehogg:
The zeus package has been added to the Sampler class in a new function called mcmc_zeus(). Zeus can be used by declaring 'sampler_type': 'ZEUS' in the dictionary passed to Fitting Sequence.

### DIFF
--- a/lenstronomy/Sampling/sampler.py
+++ b/lenstronomy/Sampling/sampler.py
@@ -118,7 +118,8 @@ class Sampler(object):
             print('===================')
         return result, [chi2_list, pos_list, vel_list]
 
-    def mcmc_emcee(self, n_walkers, n_run, n_burn, mean_start, sigma_start, mpi=False, progress=False, threadCount=1,
+    def mcmc_emcee(self, n_walkers, n_run, n_burn, mean_start, sigma_start,
+                   mpi=False, progress=False, threadCount=1,
                    initpos=None, backend_filename=None, start_from_backend=False):
         """
         Run MCMC with emcee.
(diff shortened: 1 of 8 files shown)


lenstronomy/Sampling/sampler.py
```diff
@@ -190,3 +190,59 @@ def mcmc_emcee(self, n_walkers, n_run, n_burn, mean_start, sigma_start, mpi=Fals
...
+        :return: samples, ln likelihood value of samples
+        :rtype: numpy 2d array, numpy 1d array
+        """
+        # todo:
+        # check that it's ok to pass the truncated sample ball from sampling_util as initpos
+        # add the MPI functionality
+        # add any other options present in zeus like light_mode
+        # add ability to plot zeus chains to chain_plot
+
+        import zeus
```
sibirrer, line 223: you probably need to put zeus in the requirements.txt and the setup.py (there in the same part as e.g. dynesty). I guess the tests fill fail as it can't find zeus in the environment that is created.

test/test_Sampling/test_sampler.py
```diff
@@ -117,9 +117,33 @@ def test_mcmc_emcee(self):
                                                     backend_filename=backup_filename, start_from_backend=True)
         assert len(samples_2) == len(samples_1) + n_walkers * n_run
         assert len(dist_2) == len(samples_2)
-        
+
         os.remove(backup_filename)  # just remove the backup file created above
 
+    def test_mcmc_zeus(self): # NHmod
+        n_walkers = 44
```
sibirrer, line 124: do you need that many walkers for testing? It's just that if tests run faster it is more convenient :)

sibirrer (changes requested):
Looks great! Tests will likely fail as zeus was not added to the environment.
I only have minor comments in addition. Let me know if you have questions!

lenstronomy/Sampling/sampler.py
```diff
@@ -190,3 +190,59 @@ def mcmc_emcee(self, n_walkers, n_run, n_burn, mean_start, sigma_start, mpi=Fals
             time_end = time.time()
             print(time_end - time_start, 'time taken for MCMC sampling')
         return flat_samples, dist
+
+    def mcmc_zeus(self, n_walkers, n_run, n_burn, mean_start, sigma_start, progress=False, initpos=None, backend_filename=None): # NHmod
+        """
+        Lightning fast MCMC with zeus.
```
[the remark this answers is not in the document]
nataliehogg: Sure thing!

lenstronomy/Sampling/sampler.py
nataliehogg: I have added zeus to the requirements and setup, though the tests still fail because it can't find the back up file which is created for the chains when it looks to remove it (line 145 in test_sampler.py). I'm not sure how I can debug this without running the full tests each time, any tips?

nataliehogg: Ah, I think I solved it, it's because by default zeus only saves to file after 100 steps and I'm only running for 4 in the test. I'll set the default for saving to file to one.

sibirrer:
who, the tests run and pass @nataliehogg!
Can you add a test function for running zeus through the FittingSequence() class too? See the test_fitting_sequence.py file.

sibirrer (approved):
Wonderful! I can approve it :)

PR #327, opened 2022-07-13
Roman pixel psf support

Vikramb1:


sibirrer (changes requested):
Looks great! Can you add a test for this feature please.
Thank you very much!

Vikramb1:
Hi Simon, I think I added the correct test case just now.

sibirrer (approved):

PR #334, opened 2022-08-10
nfw_ellipse_cse profile ellipticity normalization changed

sibirrer:
This is an attempt to have the NFW_ELLIPSE_CSE profile normalized for elliptical profiles. Currently this profile was introduced in the semi-major axis and the rescaling did not lead to results comparable with NFW_ELLIPSE (which defines the ellipticity in the potential).


mattgomer:
The Einstein radius changes less with q, but still changes. 
image
image
What is the rationale for dividing by c as opposed to q or some power of q? Is there an equation for this somewhere?

mattgomer:
So it's not motivated by a known formula, it's just an attempt to match empirically. It looks pretty good: I can try other attempts, but I doubt I can get much better. 
I see from the tests that the CSE itself is done through semi-major axis (in that no matter what you change e1 to, kappa_major=kappa_round in `test_cored_steep_ellipsoid.test_ellipticity()`.) This probably explains why in `test_nfw_ellipse_cse.test_ellipticity()`, kappa_major is a constant factor above kappa_round. But since kappa_minor is not a constant factor below, the combination is never equal to kappa_round. I don't think this can be fixed with a normalization. We'd have to modify cored_steep_ellipsoid to be wrt. an average radius instead of wrt. major axis. Do you think this could work?

sibirrer:
> I see from the tests that the CSE itself is done through semi-major axis (in that no matter what you change e1 to, kappa_major=kappa_round in `test_cored_steep_ellipsoid.test_ellipticity()`.) This probably explains why in `test_nfw_ellipse_cse.test_ellipticity()`, kappa_major is a constant factor above kappa_round. But since kappa_minor is not a constant factor below, the combination is never equal to kappa_round. I don't think this can be fixed with a normalization. We'd have to modify cored_steep_ellipsoid to be wrt. an average radius instead of wrt. major axis. Do you think this could work?

Yes, I think making the CSE profiles wrt. to an average radius is probably the right approach. Whether or not that changes the radial shape (a bit) in the elliptical regime has to be seen. I don't have time to look into this right now, though. I leave it up to you to what level you find the CSE function useful or want to try it.

Thanks a lot!

mattgomer:
I added a version of CSE that uses the product averaged r instead of the major axis. This reduces the error in the NFW_ELLIPSE_CSE from a max of 36% to a max of 2% (when q=0.3). The remaining mismatch still fails tests. 

Here is what I think is happening: It comes down to the CSE function mismatching worst at r=s. I think, based off the attached image, that the problem is that a circular case with circular core does not exactly match an elliptical case with a circular core. Really the core should be somehow slightly different in the y direction compared to the x direction, which leads to the greatest mismatch of about 7% at the core radius itself (for q=0.3, worsens for higher ellipticity). Is a systematic 1%  (max 2%) difference in the NFW_ELLIPSE_CSE function a problem? Do you see any solutions?
CSE_ProductAvg_kappa

lenstronomy/LensModel/Profiles/nfw_ellipse_cse.py
```diff
@@ -167,6 +169,7 @@ def _normalization(self, alpha_Rs, Rs, q):
         :return: normalization (m)
         """
         rho0 = self.nfw.alpha2rho0(alpha_Rs, Rs)
-        rs_ = Rs / np.sqrt(q)
-        const = 4 * rho0 * rs_ ** 3
+        c = 2 * q / (1 + q)  # this is the same as 1 - e with e = (1. - q) / (1. + q)
+        rs_ = Rs  #/ np.sqrt(q)
```
dangilman, line 173: it appears the constant "c" in the line above never gets used after it is defined

dangilman (commented):
Hi @sibirrer @mattgomer, I think the only way to determine the correct prefactor to normalize the profiles (beyond guess and check) would be to actually do the integrals, which probably don't have closed form solutions. Your approach that yields the desired behavior up to 3% seems reasonable to me, but I suppose the precision you need will depend on your application.

lenstronomy/LensModel/Profiles/nfw_ellipse_cse.py
mattgomer: Thanks, I commented out  rather than deleting the c because it's not necessary now, but it is still required if you were to change the cse_major_axis_set to CSEMajorAxisSet(), which is how it was before.

mattgomer:
I added a test to check that the Einstein radius doesn't change with ellipticity. It's only accurate to within 1 decimal. But it's better than before: off by 0.02 whereas before I think it was off by 0.1 or so. I think this is good enough for the present applications, but we may want to add a note somewhere to state that we know this is only accurate to xx%.

mattgomer:
I think these failed checks come from taking a derivative wrt. x (or y), when I have rescaled x and y by a factor of sqrt(q), so the derivative should maybe be wrt x_new=sqrt(q)x, or something like this. I'll play with the algebra a bit and see if I can make sense of it. It looks like a constant factor between the analytical and numerical Hessians.

lenstronomy/LensModel/Profiles/cored_steep_ellipsoid.py
```diff
@@ -249,3 +257,135 @@ def hessian(self, x, y, a_list, s_list, q):
...
+        :param y: coordinate in image plane (angle)
+        :param a: lensing strength
+        :param s: core radius
+        :param q: axis ratio
+        :return: lensing potential
+        """
+        x = x * np.sqrt(q)
+        y = y * np.sqrt(q)
+        # potential calculation
+        psi = np.sqrt(q**2*(s**2 + x**2) + y**2)
```
sibirrer, line 286: I would also scale 's' to be the product average smoothing radius.

lenstronomy/LensModel/Profiles/cored_steep_ellipsoid.py
```diff
@@ -249,3 +257,135 @@ def hessian(self, x, y, a_list, s_list, q):
...
+        """
+
+        :param x: coordinate in image plane (angle)
+        :param y: coordinate in image plane (angle)
+        :param a: lensing strength
+        :param s: core radius
+        :param q: axis ratio
+        :return: lensing potential
+        """
+        x = x * np.sqrt(q)
```
sibirrer, line 283: since these re-scalings or 'x', 'y' (and 's') are common among all functions in this class, I would dedicate a single definition to those parameter conversions.

lenstronomy/LensModel/Profiles/cored_steep_ellipsoid.py
```diff
@@ -249,3 +257,135 @@ def hessian(self, x, y, a_list, s_list, q):
             f_xy += f_xy_
             f_yy += f_yy_
         return f_xx, f_xy, f_xy, f_yy
+
+class CSEProductAvg(LensProfileBase):
```
sibirrer, line 261: In this class, I would not duplicate the calculations from the MajorAxis part but instead use one single definition applied to both calculations

sibirrer:
Thanks @mattgomer for looking into this! Indeed, the derivative scaling might be the issue. I also highly suggest re-scaling 's' such that 's' will have the meaning of the product average as well (such that the term (s^2 + x^2/q + qy^2) is the relevant one)

lenstronomy/LensModel/Profiles/cored_steep_ellipsoid.py
mattgomer: I wasn’t sure if you meant that this class should be replaced, and just be an argument within what is currently called CSEMajorAxis(), or if you meant that this class should exist, but just call CSEMajorAxis() instead of copying and pasting all the functions. I interpreted this the second way, but I can instead write all of this in one class with an input option specifying the axis if you think that architecture is cleaner

lenstronomy/LensModel/Profiles/cored_steep_ellipsoid.py
mattgomer: I believe that s is already the correct radius and doesn't need renormalized, based on a few checks. If I try to adjust it by a factor of sqrt(q), the whole normalization is thrown way off.

PR #349, opened 2022-08-21
minor updates on code and documentation

sibirrer:
The nautilus version 0.1.0 had some bugs that got fixed in version 0.2.0 and the latest changes make it possible to sample with nautilus 0.2.0


sibirrer:
@johannesulf it sill does not work since nautilus complains now about the dimension argument, thought I did not change anything from v 0.1.0 with explicitly sampling two dimensions.

PR #350, opened 2022-08-22
General parameter scaling

jhod0:
# General parameter scaling

This is an implementation of a feature I have discussed with Simon and Anowar. I added a few tests which pass but have not played with it too much yet.

It also includes a small refactor of how parameters are handled, which made it easier for me to implement this. The interface remains the same.

## Explanation

I am working on modeling lensing galaxy clusters and would like to fit the member mass profiles together in a uniform way, scaling their overall mass and radius by their magnitude. Something similar to these equations from [Bergamini 2019](https://ui.adsabs.harvard.edu/abs/2019A%26A...631A.130B/abstract):

image

The existing `mass_scaling` parameter is similar but not exactly what I need; it only scales the overall mass linearly.

This PR implements generalized scaling, where any lens parameter can be scaled like this::

```
kwargs_lens[i]['param'] = param_scale * kwargs_lens[i]['param'] ** param_power
```

Where both `param_scale` and `param_power` can be fixed or sampled.

## Use

The interface is similar to `mass_scaling`, a mask is provided for each scaled parameter listing which lenses will be scaled.

This example will scale the `sigma0` and `Rs` parameters of `PJAFFE` profiles, like in the Bergamini example above. The member lens masses are labeled with `1`.

```
kwargs_constraints = {
    'general_scaling': {
        'sigma0': [False]*num_halo + [1]*nmembers,
        'Rs': [False]*num_halo + [1]*nmembers,
    }
}
```

And the initial values and bounds can be provided like this. This fixes the power-law exponent but varies the overall scaling for both `sigma0` and `Rs`.

```
fixed_special = {'sigma0_scale_pow': [alpha*2], 'Rs_scale_pow': [beta]}
kwargs_special_init = {'sigma0_scale_factor': [17.0], 'Rs_scale_factor': [8]}
kwargs_special_sigma = {'sigma0_scale_factor': [10.0], 'Rs_scale_factor': [3]}
kwargs_lower_special = {'sigma0_scale_factor': [0.5], 'Rs_scale_factor': [1]}
kwargs_upper_special = {'sigma0_scale_factor': [40], 'Rs_scale_factor': [20]}

special_params = [kwargs_special_init, kwargs_special_sigma, fixed_special, kwargs_lower_special, kwargs_upper_special]
```

PR #352, opened 2022-08-25
Refactor sampling parameter handling

jhod0:
This PR would continue the refactor of parameter handling implemented in #350 

It will reimplement the following two parameter classes with the framework in `Sampling/param_group.py`:

[x] Refactor LightParams 
[ ] Refactor LightParams

Of course, this is low-priority. It adds no features but would hopefully ease maintenance in the future, and make it simpler to add new sampled parameters to these classes.

Just opening the PR for discussion

### DIFF
--- a/lenstronomy/LensModel/lens_param.py
+++ b/lenstronomy/LensModel/lens_param.py
@@ -1,7 +1,12 @@
 from lenstronomy.LensModel.single_plane import SinglePlane
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


sibirrer:
@jhod0 thanks a lot! This is super helpful. There are some tests failing in the CI. You can also test locally, e.g.
>>> py.test test/test_Workflow

PR #353, opened 2022-08-26
updates in numerical lens equation solver to allow for more candidate solutions

sibirrer:
The current local minima finder for candidate solutions to the lens equations check in the off-diagonal axis beyond the neighboring points whether there is a local minima. This was intended to avoid many candidate solutions along a highly magnified region that all result eventually to the same solution.
However, there are examples where this fails and no candidate solution is provided and hence some images are missing.

Below is an example of this.

The proposed PR simplifies the candidate solution calculation (also speed-up) resulting in generally more candidate solutions that are being investigated (more computational costs). I think this is a better trade-off.

#lens model
lens_model_list = ['EPL_NUMBA','SHEAR']
lens_model = LensModel(lens_model_list=lens_model_list)
kwargs_lens = [
    {
        'theta_E':0.7666888232,
        'gamma':2.0269779852,
        'e1':-0.2836849885,
        'e2':0.0991617441,
        'center_x':0.0892853015,
        'center_y':0.0845201544
    },
    {
        'gamma1':-0.2211806041,
        'gamma2':-0.026305004,
        'ra_0':0.0,
        'dec_0':0.0
    }
]

kwargs_ps = [
    {
        'ra_source':0.1122538938,
        'dec_source':0.0304589556,
        'mag_pert':[1.221,1.008,0.933,0.838],
        'source_amp':15.030721591529167
    }
]
solver = LensEquationSolver(lens_model)

lens_equation_params={
    'min_distance':0.03999594047039023,
    'search_window':2.5597401901049746,
    'x_center':0.0199923838,
    'y_center':-0.0200037159
}

x_image, y_image = solver.image_position_from_source(kwargs_ps[0]['ra_source'], 
                                                     kwargs_ps[0]['dec_source'], 
                                                     kwargs_lens, verbose=True,
                                                     **lens_equation_params)
print(x_image, y_image)

Credits to @smericks for the undesired behavior and the example posted here.

### DIFF
--- a/lenstronomy/LensModel/Solver/epl_shear_solver.py
+++ b/lenstronomy/LensModel/Solver/epl_shear_solver.py
@@ -165,11 +165,13 @@ def _check_center(kwargs_lens):
     # calculate (inverse) displacement caused by the offset between shear and lens centroid
     # this shift needs to be added to the source position such that the solution of the lens equation
     # without this shift in the shear is the correct one
-
-    shear = Shear()
-    # calculate shift from the deflector centroid from the shear field
-    alpha_x, alpha_y = shear.derivatives(kwargs_lens[0]['center_x'], kwargs_lens[0]['center_y'], **kwargs_lens[1])
-    return alpha_x, alpha_y
+    if len(kwargs_lens) > 1:
+        shear = Shear()
+        # calculate shift from the deflector centroid from the shear field
+        alpha_x, alpha_y = shear.derivatives(kwargs_lens[0]['center_x'], kwargs_lens[0]['center_y'], **kwargs_lens[1])
+        return alpha_x, alpha_y
+    else:
+        return 0, 0
 
 
 def solve_lenseq_pemd(pos_, kwargs_lens, Nmeas=400, Nmeas_extra=80, **kwargs):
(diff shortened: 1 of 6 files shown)


dangilman (commented):
Hi @sibirrer, I noticed this weird behavior before as well, it's good to do something about it. It all looks reasonable to me, but I guess the biggest slowdown in speed is the for loop over the components of the "a" array in the local_minima_2d function. There is apparently a scipy routine called "argrelextrema" that finds the local minima of an array in python. Do you think it would be faster than the method currently implemented? See here: https://stackoverflow.com/questions/4624970/finding-local-maxima-minima-with-numpy-in-a-1d-numpy-array

smericks (approved):
It would be great to get this fixed, thank you Simon for working this out! I see that you also added analytic solutions for non-aligned shear & mass, which is great.

smericks (approved):
I meant to review all changes with my previous comment, but again thank you Simon for looking into this! For my use case, it's very helpful to get this fixed.

sibirrer:
> Hi @sibirrer, I noticed this weird behavior before as well, it's good to do something about it. It all looks reasonable to me, but I guess the biggest slowdown in speed is the for loop over the components of the "a" array in the local_minima_2d function. There is apparently a scipy routine called "argrelextrema" that finds the local minima of an array in python. Do you think it would be faster than the method currently implemented? See here: https://stackoverflow.com/questions/4624970/finding-local-maxima-minima-with-numpy-in-a-1d-numpy-array

Hi @dangilman , thanks for the pointer! The function you refer to seems to be only applicable for 1d arrays while we are interested in 2d local minima. This PR is more of a 'hot fix' for better performance. If you think this is the bottleneck and can be improved, any further PR is welcomed

sibirrer:
@dangilman if you could approve & merge or give other comments, that would be great - thank you!

sibirrer:
I decide to merge it for now as it is a bit time critical, but please come back to this PR if there are more requests/issues.

PR #362, opened 2022-09-12
changed the PSO return from chi2_list to log_likelihood list

sibirrer:
changed the PSO return from chi2_list to log_likelihood list to be consistent with other outputs of e.g. emcee and the chain_plot routines.


ajshajib (approved):

ajshajib:
Looks good to me. 🙂

PR #369, opened 2022-10-04
cmap.copy() AttributeError fix

Maverick-Oh:
Hello,
`plot_util.py` makes an error while copying `cmap`.
The error message is as follows.
`AttributeError: 'LinearSegmentedColormap' object has no attribute 'copy'`

I changed the copying method from `cmap.copy()` to `copy.deepcopy(cmap)`, which does not make the same error.

Please review my first-ever pull request!
Thanks :)

### DIFF
--- a/lenstronomy/Plots/plot_util.py
+++ b/lenstronomy/Plots/plot_util.py
@@ -1,6 +1,9 @@
+import copy
+
 import numpy as np
 import math
 import matplotlib.pyplot as plt
+import copy
 
 from lenstronomy.Util.package_util import exporter
 export, __all__ = exporter()
@@ -238,7 +241,8 @@ def cmap_conf(cmap_string):
         cmap = plt.get_cmap(cmap_string)
     else:
         cmap = cmap_string
-    cmap_new = cmap.copy()
+    #cmap_new = cmap.copy()
+    cmap_new = copy.deepcopy(cmap)
     cmap_new.set_bad(color='k', alpha=1.)
     cmap_new.set_under('k')
     return cmap_new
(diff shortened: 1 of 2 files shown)


sibirrer (changes requested):
Thanks a lot @Maverick-Oh !
I was not aware of this error and the tests are passing. Can you write a short test case that did not pass but with your change is passing? You can put it in the /test/test_Plots/test_plot_util.py file

And you can also add your name and GitHub link to the AUTHORS.rst file ;)

sibirrer:
Hi @Maverick-Oh , just pinging you here to see whether you noticed my comment.

Maverick-Oh:
@sibirrer Hello Simon, thanks for checking in. Yes I saw your comment but I was delaying a reply until I make a simple test case for test_plot_util.py. Can you let me know how urgent it is? Thanks!

sibirrer:
it's not urgent, I was just checking whether you are still on it, as I don't want PR's to die out :) all good, no rush.
Thank you very much!

Maverick-Oh:
Okay Thanks more from my side 😀

Maverick-Oh:
@sibirrer Hello Simon, sorry for my late update. I wasn't sure how the test code should be--if it needs to make use of lenstronomy's other automatic plotting tools or if it can be as simple as just importing&using one part of it. I made a very simple test code.

It is just a two-line code
```
from lenstronomy.Plots.plot_util import cmap_conf
        cmap_new = cmap_conf("gist_heat")
```

Please check and confirm. Thanks!

sibirrer (approved):
Yes, that's what I was thinking about. If you were struggling/failing with this part before and your changes improved the handling in this form.
Thank you!

PR #379, opened 2022-10-25
Add primary beam to image simulation (ImageModel) and linear solver (…

nanz6:
…ImageLinearFit)

Add the primary beam to image simulation and linear solver (where the light response is computed).

The 'primary beam' is input into ImageData class and is inherited by ImageModel and further ImageLinearFit. It should be a 2d numpy array and has the same size of the data image.

Physically, the primary beam is the angular response function of each antenna of an interferometry array. Numerically, it serves as a multiplier on an unconvolved image.

In other words, a convolved interferometric image is I_convolved = (I_unconvolved . primary beam ) * PSF, where . denotes multiplication and * denotes convolution.

For more info on primary beam, consult
https://casaguides.nrao.edu/index.php/VLA_CASA_Imaging-CASA5.0.0#Primary_and_Synthesized_Beam https://casadocs.readthedocs.io/en/stable/notebooks/synthesis_imaging.html?highlight=primary%20beam%20#Antenna-Voltage/Power-Patterns:-Primary-Beam

### DIFF
--- a/lenstronomy/Data/pixel_grid.py
+++ b/lenstronomy/Data/pixel_grid.py
@@ -1,15 +1,16 @@
 import numpy as np
 from lenstronomy.Data.coord_transforms import Coordinates
+from lenstronomy.Data.angular_sensitivity import AngularSensitivity
 
 __all__ = ['PixelGrid']
 
 
-class PixelGrid(Coordinates):
+class PixelGrid(Coordinates, AngularSensitivity):
     """
     class that manages a specified pixel grid (rectangular at the moment) and its coordinates
     """
 
-    def __init__(self, nx, ny, transform_pix2angle, ra_at_xy_0, dec_at_xy_0, primary_beam=None):
+    def __init__(self, nx, ny, transform_pix2angle, ra_at_xy_0, dec_at_xy_0, antenna_primary_beam=None):
         """
 
         :param nx: number of pixels in x-axis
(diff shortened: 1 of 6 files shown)


nanz6:
The primary beam does not apply to the point source light. In other words, the simulated point source image and the light response in ImageLinearFit will not be multiplied by the primary beam.

nanz6:
To pass the test, the primary beam now is defined in PixelGrid and will be inherited by ImageData, and be used by ImageModel when invoking data_class there.

sibirrer (changes requested):
Thank you very much!
The additional feature does not compromise and previous usage. I have some minor comments about documentation and naming conventions.

One other thing is that point sources are not (yet) supported. Can you add an raise statement with an error message when this condition is hit?

Then for testing, you can put specific test functions with assert options into the /test/ folder in the same folder structure as the main lenstronomy code. The idea is to have the definitions tested at the lowest level (basically input-output tests of functions). In your case, having an example that goes through to generating an image with the primary beam incorporated would be great too.

Let me know if you have questions!

nanz6:
Physically the primary beam is introduced by the sensitivity of telescopes along different directions due to the diffraction of aperture shape, so I have made a new class called AngularSensitivity in the Data folder.
The 'angular primary beam' is defined in AngularSensitivity and get inherited by PixelGrid and further ImageData, to participate in the image simulation. But rigorously, the primary beam should be an indepdent class that paralells psf. But then we need to change a lot of places so we first just make it inherited by PixelGrid.

I also added raise warnings if the primary beam is used for point sources, both in ImageModel and ImageLinearFit.

I added one test file to test the primary beam feature. This file could further be modified to include the testing of all interferometric changes.

Added some more comments in the AngularSensitivity class about the primary beam. Although that might not be clear enough, so maybe I will make one example notebook in the future to introduce all the interferometric changes that I am attempting to add.

sibirrer (approved):
This looks great!
I think in the future we should rename the PixelGrid() classes to reflect broader instrument descriptions without the actual data vector, but that is not this PR.

There are some minor test decreases due to raise statements not tested. I can also do these after the fact. I approve this PR but let @ajshajib have a second look at it.
And if you want, you can also add your name and GitHub link to the AUTHORS.rst file (otherwise I will do it afterwards)

Thanks a lot!

lenstronomy/Data/pixel_grid.py
```diff
@@ -1,27 +1,35 @@
...
+        :param antenna_primary_beam: 2d numpy array with the same size of imaga_data;
+         more descriptions of the primary beam can be found in the AngularSensitivity class
         """
         super(PixelGrid, self).__init__(transform_pix2angle, ra_at_xy_0, dec_at_xy_0)
         self._nx = nx
         self._ny = ny
         self._x_grid, self._y_grid = self.coordinate_grid(nx, ny)
+        if antenna_primary_beam is not None:
+            pbx,pby=np.shape(antenna_primary_beam)
+            if (pbx,pby) != (nx,ny):
```
ajshajib, line 30: PEP8: 
```if (pbx, pby) != (nx, ny):```

ajshajib (approved):
Looks good to me, but I left a couple of requests to fix PEP8 issues.

Issue #380, opened 2022-10-27
`ModelPlot()` executes linear solver when `linear_solver=False`


ajshajib:
https://github.com/lenstronomy/lenstronomy/blob/c6e912500aac5d50c4812dd268dc196abb0223f3/lenstronomy/Plots/model_plot.py#L53

Is this behavior intended when `linear_solver=False`? I was trying to plot corresponding to provided `amp` values in light profiles without the linear solver to change them.

sibirrer:
HI @ajshajib, currently, I think, the plotting is mostly used by inserting directly results from PSO or MCMC chains, and those are in default mode without the amp parameters. Also, if you have multiple imaging bands, it might not be clear what bands the amp parameters refer to. I am happy to have it changed as long as there does not to be any tweaks to the outputs of the previous most likely use case.

sibirrer:
Ah, sorry, this is actually a bug. The linear_solve=False gets only applied when calculating the likelihood but not by directly computing the image with image_linear_solve(), so this needs to be fixed indeed

ajshajib:
@sibirrer Thanks for the clarification. I was proividing the `amp` values inserted into there corresponding dictionaries within `kwargs_params`. So for multiple bands it should not be ambiguous as `kwargs_model` and `kwargs_params` in combination would set the appropriate parameter handling.  Since it is possible to have `amp` treated as non-linear parameters in the likelihood during PSO/MCMC, it will be useful to not alter the `amp` values coming out of PSO or MCMC if `linear_solve = False` is set. I am just wondering about consistency between fitting and plotting with `linear_solver = False` chosen.

sibirrer:
@ajshajib do you want to assign either me or you to this issue such that we are not duplicating effort to solve it? (if you can solve it, I'd appreciate it). Thank you!

ajshajib:
@sibirrer, actually, would it be possible for you to fix it rather easily? I initially thought it would be easy to fix it by putting these two lines within an `if` statement:

```
model, error_map, cov_param, param = self._imageModel.image_linear_solve(inv_bool=True, **kwargs_params)
            check_solver_error(param)
```

[...]

PR #381, opened 2022-10-31
Model plot fix

sibirrer:
This pull request should fix bug #380 for at least a single band. A new routine to extract the 'amp' parameters from the kwargs_list was required.

### DIFF
--- a/lenstronomy/Plots/model_plot.py
+++ b/lenstronomy/Plots/model_plot.py
@@ -51,6 +51,18 @@ class ModelPlot(object):
                                                        image_likelihood_mask_list=image_likelihood_mask_list)
 
         model, error_map, cov_param, param = self._imageModel.image_linear_solve(inv_bool=True, **kwargs_params)
+        if linear_solver is False:
+            if len(multi_band_list) > 1:
+                raise ValueError('plotting the solution without the linear solver currently only works with one band.')
+            im_sim = class_creator.create_im_sim(multi_band_list, 'single-band', kwargs_model,
+                                                 bands_compute=bands_compute, linear_solver=linear_solver,
+                                                 image_likelihood_mask_list=image_likelihood_mask_list)
+            # overwrite model with initial input without linear solver applied
+            model[0] = im_sim.image(**kwargs_params)
+            # retrieve amplitude parameters directly from kwargs_list
+            param[0] = im_sim.linear_param_from_kwargs(kwargs_params['kwargs_source'], kwargs_params['kwargs_lens_light'],
+                                                       kwargs_params['kwargs_ps'])
+
 
         check_solver_error(param)
         log_l = self._imageModel.likelihood_data_given_model(source_marg=source_marg, linear_prior=linear_prior,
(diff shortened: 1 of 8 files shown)


sibirrer:
@ajshajib I still need to add some test functions. But let me know if you see issues/concerns in the meantime

sibirrer:
ok, @ajshajib , ready for review :)

lenstronomy/Plots/model_plot.py
```diff
@@ -51,6 +51,18 @@ def __init__(self, multi_band_list, kwargs_model, kwargs_params, image_likelihoo
                                                        image_likelihood_mask_list=image_likelihood_mask_list)
 
         model, error_map, cov_param, param = self._imageModel.image_linear_solve(inv_bool=True, **kwargs_params)
```
ajshajib, line 53: Lines 53-56 needs to be changed like the code below to make the new changes to work.

```
        kwargs_params_copy = copy.deepcopy(kwargs_params)
        model, error_map, cov_param, param = self._imageModel.image_linear_solve(inv_bool=True, **kwargs_params)

        if linear_solver is False:
            if len(multi_band_list) > 1:
                raise ValueError('plotting the solution without the linear solver currently only works with one band.')

            kwargs_params = kwargs_params_copy
```

test/test_Plots/test_model_plot.py
```diff
@@ -214,6 +214,19 @@ def test_check_solver_error(self):
...
+        kwargs_data = sim_util.data_configure_simple(numPix=10, deltaPix=1, background_rms=1, exposure_time=1)
+        # kwargs_data['image_data'] = np.zeros((10, 10))
+        kwargs_model = {'source_light_model_list': ['GAUSSIAN']}
+        kwargs_params = {'kwargs_lens': [],
+                         'kwargs_source': [{'amp': 2, 'sigma': 1, 'center_x': 0, 'center_y': 0}],
+                         'kwargs_ps': [], 'kwargs_lens_light': []}
+        lensPlot = ModelPlot(multi_band_list=[[kwargs_data, {'psf_type': 'NONE'}, {}]],
+                             kwargs_model=kwargs_model, kwargs_params=kwargs_params, bands_compute=[True],
+                             arrow_size=0.02, cmap_string="gist_heat", linear_solver=False)
+        lensPlot.plot_main(with_caustics=True)
```
ajshajib, line 227: I suggest checking that `amp` keyword(s) within `kwargs_params` are staying the same as the initial.

ajshajib (changes requested):
Hi @sibirrer, thanks a lot for implementing the fix! In my test, it's still not fully working, which I think can be solved by using `deepcopy`  as I suggested.

lenstronomy/Plots/model_plot.py
sibirrer: ah, great catch! I made it slightly different such that the 'amp' parameters are updated if linear_solver=True but it's not happening if =False

ajshajib (approved):

ajshajib:
Great! Thanks for implementing the requested changes, feel free to merge with the main branch. 🙂

PR #385, opened 2022-11-11
Improving zeus implementation

nataliehogg:
I added the functionality to access three of the zeus callbacks, mainly useful for convergence diagnostics on running chains: https://zeus-mcmc.readthedocs.io/en/latest/api/callbacks.html.

I also moved all the zeus kwargs to be read from a dictionary that can be optionally passed as part of the fitting_kwargs_list to FittingSequence.


lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -277,10 +275,24 @@ def mcmc(self, n_burn, n_run, walkerRatio=None, n_walkers=None, sigma_scale=1, t
...
+                                                 miniter_callback=zeus_settings.get('miniter_callback', False),
+                                                 nmin=zeus_settings.get('nmin', 1000),
+                                                 moves=zeus_settings.get('moves'), tune=zeus_settings.get('tune', True),
+                                                 tolerance=zeus_settings.get('tolerance', 0.05), patience=zeus_settings.get('patience', 5),
+                                                 maxsteps=zeus_settings.get('maxsteps', 10000), mu=zeus_settings.get('mu', 1.0),
+                                                 maxiter=zeus_settings.get('maxiter', 10000), pool=zeus_settings.get('pool', None),
+                                                 vectorize=zeus_settings.get('vectorize', False), blobs_dtype=zeus_settings.get('blobs_dtype'),
+                                                 verbose=zeus_settings.get('verbose', True), check_walkers=zeus_settings.get('check_walkers', True),
+                                                 shuffle_ensemble=zeus_settings.get('shuffle_ensemble', True), light_mode=zeus_settings.get('light_mode', False))
```
sibirrer, line 295: the .get() statement requires zeus_setting not to be None. This test fails. Can't you have **kwargs_zeus as the input of this definition?

lenstronomy/Sampling/sampler.py
```diff
@@ -239,12 +247,29 @@ def mcmc_zeus(self, n_walkers, n_run, n_burn, mean_start, sigma_start,
...
+        if miniter_callback == True:
+            cb2 = zeus.callbacks.MinIterCallback(nmin=nmin)
+        else:
+            cb2 = None
+
+        callback_list = [backend, cb0, cb1, cb2]
+
         pool = choose_pool(mpi=mpi, processes=threadCount, use_dill=True)
 
         sampler = zeus.EnsembleSampler(nwalkers=n_walkers, ndim=num_param, logprob_fn=self.chain.logL,
```
sibirrer, line 275: is it possible to have here a **kwargs_zeus statement of putting the arguments in here, to be flexible and the user knows that these arguments go directly into the Zeus code?

lenstronomy/Sampling/sampler.py
```diff
@@ -195,6 +195,14 @@ def mcmc_emcee(self, n_walkers, n_run, n_burn, mean_start, sigma_start,
     def mcmc_zeus(self, n_walkers, n_run, n_burn, mean_start, sigma_start,
                   mpi=False, threadCount=1,
                   progress=False, initpos=None, backend_filename=None,
+                  autocorrelation_callback=False,
```
sibirrer, line 198: these arguments are not documented here. Perhaps keep them as kwargs_zeus and point towards the input in the actual Zeus function where they get called as **kwargs_zeus?

sibirrer (changes requested):
Thank you @nataliehogg !
The tests fail currently. You should avoid calling kwargs.get() in inputs of function calls. I made suggestions for keeping those all (or a portion of them) as keyword arguments passed as keyword arguments.

lenstronomy/Workflow/fitting_sequence.py
nataliehogg: This function now takes **kwargs_zeus

nataliehogg:
Hi @sibirrer I tried to modify things according to your suggestion. In particular, fitting sequence now only deals with **kwargs_zeus. However, I wanted to keep the kwargs_zeus.get() structure in mcmc_zeus in Sampler, because **kwargs_zeus cannot be passed to the individual callback functions of zeus (it will give an unexpected argument error). I didn't add the full documentation of all these kwargs; do you want me to?

The tests now pass locally for me.

lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -238,6 +235,8 @@ def mcmc(self, n_burn, n_run, walkerRatio=None, n_walkers=None, sigma_scale=1, t
         :param start_from_backend: if True, start from the state saved in `backup_filename`.
          Otherwise, create a new backup file with name `backup_filename` (any already existing file is overwritten!).
         :type start_from_backend: bool
+        :param zeus_settings: dictionary of zeus-specific kwargs
```
sibirrer, line 238: the naming is different here than the input. In this definition, since it's not exclusive to the zeus sampler, I would have the kwargs_zeus input statement not as '**' but as kwargs_zeus=None and then if kwargs_zeus=None: kwargs_zeus = {} or something like that

test/test_Workflow/test_fitting_sequence.py
```diff
@@ -20,7 +20,7 @@ class TestFittingSequence(object):
     test the fitting sequences
     """
 
-    def setup_method(self):
+    def setup(self):
```
sibirrer, line 23: please change to setup_method as there were deprecation warnings

lenstronomy/Sampling/sampler.py
```diff
@@ -253,10 +302,12 @@ def mcmc_zeus(self, n_walkers, n_run, n_burn, mean_start, sigma_start,
                                        blobs_dtype=blobs_dtype, verbose=verbose, check_walkers=check_walkers,
                                        shuffle_ensemble=shuffle_ensemble, light_mode=light_mode)
 
-        sampler.run_mcmc(initpos, n_run_eff, progress=progress, callbacks=backend)
+        sampler.run_mcmc(initpos, n_run_eff, progress=progress, callbacks=callback_list)
 
         flat_samples = sampler.get_chain(flat=True, thin=1, discard=n_burn)
 
+        print(len(flat_samples))
```
sibirrer, line 309: remove

lenstronomy/Sampling/sampler.py
```diff
@@ -253,10 +302,12 @@ def mcmc_zeus(self, n_walkers, n_run, n_burn, mean_start, sigma_start,
                                        blobs_dtype=blobs_dtype, verbose=verbose, check_walkers=check_walkers,
```
sibirrer, line 302: is it possible to have it organized that you can have a keyword argument in here that is flexible to changes of the Zeus API? Meaning that it is not needed to be hardcoded in lenstronomy what the arguments are? (at least for some of those). This might also be easier for testing of lenstronomy since I guess some options of the code above with if clauses were not tested

sibirrer (changes requested):
Thank you very much!
I have some minor suggested changes but I think other than that should be good.

lenstronomy/Workflow/fitting_sequence.py
nataliehogg: Docstring updated; kwargs_zeus is not a dictionary. I'm keeping it as **kwargs_zeus because that seems cleaner (if you would prefer it to be a separate dictionary I can change the rest of the implementation).

lenstronomy/Sampling/sampler.py
nataliehogg: I have changed this to **kwargs_zeus in the definition of mcmc_zeus as well. I will update the tests so that the if statements for the callbacks get checked as well.

sibirrer (approved):
Thank you @nataliehogg !
This looks good and things can always change in the future

PR #389, opened 2022-11-19
Change raise error of psf negative pixels to warning, allow user to choose if they need the automatical psf normalization

nanz6:
Change raise error of psf negative pixels to warning, allow user to choose if they need the automatical psf normalization

4 files have been changed.

psf.py: change the raise ValueError to warnings.warn if the PSF contains negative pixels; add one more parameter 'kernel_point_source_normalisation' to control if the user want the PSF auto-normalization. The default of 'kernel_point_source_normalisation' is True, so it should not change anything if people just igore it.

kernel_util.py: change the cut_psf function to have one more parameter 'normalisation' to control if the normalization should be done by the psf_cut.

test_psf.py: delete the 'assertRaises(ValueError)' part for negative PSF pixels. It should be replaced by some catching warning functions of pytest. But I am not sure how to do that.

test_image_model_with_interferometric_changes.py: modify this test file to test the primary beam and psf (with negative pixels) in ImageModel.

### DIFF
--- a/lenstronomy/Data/psf.py
+++ b/lenstronomy/Data/psf.py
@@ -14,7 +14,8 @@ class PSF(object):
     """
 
     def __init__(self, psf_type='NONE', fwhm=None, truncation=5, pixel_size=None, kernel_point_source=None,
-                 psf_error_map=None, point_source_supersampling_factor=1, kernel_point_source_init=None):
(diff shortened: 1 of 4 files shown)


sibirrer (approved):
This looks good to me. I let @ajshajib also look through it.

ajshajib (approved):

PR #394, opened 2022-12-04
minor bug fix for critical_curve_caustic

sibirrer:
the critical curve is shifted for odd pixel grids with critical_curve_caustic() and the skimage convention. To fix this problem, the code makes the grid even. Tests added that pass (but failed before the fix). Issue reported by Astha


sibirrer:
Credit for the bug reporting goes to @AsthaAstrosoul

PR #396, opened 2022-12-09
added stretch_plot and shear_plot to lens_plot with tests

mattgomer:
Two new plots for visualization of magnification and shear effects. 

stretch_plot plots a grid of ellipses stretched in accordance with the local Jacobian eigenvalues to show the direction and magnitude of stretch for an image in any given location. 
image

shear_plot plots a line segment corresponding in length to the magnitude of the shear (both internal and external) at each location. 
image

### DIFF
--- a/lenstronomy/Plots/lens_plot.py
+++ b/lenstronomy/Plots/lens_plot.py
@@ -1,7 +1,9 @@
 
 import lenstronomy.Util.util as util
+from lenstronomy.Util.param_util import shear_cartesian2polar
 import lenstronomy.Util.simulation_util as sim_util
 import matplotlib.pyplot as plt
+from matplotlib import patches
 import numpy as np
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 from lenstronomy.LensModel.lens_model_extensions import LensModelExtensions
@@ -9,6 +11,7 @@ from lenstronomy.LensModel.Profiles.curved_arc_spp import center_deflector
 from lenstronomy.Data.imaging_data import ImageData
 from lenstronomy.Plots import plot_util
 import scipy.ndimage as ndimage
+from lenstronomy.Data.pixel_grid import PixelGrid
 
 from lenstronomy.Util.package_util import exporter
 export, __all__ = exporter()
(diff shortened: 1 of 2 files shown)


lenstronomy/Plots/lens_plot.py
```diff
@@ -451,3 +454,70 @@ def _plot_frame(ax, map, vmin, vmax, text_string):
...
+def stretch_plot(ax, lens_model, kwargs_lens,
+                       plot_grid= None, scale= 1, ellipse_color= 'k',
+                       **kwargs):
+    """
+    Plots ellipses at each point on a grid, scaled corresponding to the local Jacobian eigenvalues
+    :param ax: matplotib axis instance
+    :param plot_grid: pixelgrid instance at which to draw ellipses
+    :param scale: scales sizes of drawn ellipses, bigger number=larger
+    :param ellipse_color: color of ellipses, defaults to black
+    :param kwargs: kwargs to send to matplotlib.pyplot.matshow()
```
sibirrer, line 468: aren't these kwargs going to add_patch()? Do you also want to have other kwargs going to patches.Ellipse()?

lenstronomy/Plots/lens_plot.py
```diff
@@ -451,3 +454,70 @@ def _plot_frame(ax, map, vmin, vmax, text_string):
     _plot_frame(axes[2, 3], dphi_rad_dtan2d, vmin=0, vmax=20, text_string='dphi_rad_dtan')
 
     return f, axes
+
+
+def stretch_plot(ax, lens_model, kwargs_lens,
+                       plot_grid= None, scale= 1, ellipse_color= 'k',
```
sibirrer, line 460: formatting: can you have the format in line with the arguments of the first line?

sibirrer: minor PEP8 formatting: no space in scale=1 arguments (and others) for input default arguments

lenstronomy/Plots/lens_plot.py
```diff
@@ -451,3 +454,70 @@ def _plot_frame(ax, map, vmin, vmax, text_string):
...
+    if plot_grid==None:
+        #define default ellipse grid (20x20 spanning from -2 to 2)
+        plot_grid=PixelGrid(20, 20, np.array([[1, 0],[0, 1]])*0.2,
+                 -2, -2)
+    lme=LensModelExtensions(lens_model)
+    x_grid, y_grid = plot_grid.pixel_coordinates
+    x = util.image2array(x_grid)
+    y = util.image2array(y_grid)
+    w1, w2, v11, v12, v21, v22 = lme.hessian_eigenvectors(x, y, kwargs_lens)
+    stretch_1 = np.abs(1. / w1)  # stretch in direction of first eigenvalue (unsorted)
```
sibirrer, line 481: these stretches can go to infinity. Do you want to cap them to a certain size or not? (perhaps as an optional input argument?) Just a thought.

lenstronomy/Plots/lens_plot.py
```diff
@@ -451,3 +454,70 @@ def _plot_frame(ax, map, vmin, vmax, text_string):
...
+                             stretch_2[i] * scale/40,
+                             angle=stretch_direction[i] * 180 / np.pi,
+                             linewidth=1, fill=False, color=ellipse_color)
+        ax.add_patch(ell,**kwargs)
+    ax.set_xlim(np.min(x),np.max(x))
+    ax.set_ylim(np.min(y), np.max(y))
+    return ax
+
+def shear_plot(ax, lens_model, kwargs_lens,
+                       plot_grid= None, scale= 5, color= 'k',
```
sibirrer, line 496: same with formatting

lenstronomy/Plots/lens_plot.py
```diff
@@ -451,3 +454,70 @@ def _plot_frame(ax, map, vmin, vmax, text_string):
...
+                       **kwargs):
+    """
+    Plots shear (internal+external) pseudovectors at each point on a grid
```
sibirrer, line 499: can you give a bit more detailed information about this routine? Also, leave a space between the description and the parameter definitions

test/test_Plots/test_lens_plot.py
```diff
@@ -70,6 +70,18 @@ def test_curved_arc_illustration(self):
         lens_plot.curved_arc_illustration(ax, lensModel, kwargs_lens)
         plt.close()
 
+    def test_stretch_plot(self):
+        f, ax = plt.subplots(1, 1, figsize=(4, 4))
+        lensModel = LensModel(lens_model_list=['SIE'])
+        kwargs_lens = [{'theta_E': 1, 'e1': 0.2, 'e2': 0., 'center_x': 0, 'center_y': 0}]
+        lens_plot.stretch_plot(ax, lensModel, kwargs_lens)
+        plt.close()
+    def test_shear_plot(self):
```
sibirrer, line 79: space between definitions

sibirrer (changes requested):
Looks great, thank you @mattgomer ! I left some minor formatting and style requests but otherwise ready to go/

lenstronomy/Plots/lens_plot.py
mattgomer: Good idea. Added a max_stretch argument that caps the stretches

lenstronomy/Plots/lens_plot.py
mattgomer: Updated. Is this better?

lenstronomy/Plots/lens_plot.py
mattgomer: I don't know of a way to have two mappable kwargs like **kwargs_patch and **kwargs_ax. I did play with it a bit and found that **kwargs_patch seems much more useful, so I changed it as such. Do you think a **kwargs_ax is necessary? If so, how can it be done?

lenstronomy/Plots/lens_plot.py
mattgomer: I don't know of a way to have two mappable kwargs like **kwargs_patch and **kwargs_ax. I did play with it a bit and found that **kwargs_patch seems much more useful, so I changed it as such. Do you think a **kwargs_ax is necessary? If so, how can it be done?

sibirrer: It's ok for now. We add the feature when it's requested or being used. The idea was to have the function arguments as kwargs_blabla but without the '**' in this function call so it gets passed as a keyword argument instead as unpack as arguments. Then you can do several of them.

lenstronomy/Plots/lens_plot.py
sibirrer: almost, the idea is that when a new line starts that it starts at the same argument of the function call (meaning right under the '(' of the first line. It's ok and I adopt it.

sibirrer (approved):

PR #400, opened 2023-01-06
add interferometry natwt likelihood and linear solver functions (2nd PR try)

nanz6:
Add interferometry natural weighting likelihood function and linear solver function. With this pull request merged, lenstronomy is able to do the parametric fitting for interferometric images.

Changes speciafically:

ImageData:
1. Variables added: log_likelihood_constant and likelihood_method
i. log_likelihood_constant allows user to input a float that will added to the output of the log_likelihood function.
ii. likelihood_method allows user to choose the likelihood method, the default is 'diagonal', which will compute the likelihood in the normal way of lenstronomy. If the user set this parameter to 'interferometry_natwt', it will compute the logL using our natwt special method.
2. Changed function:
i. log_likelihood. added a if to decide which likelihood method to use, by default it should use the normal one.
ii. added likelihood_method function, which passes 'likelihood_method' to ImageLinearFit to decide which linear solver method to use.
3. Raise error if likelihood_method is not input as 'diagonal' or 'interferometry_natwt'.

ImageLinearFit:
1. Changed _image_linear_solve function. Added a if to decide which linear solver to use.
2. Added _image_linear_solve_natwt_special function, where it executes the linear solving for our natwt method.

test_imaging_data.py:
1. Added one more raiseError checking. If the likelihood_method is not input as 'diagonal' or 'interferometry_natwt', it should raise a ValueError.

Added and fix the test files:
test_imaging_data_with_interferometric_changes.py
test_image_linear_solve_with_interferometric_changes.py
test_image_model_with_interferometric_changes.py
to test the interferometry natwt likelihood and linear solver changes.

Change variables names controling PSF auto normalization updated last time:
Changed 'normalisation' to 'normalization' in kernel_util.py and psf.py

### DIFF
--- a/lenstronomy/ImSim/image_linear_solve.py
+++ b/lenstronomy/ImSim/image_linear_solve.py
@@ -1,6 +1,7 @@
 from lenstronomy.ImSim.image_model import ImageModel
 import lenstronomy.ImSim.de_lens as de_lens
 from lenstronomy.Util import util
+from lenstronomy.ImSim.Numerics.convolution import PixelKernelConvolution
 import numpy as np
 
 __all__ = ['ImageLinearFit']
@@ -49,6 +50,10 @@ class ImageLinearFit(ImageModel):
         if self._pixelbased_bool is True:
             # update the pixel-based solver with the likelihood mask
             self.PixelSolver.set_likelihood_mask(self.likelihood_mask)
+            
+        # prepare to use fft convolution for the natwt linear solver 
+        if self.Data.likelihood_method() == 'interferometry_natwt':
+            self._convolution = PixelKernelConvolution(kernel = self.PSF.kernel_point_source)
 
     def image_linear_solve(self, kwargs_lens=None, kwargs_source=None, kwargs_lens_light=None, kwargs_ps=None,
                            kwargs_extinction=None, kwargs_special=None, inv_bool=False):
@@ -88,13 +93,23 @@ class ImageLinearFit(ImageModel):
             model, model_error, cov_param, param = self.image_pixelbased_solve(kwargs_lens, kwargs_source, 
                                                                                kwargs_lens_light, kwargs_ps, 
                                                                                kwargs_extinction, kwargs_special)
-        else:
+        elif self.Data.likelihood_method() == 'diagonal':
             A = self._linear_response_matrix(kwargs_lens, kwargs_source, kwargs_lens_light, kwargs_ps, kwargs_extinction, kwargs_special)
             C_D_response, model_error = self._error_response(kwargs_lens, kwargs_ps, kwargs_special=kwargs_special)
             d = self.data_response
             param, cov_param, wls_model = de_lens.get_param_WLS(A.T, 1 / C_D_response, d, inv_bool=inv_bool)
             model = self.array_masked2image(wls_model)
             _, _, _, _ = self.update_linear_kwargs(param, kwargs_lens, kwargs_source, kwargs_lens_light, kwargs_ps)
+        elif self.Data.likelihood_method() == 'interferometry_natwt':
+            # 'interferometry_natwt' method does not support the correct use of model_error, cov_param.
+            A = self._linear_response_matrix(kwargs_lens, kwargs_source, kwargs_lens_light, kwargs_ps, kwargs_extinction, kwargs_special, unconvolved=True)
+            d = self.data_response
+            model, param = self._image_linear_solve_natwt_special(A, d)
+            model_error = 0 # just a place holder 
+            cov_param = None # just a place holder
+            _, _, _, _ = self.update_linear_kwargs(param, kwargs_lens, kwargs_source, kwargs_lens_light, kwargs_ps)
+        else:
+            raise ValueError("likelihood_method %s not supported!" % self.Data.likelihood_method())
         return model, model_error, cov_param, param
 
     def image_pixelbased_solve(self, kwargs_lens=None, kwargs_source=None, kwargs_lens_light=None, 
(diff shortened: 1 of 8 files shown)


nanz6:
It seems this 2nd PR passes the test, so I have closed the 1st one (https://github.com/lenstronomy/lenstronomy/pull/399).
Though I thought they should do the same things..

lenstronomy/ImSim/image_linear_solve.py
```diff
@@ -88,13 +93,23 @@ def _image_linear_solve(self, kwargs_lens=None, kwargs_source=None, kwargs_lens_
...
             C_D_response, model_error = self._error_response(kwargs_lens, kwargs_ps, kwargs_special=kwargs_special)
             d = self.data_response
             param, cov_param, wls_model = de_lens.get_param_WLS(A.T, 1 / C_D_response, d, inv_bool=inv_bool)
             model = self.array_masked2image(wls_model)
             _, _, _, _ = self.update_linear_kwargs(param, kwargs_lens, kwargs_source, kwargs_lens_light, kwargs_ps)
+        elif self.Data.likelihood_method() == 'interferometry_natwt':
+            # 'interferometry_natwt' method does not support the correct use of model_error, cov_param.
+            A = self._linear_response_matrix(kwargs_lens, kwargs_source, kwargs_lens_light, kwargs_ps, kwargs_extinction, kwargs_special, unconvolved=True)
+            d = self.data_response
+            model, param = self._image_linear_solve_natwt_special(A, d)
```
sibirrer, line 107: I am surprised the error covariance matrix does not play a role in the linear inversion (C_D_response). There has to be an option for having a weighted linear least square method with different weights for each pixel (see de_lens.get_param_WLS() method for comparison)

sibirrer (changes requested):
Thank you very much for this PR @nanz6!
I have some minor comments/requests for changes. The largest, where I am a bit unsure, is that the individual pixel errors are not considered in the interferometric least square minimization.

nanz6:
Hi @sibirrer, I have made the changes suggested by you. 
Specifically these further changes are

1. giveup the change from 'normalisation' to 'normalization' of psf, maybe it will be done in another PR.

2. add more comments as required, change logL_constant and logL_method to be private, remove the space beside the '='  in the definition of functions.

3. move the logL_constant to be added for the interferometric logL only, since it will only be used for this mode.

4. add more comments in the _image_linear_solve_natwt_special function to explain the method. Although I think we should explain it more clearly in docs. 

For the noise covariance matrix, the short answer is that we have used it implicitly in our method. Because for the natural weighting (natwt for short) image, the pixel-pixel noise covariance matrix is encoded by the PSF. So in our method, we have played between unconvolved and convolved light profiles, which will extract out the noise covariance matrix and use it in the correct way. 

For each pixel, as far as I know, each individual pixel of interferometric image does not make much sense because they just make sense in the Fourier space. We can only use the non-sparse cov matrix to compute the chi^2, without thinking about the pixel-wise noises.

Further, I think I should write done the method very clearly in the lenstronomy documentations, if I want more people to know and use it for interferometric fitting.

nanz6:
By the way, the 'likelihood_method' variable might seem irrelevant in the ImageData class, as (maybe the entire interferometric changes as well) it might be set in the kwargs_likelihood and entered into the fitting. However, that requires more changes among various files, I don't think I can carry out that easy now..

Another thing is that the default of 'likelihood_method' is 'diagonal' is chosen by myself, which I just want it to represent the likelihood method that is used by lenstronomy before my change. Maybe you can think out a better name than 'diagonal' that summarises your likelihood method more properly.

sibirrer:
> For the noise covariance matrix, the short answer is that we have used it implicitly in our method. Because for the natural weighting (natwt for short) image, the pixel-pixel noise covariance matrix is encoded by the PSF. So in our method, we have played between unconvolved and convolved light profiles, which will extract out the noise covariance matrix and use it in the correct way.

Thank you @nanz6 !
This is not really clear to me. How are datasets with different signal-to-noise ratios distinguishable from each other? A likelihood normalization will not do that. It's important to very explicitly and specifically point out what the meaning of the different terms are when it comes to the likelihood. Also, the user might be confused when seeing the 'usual' noise terms but then these are not even being used at all.

Having these things clearly specified is probably more important than having a good design structure, as we will be able to modify the design once the documentation is there in good shape.

nanz6:
Hi @sibirrer, 
more specifically, we have C_n = (1 / sigma^2) A for the natwt interferometric images. C_n is the covariance matrix and A is the convolution operation of PSF acting on the unconvolved surface brightness image. Using this relation, the chi^2 computation is simplified as 
chi^2 = (d - A x) C_n^{-1} (d- A x) = d C_n^{-1} d + (1 / sigma^2) (x.Ax - 2 d.x) ,
where x, Ax and d are unconvolved model, convolved model and natwt data image respectively. The r.h.s is exactly how the chi^2 is computed in the ImageData log_likelihood function, where the first term is the input as logL_const and [x, Ax] is input by model and d is given by data.
This is what is happening in the image space (or pixel space). And we have input the covariance matrix information by input the PSF. 

For the interferometric PSF, it is different for different observations, and so for different data. 
In the Fourier space (which is also known as the uv-plane), an astronomic interferometry array measures the radio flux from the sky by just measuring discrete points of the uv-plane. If we denote the measured visibility as **V(u,v)W(u,v)w(u,v)**, then **V(u,v)** is the Fourier transformation of the real sky flux **V(x,y)**. **W(u,v)** is the sample function, which returns 1 for the points that are measured by the array and returns 0 for the points that are not measured. **w(u,v)** is the weighting factors, that might be adjusted to emphasize different properties of the measurement. (For example, if **w(u,v)** is set to be larger for higher frequency mode in the Fouier space, then the synthesized image will display more small scale details, although usually the high frequency information is much more uncertain than the low frequency information.)

Using the notations given above, the convolved image is the anti-Fourier transformation of **V(u,v)W(u,v)w(u,v)**, while the PSF is the anti-Fourier transformation of **W(u,v)w(u,v)**. This shows that the image and PSF are both subject to the sampling W(u,v) and weighting w(u,v), and explains why we always called the images we are working on with 'natural weighting'. The 'natural weighting' is to set w(u,v) to be the inverse of the variance of each visibility point, and will result in the simple relation of noise covariance matrix and convolution operation matrix in the image space and simplifies the chi^2 computation, as discussed in the first paragraph of this comment. It also explains why there are negative pixels of the interferometric PSF. In general, the idea of interferometry is quite different than the optical telescopes.

sibirrer:
Hi @nanz6 , thank you very much for the detailed elaboration! Perhaps putting in a raise statement that the 'standard' noise properties are/must not be used might be helpful, and perhaps some of these equations actually directly in the code might also be useful (with ..math:). Anyways, I am not sure if I am holding you back or not, please ping me when you want to review something and move forward! Thank you very much!

PR #407, opened 2023-01-31
fixed_lens_model keyword argument in likelihood evaluations 

dangilman:
Hi @sibirrer, this is a first go at adding a "fixed_lens_model" option when calling the likelihood function in FittingSequence. The changes are in the Param class and in the Image2SourceMapping class, where the new keyword argument stops lenstronomy from iterating through the array of lens model keyword arguments during each function call, and only ray traces to the source plane once during the first likelihood evaluation, respectively. This leads to about a factor of 2 improvement in speed when fitting source and lens light models to imaging data with a fixed lens model. I was, however, expecting more of an increase in speed


sibirrer (changes requested):
Hi Daniel, thanks for the PR. I have some suggestions on the design, not sure they are all doable or not. About the speed difference: I would be curious to see a profiling run to see where the most computational time is spent for your problem and how much of it is the lensing calculation, and what is other things, before further improving a specific aspect.

lenstronomy/ImSim/image2source_mapping.py
```diff
@@ -105,6 +109,22 @@ def image2source(self, x, y, kwargs_lens, index_source):
                 y_source = y_comov / T_z
         return x_source, y_source
 
+    def _image2source_mapping(self, x, y, kwargs_lens):
```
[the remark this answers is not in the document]
dangilman: yes I can implement something like this, although I thought the (x, y) coordinates when passed through the likelihood calls would always be the same when fitting imaging data since you're raytracing through a pre-determined set of pixels (unless I don't understand something).

dangilman:
Hi, I'll work on implementing your comments. I profiled the code and the most time is spent calling functions in the TNFW class, which means there are still ray shooting computations being performed somewhere. Do you know where that might be besides the image2source class?

sibirrer:
> Hi, I'll work on implementing your comments. I profiled the code and the most time is spent calling functions in the TNFW class, which means there are still ray shooting computations being performed somewhere. Do you know where that might be besides the image2source class?

Perhaps when calculating the positions of objects when defined in the image plane? Or solving the lens equation? I don't know all the options you are choosing but there are certainly some aspects that could require evaluations of the lens model prior to ray-tracing on the entire grid.

lenstronomy/ImSim/image2source_mapping.py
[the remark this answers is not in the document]
sibirrer: It is most likely always the case in the current way how the code works, but the function as it is coded is not restricted to that case and hence could be broken moving forward.

dangilman:
Hi @sibirrer, I addressed some of your comments. I don't really understand the cause of the test failure in the latest pull request. It seems to crash only on the dynesty part of test_FittingSequence

lenstronomy/ImSim/image2source_mapping.py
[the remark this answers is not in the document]
dangilman: @sibirrer I will implement your suggestion here in the next pull request

lenstronomy/Util/class_creator.py
```diff
@@ -156,7 +156,8 @@ def create_class_instances(lens_model_list=None, z_lens=None, z_source=None, z_s
 
 
 @export
-def create_image_model(kwargs_data, kwargs_psf, kwargs_numerics, kwargs_model, image_likelihood_mask=None):
+def create_image_model(kwargs_data, kwargs_psf, kwargs_numerics, kwargs_model, image_likelihood_mask=None,
+                       fixed_lens_model=False):
```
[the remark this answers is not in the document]
dangilman: done

dangilman:
> 

Yes, I tracked down all calls to ray-shooting and now the fixed_lens_model keyword argument will make sure they only happen once.

sibirrer:
> Hi @sibirrer, I addressed some of your comments. I don't really understand the cause of the test failure in the latest pull request. It seems to crash only on the dynesty part of test_FittingSequence

Hi @dangilman , there was a new dynesty release and apparently the test settings triggered an error message. I could fix it and merged the changes. If you pull the main branch, your test hopefully pass

dangilman:
@sibirrer I realized that the functionality I wanted when introducing the fixed_lens_model keyword argument could be achieved with the existing NumericalAlpha class in lenstronomy. The only change I needed to make was to add a keyword argument to kwargs_model that passes a user-defined deflection angle class to lenstronomy for sampling

dangilman:
> > @sibirrer I realized that the functionality I wanted when introducing the fixed_lens_model keyword argument could be achieved with the existing NumericalAlpha class in lenstronomy. The only change I needed to make was to add a keyword argument to kwargs_model that passes a user-defined deflection angle class to lenstronomy for sampling.
> > I also updated the name and documentation of the custom class to hopefully be more informative
> 
> Oh, that's great! Ping me once you want me to review the updated PR. Thanks a lot @dangilman !

I think it's ready, the test that failed doesn't seem to be related to the changes I made...

lenstronomy/LensModel/QuadOptimizer/param_manager.py
```diff
@@ -99,16 +99,15 @@ def bounds(self, re_optimize, scale=1.):
             thetaE_shift = 0.005
             center_shift = 0.01
             e_shift = 0.05
-            g_shift = 0.025
+            g_shift = 0.005
```
sibirrer, line 102: does this have to do why the test is failing? Tests are failing in this file and there were done modifications.

lenstronomy/LensModel/lens_param.py
```diff
@@ -25,6 +25,8 @@ def __init__(self, lens_model_list, kwargs_fixed,
          (only relevant in this class when 'SHAPELETS' is the solver type)
         :param num_shapelet_lens: integer, number of shapelets in the lensing potential
          (only relevant when 'SHAPELET' lens model is used)
+        :param fixed_lens_model: keeps the lens model fixed during likelihood calls; this setting should only be set to
+        true if all lens components are fixed
```
sibirrer, line 29: can you add an additional space at the beginning of this line break (see format of other :param doc strings above). It makes a difference in the sphinx documentation.

lenstronomy/PointSource/Types/base_ps.py
```diff
@@ -8,7 +8,7 @@ class PSBase(object):
     """
     base point source type class
     """
-    def __init__(self, lens_model=None, fixed_magnification=False, additional_image=False):
+    def __init__(self, lens_model=None, fixed_magnification=False, additional_image=False, fixed_lens_model=False):
```
sibirrer, line 11: can you add a doc string for fixed_lens_model?

lenstronomy/Util/class_creator.py
```diff
@@ -65,6 +66,8 @@ def create_class_instances(lens_model_list=None, z_lens=None, z_source=None, z_s
     :param sersic_major_axis: boolean or None, if True, uses the semi-major axis as the definition of the Sersic
      half-light radius, if False, uses the product average of semi-major and semi-minor axis. If None, uses the
      convention in the lenstronomy yaml setting (which by default is =False)
+    :param fixed_lens_model: keeps the lens model fixed during likelihood calls; this setting should only be set to
```
sibirrer, line 69: can you also add a doc string for tabulated_deflection_angles?

sibirrer (changes requested):
Looks good! I have a few minor documentation requests and the tests fail in a file where yo made edits, so I expect the tests to pass before merging. Thank you!

lenstronomy/LensModel/QuadOptimizer/param_manager.py
dangilman: yep

sibirrer (approved):

PR #409, opened 2023-01-31
Gnfw cse

mattgomer:
Created a new GNFW_ELLIPSE_CSE model which approximates the general_nfw model using a linear sum of many cored steep ellipsoids. Similar to NFW_ELLIPSE_CSE, this enables one to add ellipticity to an NFW profile, but unlike NFW_ELLIPSE_CSE, the coefficients for the sum are not imported, but fit for by inverting a matrix. The user specifies the number of components to use and the radial range over which the approximation is desired, and the output is the same as the GNFW, but with the capability for ellipticity. The accuracy of the approximation varies, but can be estimated by the `estimate_deflection_error` function.


lenstronomy/LensModel/Profiles/gnfw_ellipse_cse.py
```diff
@@ -0,0 +1,240 @@
...
+        x__, y__ = util.rotate(x_, y_, phi_q)
+
+        # potential calculation
+        a_list = self.linear_fit(gamma_inner, gamma_outer, num_components, lower_log_bound, upper_log_bound)
+        f_ = self.cse_major_axis_set.function(x__/Rs, y__/Rs, a_list, self.s_list, q)
+        const = self._normalization(alpha_Rs, Rs, gamma_inner, gamma_outer)
+        return const * f_
+
+    def derivatives(self, x, y, Rs, alpha_Rs, e1, e2, gamma_inner, gamma_outer, center_x=0, center_y=0,
+                    num_components=None, lower_log_bound=None, upper_log_bound=None):
```
sibirrer, line 149: can the numerical parameters/settings go in the __init__() statement of the class instead? Then it's easier to do parameter inference with fixed numerical settings

sibirrer (changes requested):
This is great! Perhaps even worth writing it up in a paper or so, if it gets a more general scope.
What I don't know is the computational cost of inverting the matrixes for the CSE solution. This might become relevant for e.g. solving the lens equation where a lot of single executions are happening and hence in the current implementation this inversion is going to happen a lot. For these cases, we can use the functions set_static() (see the base class and some other lens models for examples) such that you can cache these values and only release it when you know that a different parameter input is used.

lenstronomy/LensModel/Profiles/gnfw_ellipse_cse.py
```diff
@@ -0,0 +1,240 @@
...
+
+__all__ = ['GNFW_ELLIPSE_CSE']
+
+
+class GNFW_ELLIPSE_CSE(GNFW):
+    """
+    This class is used to approximate a generalized NFW profile (see LensModel.Profiles.general_nfw) using
+    a sum of CSE profiles (same method as lenstronomy.LensModel.Profiles.nfw_ellipse_cse).
+    This enables an elliptical version of a GNFW profile.
+
```
sibirrer, line 19: can you add a latex version of this profile with
..math::
   \\kappa(x, y) = ...

mattgomer:
Regarding caching the values, I'm not super familiar with this trick, so I'll have to look into it a bit. But one potential snag I've been trying to work around is that the linear inversion is different depending on the kwargs values (particularly the inner and outer slopes), so it would have to be done at each iteration in a fit, Fortunately the matrix inversion is pretty fast, but I don't know if it will be fast enough.

lenstronomy/LensModel/Profiles/gnfw_ellipse_cse.py
mattgomer: I initially had them in init(), but moved them here for 2 reasons: 
First, I wasn't sure how to get LensModel(lens_model_list) to input arguments into the init function in case the user wanted to change num_components, for example. This problem is probably fixable with a little more knowledge.
Second, I was imagining that if someone does a fit and gives a particular result, they might calculate that their error there is not up to standards and decide to restart the fit with more components. If the numerical parameters are in the init instead of the kwargs, this requires re-initializing the class. Maybe that's okay. I'm not sure how often this would happen.

On the other had, and the original reason I thought to put them in init, I don't think we want people optimizing over num_components (or worse, the bounds of evaluation) in a PSO or otherwise, so maybe they don't belong in kwargs_lens.

sibirrer:
> Regarding caching the values, I'm not super familiar with this trick, so I'll have to look into it a bit. But one potential snag I've been trying to work around is that the linear inversion is different depending on the kwargs values (particularly the inner and outer slopes), so it would have to be done at each iteration in a fit, Fortunately the matrix inversion is pretty fast, but I don't know if it will be fast enough.

You can, for example, look at the NIE class here: https://github.com/lenstronomy/lenstronomy/blob/e686674787c723c87371defb35888ac0b7ac2805/lenstronomy/LensModel/Profiles/nie.py#L166

set_static() and set_dynamic() are used to cache pre-computed values applicable for a set of input parameters but not for a set of coordinates that those are evaluated.

lenstronomy/LensModel/Profiles/gnfw_ellipse_cse.py
sibirrer: in the profiel_list_base class, you can add/define other input parameters. It's a bit hard to generalize them but there are few other classes that were using (optional) additional parameters in their initialization.

mattgomer:
I created this synthesis model, which is a more general class than the gnfw_ellipse_cse class from before. I'm thinking of removing the gnfw_ellipse_cse class, since it's now redundant. Or I could make it simply call an instance of the synthesis class. I thought I'd at least push what I have before making this decision.

lenstronomy/LensModel/Profiles/synthesis.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+
+        self.target_class = LensModel([target_lens_model])
+        self.component_class = LensModel([component_lens_model])
+        self.kwargs_list=kwargs_list
+        self.lin_fit_hyperparams=lin_fit_hyperparams
+        self.check_num_evals()
+
+
+    def LinearWeightMLEFit(self, kwargs_target, kwargs_list):
```
sibirrer, line 42: definitions are linear_weight_mle_fit() syntax style

sibirrer (commented):
Hi Matt, this looks great.
I won't have a chance to go in detail through all of it today (not before Monday). I'd like to keep a dedicated NFW and Hernquist CSE profile such that it's easier for the user to access it as well as the linear fitting is not needed as the parameters are pre-computed.

Thanks a lot!

sibirrer:
@mattgomer see also that tests failed. Ping me when you want to have the next review.

mattgomer:
@sibirrer I got a chance to take a look at this again. I implemented the set_static method like you suggested from the NIE example. Does this look right to you?

I also removed the gnfw_ellipse_cse profile I created since it can be fashioned using synthesis, but I agree that the NFW_ELLIPSE_CSE and HERNQUIST_CSE models should stay as they are: the pre-computed lists are more accurate than my linear fits.

sibirrer:
Looks great @mattgomer ! There is, however, a conflict with the main branch. Can you resolve it on your side when merging the main branch into your branch? Thanks

mattgomer:
Sorry about that @sibirrer. Should be resolved now!

sibirrer (approved):
Looks great, thank you very much!

PR #418, opened 2023-02-17
plotting bug fix

astrobenji:
Use default v_min, v_max if None is supplied. Necessary to prevent a plotting bug.


sibirrer (approved):
Looks good, note that the default bounds are currently meant for log10() scalings and not linear scaling

PR #422, opened 2023-03-01
add JWST simulations configuration

martin-millon:


sibirrer (approved):
Thank you very much @martin-millon !
This is great! More bands can be added as needed.

PR #423, opened 2023-03-03
Psf error map change in definition

sibirrer:
The previous psf_error_map that was required as input into the PSF() class was defined such that the excess variance of the error is given by sigma2_psf(i,j) = psf_error_map(i,j) * (<point source amplitude> * psf_kernel(i,j)^2). This is not an intuitive convention and in the construction of the input value, might lead to ill defined values for numbers where the psf_kernel is effectively zero.

This PR changes the convention to
sigma2_psf(i,j) = psf_error_map(i,j) * <point source amplitude>^2.

The changes make the results NOT BACKWARDS COMPATIBLE. Any psf_error_map previously being used needs to be multiplied by the kernel_point_source^2.

I am happy to first make a minor pip release of lenstronomy before these changes come into effect.


sibirrer:
oh, I made the merge in an existing branch (which is not used anymore). So I decide to merge and make the review for the merge with the main branch

PR #432, opened 2023-03-28
Fix the model plot linear solver for interferometric likelihood method

nanz6:
Fix the model plot linear solver for interferometric likelihood method.

In the SingleBandMultiModel.image_linear_solve, the wls_model should be the convolved model image, where the returned wls_model from the interferometric _image_linear_solve is [unconvolved, convolved] images. Here wls_model just needs to be assigned to get the second array of _image_linear_solve model output. By doing this, no errors would be raised.

Add a part in test_model_plot to ensure that the error will NOT be raised for the interferometry model plot linear solver, using a try, except, else structure.

### DIFF
--- a/test/test_Plots/test_model_plot.py
+++ b/test/test_Plots/test_model_plot.py
@@ -296,6 +296,23 @@ class TestRaise(unittest.TestCase):
                                                   [kwargs_data, {'psf_type': 'NONE'}, {}]],
                                  kwargs_model=kwargs_model, kwargs_params=kwargs_params, bands_compute=[True],
                                  arrow_size=0.02, cmap_string="gist_heat", linear_solver=False)
+        
+        # test no error would be raised if kwargs_data['likelihood_method'] = 'interferometry_natwt' for Model Plot
+        with self.assertRaises(ValueError):
+            try:
+                kwargs_data = sim_util.data_configure_simple(numPix=10, deltaPix=1, background_rms=1, exposure_time=1)
+                kwargs_data['likelihood_method'] = 'interferometry_natwt'
+                kwargs_model = {'source_light_model_list': ['GAUSSIAN']}
+                kwargs_params = {'kwargs_lens': [],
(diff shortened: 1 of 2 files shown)


test/test_Plots/test_model_plot.py
```diff
@@ -296,6 +296,23 @@ def test_raise(self):
                                                   [kwargs_data, {'psf_type': 'NONE'}, {}]],
                                  kwargs_model=kwargs_model, kwargs_params=kwargs_params, bands_compute=[True],
                                  arrow_size=0.02, cmap_string="gist_heat", linear_solver=False)
+        
+        # test no error would be raised if kwargs_data['likelihood_method'] = 'interferometry_natwt' for Model Plot
+        with self.assertRaises(ValueError):
+            try:
+                kwargs_data = sim_util.data_configure_simple(numPix=10, deltaPix=1, background_rms=1, exposure_time=1)
```
sibirrer, line 303: I don't really understand this test here. Do you expect this function to raise? The test function is expecting a VaueError raise, but then you have a try/except with an else statement that by definition gives you a ValueError.

sibirrer (changes requested):
Sounds great, and glad it did not take much to make it happen. I am not sure the test functions are properly made. I don't expect you to have a raise test for this part as it should run through.

test/test_Plots/test_model_plot.py
nanz6: I don't like this part neither, so I can remove it.

The idea is to check the part executed by try will not raise errors. 
If the try part raises errors, then it would excute the "except: pass", which will not finally raise an error, then assertRaises will report that the test is not passed.
If the try part does not raise errors, it would raise error in the "else: raise..." part, which would pass the assertRaises test.

sibirrer: could you simply remove the raiseTest and have this part of the test in a regular test_ function? Would be nice checking that it works with the interferometric data, such that further changes do consider this case

sibirrer (approved):
Excellent! Thank you very much!

PR #436, opened 2023-04-03
Fix copying dictionary elements

ajshajib:
This PR fixes a minor bug, where the provided `kwargs_params` would not have the linear parameter values filled in after creating a `ModelPlot` instance with `linear_solver=True`.


sibirrer (approved):
Thank you very much! Was not aware that the key elements were not copied over.

PR #455, opened 2023-04-30
Feature/update z sampling

ajshajib:
This PR adds to these functionalities:

- It sets the upper and lower limits for the PSO parameter space not to overshoot the user-defined limits.
- It updates the redshift list(s) in the `kwargs_model` after PSO if the redshift is sampled.

### DIFF
--- a/lenstronomy/Workflow/fitting_sequence.py
+++ b/lenstronomy/Workflow/fitting_sequence.py
@@ -97,13 +97,11 @@ class FittingSequence(object):
             elif fitting_type == 'PSO':
                 kwargs_result, chain, param = self.pso(**kwargs)
                 self._updateManager.update_param_state(**kwargs_result)
-                self._updateManager.update_kwargs_model(kwargs_result)
                 chain_list.append([fitting_type, chain, param])
 
             elif fitting_type == 'SIMPLEX':
                 kwargs_result = self.simplex(**kwargs)
                 self._updateManager.update_param_state(**kwargs_result)
-                self._updateManager.update_kwargs_model(kwargs_result)
                 chain_list.append([fitting_type, kwargs_result])
 
             elif fitting_type == 'MCMC':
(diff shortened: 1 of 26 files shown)


lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -124,6 +127,7 @@ def fit_sequence(self, fitting_list):
                 if not kwargs.get('one_step', False):  # this is only for testing purposes
                     kwargs_result = self.best_fit_from_samples(points, log_l)
                     self._updateManager.update_param_state(**kwargs_result)
+                    self._updateManager.update_kwargs_model(kwargs_result)
```
sibirrer, line 130: do we need this extra function or can it be part of update_param_state() instead? Just asking as it would ease a bit the interface and making sure it is usable for all the fitters.

sibirrer (changes requested):
Thanks @ajshajib ! Looks good. I have a question of a potential change of merging the function with another one. Also, I noticed that all the tests are recently crashing due to coveralls (haven't figured out what's going on there).
Can you add a small test for the definition that you wrote as well?
Thank you!

lenstronomy/Workflow/fitting_sequence.py
ajshajib: That's a good point! I have consolidated these function calls as you suggested.

ajshajib:
I have made the requested changes and added a test that passes. I see that the coveralls is throwing an error because it's unable to submit a coverage report, probably an issue in linking with the coveralls server. 🤷‍♂️

sibirrer (approved):

sibirrer:
> I have made the requested changes and added a test that passes. I see that the coveralls is throwing an error because it's unable to submit a coverage report, probably an issue in linking with the coveralls server. 🤷‍♂️

yeah, super weird. I see it happen on all my repos with coveralls links...perhaps they fix it soon. I can re-run the tests then. Just don't want to merge and cause a failed CI at the moment

ajshajib:
Yes, no worries. Merging can wait until coveralls is fixed. 🙂

PR #464, opened 2023-05-01
Enforce parameter limits in SpecialParam class

ajshajib:
This PR:
- reverts enforcing initial parameter limits within the `FittingSequence` class, since it is already being done by the `Sampler.pso()` method.
- directly imposes the user-defined bound through `SpecialParam.get_params()` to avoid error thrown when a lens-plane redshift is sampled higher than the reference source-plane redshift in the multi-plane mode, as parameters outside the user-defined bounds can be sampled by `ParticleSwarmOptimizer`.

### DIFF
--- a/lenstronomy/Workflow/fitting_sequence.py
+++ b/lenstronomy/Workflow/fitting_sequence.py
@@ -309,11 +309,6 @@ class FittingSequence(object):
         lower_start = np.array(init_pos) - np.array(sigma_start) * sigma_scale
         upper_start = np.array(init_pos) + np.array(sigma_start) * sigma_scale
 
-        lower_limit, upper_limit = param_class.param_limits()
-
-        lower_start = np.maximum(lower_start, lower_limit)
-        upper_start = np.minimum(upper_start, upper_limit)
-
         num_param, param_list = param_class.num_param()
         # run PSO
         sampler = Sampler(likelihoodModule=self.likelihoodModule)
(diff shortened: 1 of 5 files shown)


lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -309,11 +309,6 @@ def pso(self, n_particles, n_iterations, sigma_scale=1, print_key='PSO', threadC
         lower_start = np.array(init_pos) - np.array(sigma_start) * sigma_scale
         upper_start = np.array(init_pos) + np.array(sigma_start) * sigma_scale
 
-        lower_limit, upper_limit = param_class.param_limits()
```
sibirrer, line 312: I am not sure this change does the same. I understand that the kwargs_special parameters get within the bounds, but for all the rest this initial ball of parameters could still be out of bounds

sibirrer (changes requested):
I see that you want to assure that all the kwargs_special (in particular the redshifts) are always in the bounds. The changes, it looks to me, would change it for all the other parameters since the initial guess in the ball of particles would not behave this way.

lenstronomy/Workflow/fitting_sequence.py
ajshajib: These limits are already being set by `Sampler.pso()`, so I removed these lines to avoid redundancy.

sibirrer: Ah, I see. Thanks

sibirrer:
I see. So if you add a test that runs through all the new lines, that would be great. And then we are set to merge

sibirrer (approved):
Looks good. Thank you!

PR #471, opened 2023-05-18
minor bug fix for multi-frame plotting

sibirrer:


sibirrer:
fixes a minor bug and avoids plotting point sources outside the frame for convenience

PR #475, opened 2023-06-01
Cobaya in lenstronomy

nataliehogg:
Implementation of the pure Metropolis--Hastings MCMC sampler in Cobaya for use with lenstronomy as an alternative to the ensemble MCMC methods already available (emcee, zeus).

PR #479, opened 2023-06-20
Change sampler naming conventions

### LINKED ISSUE #477: Changing sampling algorithm naming conventions in FittingSequence
In the current version of FittingSequence, some sampling algorithms are accessed via a "top level" function named after the general algorithm i.e. mcmc() which contains the ensemble samplers emcee and zeus, and nested_sampling(), which contains the nested samplers dynesty, multinest and polychord. Apart from that, the Nautilus and Cobaya samplers are accessed separately by directly calling their classes from Sampling.Samplers. 

This process is a bit convoluted and could be streamlined in various ways:
1. Have all samplers accessed via a top-level function named after the specific algorithm i.e. metropolis_hastings(), ensemble_MCMC(), nested_sampling(), importance_nested_sampling(). The user would first pass a kwarg calling the sampling method, and in the cases where there is more than one option (ensemble_MCMC and nested_sampling), a kwarg for the specific software.
2. Have all samplers directly accessed via the software names i.e. by passing kwargs such as 'emcee', 'zeus', 'dynesty' etc directly to the fitting sequence.

Issue: either of these options will introduce backwards incompatibility as it will likely require changing the 'MCMC' kwarg most commonly passed to the fitting sequence.

nataliehogg:
This PR resolves #477 by renaming the various samplers available in lenstronomy so that the software names are used instead of a mixture of software names and sampling algorithms (as also discussed in #475). The software names are typeset according to the format chosen by the authors of the original packages.

Backwards compatibility is maintained with the 'MCMC' keyword argument defaulting to emcee. Backwards compatibility with 'nested_sampling' is not maintained; anyone using this option will now have to select 'dynesty', 'dyPolyChord' or 'MultiNest'. The sampler_type keyword argument in mcmc() and nested_sampling() is deprecated; if a user passes this keyword argument, a warning is printed and the option ignored.


lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -73,41 +73,48 @@ def fit_sequence(self, fitting_list):
         """
         chain_list = []
         for i, fitting in enumerate(fitting_list):
-            fitting_type = fitting[0]
+            self.fitting_type = fitting[0]
```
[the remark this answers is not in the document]
ajshajib: I think keeping the `sampler_type` variable to pass the sampler name to the corresponding function(s) instead of `self.fitting_type` may be a more explicit way to do it. The sampling functions are intended to be public, but if they depend on `self.fitting_type`, which may cause issues for a future user/extension as this is not documented. The fix would be to set `self.fitting_type` before calling the function, but sticking with a direct argument for `sampler_type` for MCMC routines would keep things straightforward. But am I missing some other issue if the argument is passed directly?

ajshajib (changes requested):

lenstronomy/Workflow/fitting_sequence.py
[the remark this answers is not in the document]
nataliehogg: My goal was to make it so that the functions of `fitting_type` and `sampler_type` are merged into one. This would mean that the user would not need to specify e.g. both 'nested_sampling' and 'dynesty' in the list that's passed to FittingSequence(), but rather just the software name. But if the two should remain separate I can revert those changes. I'm not sure I understand your suggestion just yet -- I'll play around with it.

lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -246,8 +247,10 @@ def simplex(self, n_iterations, method='Nelder-Mead'):
         kwargs_result = param_class.args2kwargs(result, bijective=True)
         return kwargs_result
 
-    def mcmc(self, n_burn, n_run, walkerRatio=None, n_walkers=None, sigma_scale=1, threadCount=1, init_samples=None,
-             re_use_samples=True, sampler_type='EMCEE', progress=True, backend_filename=None, start_from_backend=False,
+    def mcmc(self,
+             n_burn, n_run, walkerRatio=None, n_walkers=None, sigma_scale=1, threadCount=1, init_samples=None, re_use_samples=True,
+             sampler_type='EMCEE', # NH: eventually this can be removed
+             progress=True, backend_filename=None, start_from_backend=False,
              **kwargs_zeus):
```
ajshajib, line 254:
```suggestion
    def mcmc(self,
             n_burn, n_run, walkerRatio=None, n_walkers=None, sigma_scale=1, threadCount=1, init_samples=None, re_use_samples=True,
             sampler_type='EMCEE',
             progress=True, backend_filename=None, start_from_backend=False,
             **kwargs_zeus):
```

lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -73,41 +73,48 @@ def fit_sequence(self, fitting_list):
...
+            # rather than in all-caps. this will result in some backwards incompatibility
+
+            # furthermore the samplers are sorted by type and then alphabetically
+
+            elif self.fitting_type == 'MCMC' or self.fitting_type == 'emcee' or self.fitting_type == 'zeus':
+                if self.fitting_type == 'MCMC':
+                    print('MCMC selected. Sampling with default option (emcee).')
                 if 'init_samples' not in kwargs:
                     kwargs['init_samples'] = self._mcmc_init_samples
                 elif kwargs['init_samples'] is None:
```
ajshajib, line 120: Below here, I suggest to pass `sampler_type` explicitly to `mcmc()` as:

```
mcmc_output = self.mcmc(**kwargs, sampler_type=fitting_type)
```

lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -297,22 +307,21 @@ def mcmc(self, n_burn, n_run, walkerRatio=None, n_walkers=None, sigma_scale=1, t
...
-        if sampler_type == 'EMCEE':
-            samples, dist = mcmc_class.mcmc_emcee(n_walkers, n_run, n_burn, mean_start, sigma_start, mpi=self._mpi,
-                                                  threadCount=threadCount, progress=progress, initpos=initpos,
-                                                  backend_filename=backend_filename,
-                                                  start_from_backend=start_from_backend)
-            output = [sampler_type, samples, param_list, dist]
-
-        elif sampler_type == 'ZEUS':
-
+        if self.fitting_type == 'zeus':
```
ajshajib, line 310:
```suggestion
        if sampler_type == 'zeus':
```

lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -297,22 +307,21 @@ def mcmc(self, n_burn, n_run, walkerRatio=None, n_walkers=None, sigma_scale=1, t
...
+            # check if zeus is specified, if not default to emcee
             samples, dist = mcmc_class.mcmc_zeus(n_walkers, n_run, n_burn, mean_start, sigma_start,
                                                  mpi=self._mpi, threadCount=threadCount,
                                                  progress=progress, initpos = initpos, backend_filename = backend_filename,
                                                  **kwargs_zeus)
-            output = [sampler_type, samples, param_list, dist]
+            output = [self.fitting_type, samples, param_list, dist]
```
ajshajib, line 316:
```suggestion
            output = [sampler_type, samples, param_list, dist]
```

lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -297,22 +307,21 @@ def mcmc(self, n_burn, n_run, walkerRatio=None, n_walkers=None, sigma_scale=1, t
...
         else:
-            raise ValueError('sampler_type %s not supported!' % sampler_type)
+            # sample with emcee
+            samples, dist = mcmc_class.mcmc_emcee(n_walkers, n_run, n_burn, mean_start, sigma_start, mpi=self._mpi,
+                                                  threadCount=threadCount, progress=progress, initpos=initpos,
+                                                  backend_filename=backend_filename,
+                                                  start_from_backend=start_from_backend)
+            output = [self.fitting_type, samples, param_list, dist]
```
ajshajib, line 323:
```suggestion
            output = [sampler_type, samples, param_list, dist]
```

lenstronomy/Workflow/fitting_sequence.py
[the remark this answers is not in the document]
ajshajib: I think `sampler_type` does not need to be specified by the user, but this variable can still be used internally to pass the sampler type to the associated methods. In that way, `mcmc()` and `nested_sampling()` can still be used as a "public" method by a user without having to explicitly set `self.fitting_type` before calling them (here I am thinking about a general use case where a user may be using the public methods directly for any purpose outside the usual use case of calling `FittingSequence.fit_sequence()`). I have left some direct code suggestions for the example case of `mcmc()` for how I mean this to be implemented, but similar usage of the `sampler_type` variable also applies to `nested_sampling()` method. But please let me know if something is still unclear. 🙂

lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -112,72 +112,76 @@ def fit_sequence(self, fitting_list):
...
 
             elif fitting_type == "SIMPLEX":
                 kwargs_result = self.simplex(**kwargs)
                 self._updateManager.update_param_state(**kwargs_result)
-                chain_list.append([fitting_type, kwargs_result])
+                chain_list.append([self.fitting_type, kwargs_result])
 
-            elif fitting_type == "MCMC":
+            elif fitting_type in ["MCMC", "emcee", "zeus"]:
+                # == 'MCMC' or fitting_type == 'emcee' or fitting_type == 'zeus'):
```
ajshajib, line 123:
```suggestion
```

lenstronomy/Workflow/fitting_sequence.py
```diff
@@ -112,72 +112,75 @@ def fit_sequence(self, fitting_list):
...
                 mh_output = [updated_info, sampler_type, best_fit_kwargs]
-
                 # append the products to the chain list
                 chain_list.append(mh_output)
 
+            elif (
+                fitting_type == "dynesty"
+                or fitting_type == "dyPolyChord"
+                or fitting_type == "MultiNest"
+            ):
```
ajshajib, line 159:
```suggestion
            elif fitting_type in ["dynesty", "dyPolyChord", "MultiNest"]:
```

ajshajib (approved):

ajshajib:
Thanks for the PR, @nataliehogg! I went ahead and resolved the conversations with new commits. 

@sibirrer, I approved the PR. Please feel free to merge when appropriate. This PR changes the expected `"fitting_type"` keywords for the sampling, so it won't be backward compatible for some cases.

nataliehogg:
thank you @ajshajib !

sibirrer:
Thank you very much @nataliehogg @ajshajib ! This is a great improvement and adds clarity to the samplers and design

PR #485, opened 2023-07-12
Add substructure plot 

dangilman:
I made a new routine in model_band_plot that plots the effective multiplane convergence in substructure. I also made some minor fixes to other classes, including adding the "observed_convention_index" option to model API and removing a redundant line of code


lenstronomy/Plots/model_band_plot.py
```diff
@@ -192,6 +193,84 @@ def convergence_plot(self, ax, text='Convergence', v_min=None, v_max=None,
...
 
+    def substructure_plot(self, ax, index_macromodel, text='Substructure convergence', subtract_mean=True,
+                          v_min=-0.05, v_max=0.05, font_size=15, colorbar_label=r'$\kappa - \kappa_{\rm{macro}}$',
+                          cmap='bwr',  with_critical_curves=False, crit_curve_color='k', image_name_list=None,
+                          **kwargs):
+        """
+
+        Plots the convergence of a full lens model minus the convergence from a few specified lens models to
+        more clearly show the presence of substructure
+        :param ax: matplotib axis instance
```
sibirrer, line 204: can you add a empty line here. For some reasons, readthedocs otherwise messes up the html

lenstronomy/Plots/plot_util.py
```diff
@@ -157,7 +157,6 @@ def plot_line_set(ax, coords, line_set_list_x, line_set_list_y, origin=None, fli
         ax.plot((x_c + shift) * pixel_width_x + origin[0], (y_c + shift) * pixel_width + origin[1], *args, **kwargs)
     return ax
 
-
```
sibirrer, line 160: conventions for definitions outside of classes is two empty lines

lenstronomy/SimulationAPI/model_api.py
```diff
@@ -40,7 +41,9 @@ def __init__(self, lens_model_list=None, z_lens=None, z_source=None, lens_redshi
         :param z_source_convention: float, redshift of a source to define the reduced deflection angles of the lens
          models. If None, 'z_source' is used.
         :param tabulated_deflection_angles: a class that returns deflection angles given a set of (x, y) coordinates.
-         Effectively a fixed lens model. See documentation in Profiles.numerical_alpha
+        Effectively a fixed lens model. See documentation in Profiles.numerical_alpha
```
sibirrer, line 44: sphinx messes up when you do not include one whitespace in new lines with the same parameter

lenstronomy/SimulationAPI/model_api.py
```diff
@@ -40,7 +41,9 @@ def __init__(self, lens_model_list=None, z_lens=None, z_source=None, lens_redshi
...
+        :param observed_convention_index: a list of indicies that correspond to lens models where the center_x,center_y
+        values correspond to the observed (lensed positions), not the physical positions in space
```
sibirrer, line 46: single whitespace to be added

lenstronomy/Workflow/update_manager.py
```diff
@@ -33,8 +33,6 @@ def __init__(self, kwargs_model, kwargs_constraints, kwargs_likelihood, kwargs_p
         self.kwargs_model = kwargs_model
         self.kwargs_constraints = kwargs_constraints
         self.kwargs_likelihood = kwargs_likelihood
-        if 'fixed_lens_model' in self.kwargs_constraints.keys():
```
sibirrer, line 36: is there a reason to remove these lines? It might break someones application

sibirrer (changes requested):
Looks good to me. I have some very minor requests and a question about two removed lines of code. Thanks a lot!

lenstronomy/Workflow/update_manager.py
dangilman: I added 'fixed_lens_model' as part of an older pull request, but then realized I could get the same functionality another way. I then went back and deleted all "fixed_lens_model" keyword arguments from functions/classes, but apparently missed this one

sibirrer: thanks for the clarification

sibirrer:
@dangilman there seems to be an added line that is missed in the tests. I don't really know which one, or perhaps it is because of deleted lines?

dangilman:
> @dangilman there seems to be an added line that is missed in the tests. I don't really know which one, or perhaps it is because of deleted lines?

I think it was because of an if/else statement that checks for a multiplane lens model. I added a multiplane lens model to the tests which should hit the line

sibirrer:
Thank you @dangilman!

PR #489, opened 2023-07-20
Stellar Kinematics Neural Network (SKiNN)

mattgomer:
This update adds functionality through the SKiNN package, which emulates JAM via a neural network,  to perform joint lensing+dynamical modeling. This adds a kinematic_2D_likelihood term which can be sampled when "kinematic_2D_likelihood=True" in the likelihood. We have also added a data class to implement binned 2D kinematic data. 

SKiNN is limited to PEMD mass profiles and Sersic light profiles, and samples using q and phi, for which we have created the PEMD_Q_PHI profile and SERSIC_ELLIPSE_Q_PHI profile. The limitation to a PEMD leads to an issue when testing on github due to the fastell4py fortran compiler issue. We should discuss our options here, since SKiNN only works with a PEMD profile. Perhaps we can amend it to also work with an SIE for testing purposes?


lenstronomy/LightModel/linear_basis.py
```diff
@@ -76,7 +77,7 @@ def functions_split(self, x, y, kwargs_list, k=None):
                     kwargs_new.update(new)
                     response += self.func_list[i].function_split(x, y, **kwargs_new)
                     n += num
-                elif model in ['SHAPELETS', 'SHAPELETS_POLAR', 'SHAPELETS_POLAR_EXP',
+                elif model in ['SHAPELETS', 'SHAPELETS_POLsAR', 'SHAPELETS_POLAR_EXP',
```
sibirrer, line 80: I guess this is a typo

lenstronomy/Sampling/Likelihoods/kinematic_2D_likelihood.py
```diff
@@ -0,0 +1,165 @@
...
+
+        Vrms = np.array(numerator) / np.array(denominator).clip(0)
+
+        return Vrms
+
+    def _logL(self,vrms):
+        """
+        Calculates the log likelihood for a given binned model
+        """
+        log_like = (vrms - self.data)**2 / self.noise**2
```
sibirrer, line 158: this likelihood assumes uncorrelated noise properties, while in reality a lot of the systematics might be correlated between the data elements

lenstronomy/Sampling/Likelihoods/kinematic_NN_call.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self):
+        try:
+            import SKiNN
+            from SKiNN.generator import Generator
+            self.SKiNN_installed = True
+            self.generator = Generator()
+        except:
+            print("Warning : SKiNN not installed properly, \
+        but tests will be trivially fulfilled. \
+                  Get SKiNN from https://github.com/lucabig/lensing_odyssey_kinematics")
```
sibirrer, line 18: it looks like this package is not public (yet) or is the link wrong?

lenstronomy/Sampling/likelihood.py
```diff
@@ -100,7 +105,7 @@ def __init__(self, kwargs_data_joint, kwargs_model, param_class, image_likelihoo
         self._custom_logL_addition = custom_logL_addition
         self._kwargs_time_delay = {'time_delays_measured': time_delays_measured,
                                    'time_delays_uncertainties': time_delays_uncertainties}
-        self.kwargs_imaging = {'multi_band_list': multi_band_list, 'multi_band_type': multi_band_type,
+        self._kwargs_imaging = {'multi_band_list': multi_band_list, 'multi_band_type': multi_band_type,
```
sibirrer, line 108: you set this object to private. I am not sure this breaks some backwards compatibilities with other routines that would like to access this dictionary outside of the class

sibirrer:
Thanks for the PR @mattgomer! 
For the PEMD model, you could replace it with the EPL model, which is basically the same, but does not need fastell4py (I more recently tend to use the EPL model but left the PEMD the same way for prosperity).
Regarding the (phi, q) vs (e1, e2) parameterization: My experience is that the fitting in (e1, e2) space is easier. Isn't it possible to translate the (e1, e2) parameters into (phi, q) at the level of the kinematic likelihood instead of the lens model? Or was there a particular reason why you want to sample in q, phi space?

In the meantime I left some minor comments and will review things in more detail later.

lenstronomy/Data/kinematic_bin_2D.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def binned_image(self):
+        """
+        Creates the binned image of the kinemmatic
+        """
+        binned_image = np.zeros_like(self._bin_mask)
+        for idx, value in enumerate(self._data):
+            binned_image[self._bin_mask==idx] = value
+        return binned_image
+    def KinBin2kwargs(self):
```
sibirrer, line 55: can you make sure you follow the standard of leaving one blank line between definitions and definitions are lower case, such as kin_bin2kwargs(). Caps is for classes according to PEP8

mattgomer:
> Thanks for the PR @mattgomer! For the PEMD model, you could replace it with the EPL model, which is basically the same, but does not need fastell4py (I more recently tend to use the EPL model but left the PEMD the same way for prosperity). 

Cool! Does the EPL return the exact same as the PEMD, or could it lead to differences if I replace the PEMD with the EPL? (And if so, how big are these differences? Maybe it would be okay) The NN was trained with GLEE's implementation of the PEMD anyway, and we use Sebastian's code to convert between GLEE and lenstronomy conventions, so maybe it's not critical in the end which lenstronomy implementation is used, as long as it maps back to the correct GLEE parameters.

>Regarding the (phi, q) vs (e1, e2) parameterization: My experience is that the fitting in (e1, e2) space is easier. Isn't it possible to translate the (e1, e2) parameters into (phi, q) at the level of the kinematic likelihood instead of the lens model? Or was there a particular reason why you want to sample in q, phi space?

If I recall correctly, the reason was that the NN call requires the input to be in q ,phi and we didn't want to reinitialize a lenstronomy model at each call of the NN to convert from e1,e2, to q,phi. But as I'm thinking about it, I'm not sure this is necessary because we may be able to simply convert during the step where the NN is called without needing a lenstronomy model. We were worried about this conversion slowing down the call, but if we don't need to reinitialize a model I think the time sink will be essentially zero.

sibirrer:
> Cool! Does the EPL return the exact same as the PEMD, or could it lead to differences if I replace the PEMD with the EPL? (And if so, how big are these differences? Maybe it would be okay) The NN was trained with GLEE's implementation of the PEMD anyway, and we use Sebastian's code to convert between GLEE and lenstronomy conventions, so maybe it's not critical in the end which lenstronomy implementation is used, as long as it maps back to the correct GLEE parameters.

It gives the same result. I tested it with this test function: https://github.com/lenstronomy/lenstronomy/blob/977412209625b3508d08fed6241304e0d03582db/test/test_LensModel/test_Profiles/test_epl.py#L155

sibirrer:
> > Regarding the (phi, q) vs (e1, e2) parameterization: My experience is that the fitting in (e1, e2) space is easier. Isn't it possible to translate the (e1, e2) parameters into (phi, q) at the level of the kinematic likelihood instead of the lens model? Or was there a particular reason why you want to sample in q, phi space?
> 
> If I recall correctly, the reason was that the NN call requires the input to be in q ,phi and we didn't want to reinitialize a lenstronomy model at each call of the NN to convert from e1,e2, to q,phi. But as I'm thinking about it, I'm not sure this is necessary because we may be able to simply convert during the step where the NN is called without needing a lenstronomy model. We were worried about this conversion slowing down the call, but if we don't need to reinitialize a model I think the time sink will be essentially zero.

Yeah, maybe think about 'just' converting the (e1, e2) space before putting it into the NN likelihood. If this leads to overall better behavior, no need to go the longer route of having another set of definitions for lens models. There might in the longer run also a more generalizable way to switch between (q, phi) and (e1, e2) in the sampling, as many models are in that regard self-similar

sibirrer (changes requested):
ok, here some first minor comments. The tests are failing and let's focus first on the higher level discussion.

Please ping me again when you would like a more detailed review. Thank you!

lenstronomy/Sampling/likelihood.py
mattgomer: I changed  `self._kwargs_imaging` to `self.kwargs_imaging`, although note that this is inconsistent with the notation of `self._kwargs_time_delay` and `self._kwargs_flux`. Is there another routine somewhere that calls `LikelihoodModule.kwargs_imaging` instead of `LikelihoodModule._kwargs_imaging`? Is that the problem?

mattgomer:
@sibirrer Halfway through implementing these changes, I rediscovered why we used the Q_PHI profiles. SKiNN was trained with the q of the mass and light allowed to vary but the PAs are aligned. So if you wanted to fix that in lenstronomy it's hard because you can't fix e1 and e2 to allow the same phi. Instead we made these profiles so we can fix phi and sample q. I have an alternative, but I don't think it's better:

We could sample with e1 and e2, but make a check when the NN is called to confirm that the PAs are the same between mass and light. One problem with this is for small e the PA is ill-defined, so we might actually want to check if q is close to 1 OR the PAs are close. The bigger problem is that this would waste time sampling, because the e2 doesn't "know" that it should be close to e1 (for example) to get the right PA, so a penalty would waste every sample where they don't lie close together. This could get especially difficult if q goes close to 1, the e1 and e2 become uncoupled, and then q increases again with e1 and e2 totally uncorrelated, therefore having to waste time sampling to get them lined up again.

So I could make an EPL_Q_PHI class  and the current SERSIC_ELLIPSE_Q_PHI and sample with q and phi, or we can throw out a large fraction of samples. What do you think?

sibirrer:
> 
> So I could make an EPL_Q_PHI class and the current SERSIC_ELLIPSE_Q_PHI and sample with q and phi, or we can throw out a large fraction of samples. What do you think?


I see @mattgomer, then it's probably better to have a EPL_Q_PHI model. I am thinking of how to generalize such that in principle all models currently parameterized with (e1, e2) may also have an option to sample from (q, phi). This might take a while but could potentially save having new code. Potentially this could work with inheritance of a class that just takes q, phi and then calls the previous function with e1,e2 arguments transformed from the input. But that can be down the road

lenstronomy/Sampling/Likelihoods/kinematic_2D_likelihood.py
mattgomer: Now uses the correlation matrix

mattgomer:
@sibirrer  I've made these updates and the tests look like they're passing now. Note that the tests are under statements like 
```
self.kinematic_NN = kinematic_NN_call.kinematic_NN().SKiNN_installed
if self.kinematic_NN:
        ...
```

so they don't run if the NN is not installed, which is what happens on github. Locally, they run fine with the NN installed.

lenstronomy/Sampling/Likelihoods/kinematic_NN_call.py
```diff
@@ -0,0 +1,69 @@
...
+                               'incli': [0, 90]}
+        for idx,pname in enumerate(['q_mass','q_light','theta_E','n_sersic','R_sersic',
+                                    'core_size','gamma','b_ani','incli']):
+            if input_p[idx]<training_abs_bounds[pname][0] or input_p[idx]>training_abs_bounds[pname][1]:
+                if verbose:
+                    print('NN CALL WARNING: param', pname, 'is outside of training bounds!')
+                within_bounds=False
+        if input_p[4] > input_p[2] or input_p[4] < 0.5*input_p[2]:
+            if verbose:
+                print('NN CALL WARNING: R_sersic is not within theta_E training bounds!')
```
sibirrer, line 63: this print statement is a bit confusing

lenstronomy/Sampling/likelihood.py
```diff
@@ -117,11 +122,24 @@ def __init__(self, kwargs_data_joint, kwargs_model, param_class, image_likelihoo
...
         self._kwargs_flux = {'flux_ratios': flux_ratios, 'flux_ratio_errors': flux_ratio_errors}
         self._kwargs_flux.update(self._kwargs_flux_compute)
+        
+        if self._kinematic_2D_likelihood is True :
+            print("Note that the 2D kinematic likelihood assumes that the lens and lens light have the same center and orientation")
+            if len(multi_band_list) > 1 :
+                print('Kinematic Likelihood not meant for multiband, using first band by default')
+            if kwargs_model['lens_model_list'][kin_lens_idx] not in ['EPL_Q_PHI']:
+                print('Lens for kinematic is not EPL_Q_PHI, the 2D kinematic likelihood will break.')
+            if kwargs_model['lens_light_model_list'][kin_lens_light_idx] not in ['SERSIC_ELLIPSE_Q_PHI']:
```
sibirrer, line 132: is this still required or can you use the default SERSIC_ELLIPSE model and then convert e1, e2 into phi, q?

lenstronomy/Util/image_util.py
```diff
@@ -117,6 +117,17 @@ def rotateImage(img, angle):
     imgR = ndimage.rotate(img, angle, reshape=False)
     return imgR
 
+@export
+def shiftImage(img, shift):
+    """
+
+    querries scipy.ndimage.shift routine
```
sibirrer, line 124: check the documentation by typing 'make docs'. Empty line should be between description and parameter definitions

lenstronomy/Util/kin_sampling_util.py
```diff
@@ -0,0 +1,205 @@
...
+        :param xy_to_radec_matrix: transformation matrix to convert from pixel xy to ra/dec
+        :param ra_atxy0: ra coordinate at pixel (0,0)
+        :param dec_atxy0: dec coordinate at pixel (0,0)
+
+        :return: ra and dec coordinates
+        """
+        ra, dec = xy_to_radec_matrix.dot(np.array([x, y]))
+        return ra + ra_atxy0, dec + dec_atxy0
+
+    def rotateHST_into_kinNN(self, HST_x, HST_y, ellipse_PA_to_hstx_angle,
```
sibirrer, line 113: I am not sure why this is needed. All the individual data can have completely different alignments and coordinate zero points and no particular parameterization is aligned with the HST or any other imaging data

test/test_LensModel/test_Profiles/test_epl.py
```diff
@@ -193,6 +193,38 @@ def test_epl_pemd_convention(self):
                             npt.assert_almost_equal(f_x, f_x_pemd, decimal=4)
                             npt.assert_almost_equal(f_y, f_y_pemd, decimal=4)
 
+    class TestEPLvsEPL_QPHI(object):
```
sibirrer, line 196: can you add the EPL_Q_PHI model to do the numerical differential testings? There is a separate test file and you can add a simple definition, just to check whether the differentials match between potential, deflection, and hessian

test/test_Sampling/test_Likelihoods/test_kinematic_2D_likelihood.py
```diff
@@ -0,0 +1,175 @@
...
+            for key, val in self.kwargs_lens_light[0].items():
+                kwargs_lens_light_close[0][key] = val + 0.01
+                kwargs_lens_light_far[0][key] = val + 0.2
+
+            close_logL = self._KinLikelihood.logL(kwargs_lens_close, kwargs_lens_light_close, self.kwargs_special, verbose=False)
+            far_logL = self._KinLikelihood.logL(kwargs_lens_far, kwargs_lens_light_far, self.kwargs_special, verbose=False)
+            assert close_logL > -10 #instance close to truth should be likely (usually around -0.5)
+            assert close_logL > far_logL #param instance close to truth is more likely than far away
+
+            #test  that doubling variance decreases logL by factor of 2
```
sibirrer, line 99: can you also test that when you put the parameters 1-sigma off, you get a 1-sigma likelihood difference?

test/test_Util/coolest_template_update_pyAPI.json
```diff
@@ -303,14 +303,14 @@
                   "fixed": false,
                   "point_estimate": {
                     "py/object": "coolest.template.classes.parameter.PointEstimate",
-                    "value": 0.014766053302357057
+                    "value": 0.014766053302355864
```
sibirrer, line 306: why have these parameters changed?

sibirrer (changes requested):
Thank you @mattgomer !
I left quite some comments, many of them for PEP8 standards. I have not marked all of them (and did not find all for sure). I would recommend adding a plugin to check for those on your editor. There are many either missing empty lines, spacing off between operations, naming conventions off.

Then I am not sure about the alignment between HST and kinematics and what this is about. There should not be needed an alignment since the coordinate system is universally chosen. Maybe it is between the neural network training and the actual data cube, then fine, but it should be specified as such.

Another comment is that many of the NN training parts are very hard coded and not obvious to the user. I'd suggest adding a separate training configuration file that then can be linked in the actual likelihood module to cleanly separate the training from the evaluation of the likelihood, as well as the definition of the data.
The goal should be that someone else could also re-train with a different training set and still be able to use this code.

Then the test coverage has quite a bit decreased, despite moving to the EPL and adding your package to the test requirements. Unfortunately, the exact lines are not shown on the coverage report but you can also run it locally with 'make coverage' (have a look at the make file).

Thank you very much!

lenstronomy/Data/kinematic_bin_2D.py
```diff
@@ -0,0 +1,70 @@
...
+        :param transform_pix2angle: 2x2 array, mapping of pixel (unbinned image) to coordinate
+        :param ra_shift:  float, RA shift of pixel grid
+        :param dec_shift: float, DEC shift of pixel grid
+
+        """
+
+        nx, ny = np.shape(bin_mask)
+        self._nx = nx
+        self._ny = ny
+        if transform_pix2angle is None:
```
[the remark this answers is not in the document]
mattgomer: removed possibility to have no transform_pix2angle. Is that what you mean?

lenstronomy/Data/kinematic_bin_2D.py
```diff
@@ -0,0 +1,70 @@
...
+        self._data = bin_data
+        self._covariance = bin_cov
+        self._bin_mask = bin_mask
+
+    def binned_image(self):
+        """
+        Creates the binned image of the kinemmatic
+        """
+        binned_image = np.zeros_like(self._bin_mask)
+        for idx, value in enumerate(self._data):
```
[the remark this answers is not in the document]
mattgomer: I'm not sure how to do this faster? The mask isn't boolean- each pixel is an integer that identifies the bin, so I have to loop over these indices, don't I?

lenstronomy/Data/kinematic_data_2D.py
```diff
@@ -0,0 +1,21 @@
+import numpy as np
+
+from lenstronomy.Data.psf import PSF
+from lenstronomy.Data.kinematic_bin_2D import KinBin
+
+__all__ = ['KinData']
+
+class KinData(object):
```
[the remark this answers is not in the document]
mattgomer: Removed this class

lenstronomy/Util/kin_sampling_util.py
mattgomer: This part exists because we have imaging data in "HST" coordinates, and kinematic data in "MUSE" coordinates, which you can relate to one another through RA and DEC. Furthermore, we have the lens model, which is in HST coordinates (imaging model), and might be off center or with a misaligned PA compared to this coordinate systems. The NN returns a model aligned with the x axis, so this must be rotated into the same as the MUSE system, but that rotation and translation depends on the lens model PA, which relates it to the HST system, then to RA/DEC, then to the MUSE system to be interpolated.

lenstronomy/Util/image_util.py
mattgomer: How does this work exactly? I tried pip install sphinx and typed in the terminal "make docs" and it gives me a bunch of messages saying XXX.rst already exists. When I try to open for example Sampling.Likelihoods.rst, I don't see the kinematic sampling there, so I don't think it generated the up-to-date docs.

test/test_Util/coolest_template_update_pyAPI.json
mattgomer: No idea. Maybe @LyneVdV can help? Is coolest up to date for lenstronomy? Have I perhaps pushed a different version?

lenstronomy/Sampling/likelihood.py
mattgomer: We have the same problem because we allow for the mass q to be different than the light q, so both need sampled separately.

lenstronomy/Util/image_util.py
```diff
@@ -117,6 +117,17 @@ def rotateImage(img, angle):
     imgR = ndimage.rotate(img, angle, reshape=False)
     return imgR
 
+@export
+def shiftImage(img, shift):
```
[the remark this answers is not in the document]
mattgomer: fixed, but I'm not entirely sure that shift_image is ever used. I don't see any usage cases aside from the test.

lenstronomy/Data/kinematic_bin_2D.py
```diff
@@ -0,0 +1,70 @@
...
+        self._nx = nx
+        self._ny = ny
+        if transform_pix2angle is None:
+            transform_pix2angle = np.array([[1, 0], [0, 1]])
+        self.PixelGrid =   PixelGrid(nx, ny, transform_pix2angle, ra_at_xy_0 + ra_shift, dec_at_xy_0 + dec_shift)
+
+        self._data = bin_data
+        self._covariance = bin_cov
+        self._bin_mask = bin_mask
+
```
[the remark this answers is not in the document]
mattgomer: Can you give an example of what you mean by this? Maybe another part of the code I can adapt? As I understand, the point is to assign these parameters without calling class(inputs).data and instead it would just be class.data, for example. But when the class is initialized, it is given bin_data as an argument, so the inputs are required at some point, right?

test/test_Sampling/test_Likelihoods/test_kinematic_2D_likelihood.py
mattgomer: Added a test that when one of the vrms bin values is increased by 1 sigma, the likelihood changes by 1 sigma. Not sure if this is what you mean. If you want to put an input parameter off by 1 sigma, that's hard to define. What is the a priori sigma on r_eff, for example? Rather, that sigma is estimated by sampling logL, right?

lenstronomy/Sampling/Likelihoods/kinematic_NN_call.py
```diff
@@ -0,0 +1,69 @@
...
+        plt.title('Prediction')
+        plt.colorbar()
+
+    def check_bounds(self,input_p,same_orientation=True,verbose=False):
+        within_bounds=True
+        if not same_orientation:
+            if verbose:
+                print('NN CALL WARNING: Mass and light have different PAs!')
+            within_bounds = False
+        training_abs_bounds = {'q_mass': [0.6, 1.0],
```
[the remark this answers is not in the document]
mattgomer: I've partially fixed this by adding a SKiNN_1.0_config.json file. Right now I added it to Likelihoods because I'm not sure where else to put it. This reads the NN absolute bounds from this file which can be switched out if needed. This file also gives the grid settings of the NN when imported.

The reason this issue is only partially solved is because of some bounds which are expressed in terms of others. For example, the inclination must be greater than arccos(q_light) and arccos(q_mass), meaning its bounds are a rather involved function of the other parameters. I don't see how to include this in the json file, so it's still hardcoded.

lenstronomy/Sampling/Likelihoods/kinematic_2D_likelihood.py
```diff
@@ -0,0 +1,190 @@
...
+        self.z_lens = self.lens_model_class.z_lens
+        self.lens_light_model_class = lens_light_model_class
+        self._idx_lens = idx_lens
+        self._idx_lens_light = idx_lens_light
+        self.kin_class = kinematic_data_2D_class
+
+        self.kin_input = self.kin_class.KinBin.kin_bin2kwargs()
+        self.image_input = self.kwargs_data2image_input(kwargs_data)
+        self.kinematic_NN = kinematic_NN_call.kinematic_NN()
+        self.kinNN_input = {'deltaPix':0.02, 'image': np.ones((551,551))}
```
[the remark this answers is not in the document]
mattgomer: Cleared up that this is initializing the classes that run the NN function and the grid alignment/rotation. That self.kinNN_input line now imports settings from the new SKiNN_1.0_config.json file. See comment below.

mattgomer:
Hi @sibirrer. I've implemented most of these changes and left comments on some of the specifics that may not yet be entirely resolved. Coverage is better as well, but may still have room for improvement. Please take a look through and tell me what you think or if you have any more suggestions. Thanks!

lenstronomy/Util/kin_sampling_util.py
sibirrer: perhaps then re-name the HST and MUSE parts? Also, HST is a bit ambitious, because it does not necessarily has to correspond to the pixel axes as we can define arbitrary rotations and shifts of the coordinate system in the pixel data

test/test_Util/coolest_template_update_pyAPI.json
LyneVdV: since this is the _update_ file of the coolest interface (the file saving the results from a modeling procedure), I would assume that running all the tests (which includes the coolest tests) simply rewrite a new answer in the file each time. I agree this has nothing to do with SKiNN, it's just because at some point we must have launched al the tests and not just the ones related to SKiNN to see if everything works and that overwrote the value in that file.

sibirrer (commented):
Looks really good! Thank you @mattgomer !
I am still wondering about the test coverage.
Can you run 
$`make coverage `
locally (and pip install coverage, if not installed) to get a test coverage report locally, since the individual files are not displayed.
Let me know if you have questions. Sorry for being a bit slow atm

mattgomer:
@LyneVdV When I run make coverage, I get these coolest errors now, which look like 
`FAILED test/test_Util/test_coolest.py::TestCOOLESTinterface::test_load - AttributeError: 'COOLEST' object has no attribute 'exclude_keys'` 
Could I have the wrong version of COOLEST or something? Do you get these errors?

LyneVdV:
> @LyneVdV When I run make coverage, I get these coolest errors now, which look like `FAILED test/test_Util/test_coolest.py::TestCOOLESTinterface::test_load - AttributeError: 'COOLEST' object has no attribute 'exclude_keys'` Could I have the wrong version of COOLEST or something? Do you get these errors?

I don't have that error (all the test works well...) but I see where it may be coming from .

lenstronomy/Util/kin_sampling_util.py
mattgomer: Changed the names from HST and MUSE to imaging and spectra

mattgomer:
@sibirrer I tried make coverage, but had to remove test_coolest.py because it was throwing errors, see #497. When I make the coverage, it looks much better. Total coverage is 92%, kinematic_2D_likelihood is 96%, kinematic_NN_call is 94%. The remaining lines are exceptions if SKiNN is not installed properly, which I'm not entirely sure how to test.

Still having some problems making the docs. I thought I fixed the docstring issue with this commit, but I'm not sure now when viewing differences on the browser. It looks fine on the version I pushed, but I'd like to make the docs to be sure. Last issue I think is the staticmethod issue, which I was wondering if there's an example I can adapt?

lenstronomy/Data/kinematic_bin_2D.py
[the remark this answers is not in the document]
sibirrer: here is an example: https://github.com/lenstronomy/lenstronomy/blob/719a81c992fb0c7302e81163181432ea3a94fd51/lenstronomy/LensModel/Profiles/nfw.py#L127

basically, it means that the definition can also be called outside of the class (since it does not rely on any self. variables of the class itself). Don't forget to remove self from the input argument

sibirrer:
Thank you @mattgomer ! I merged @LyneVdV 's PR on COOLEST, so I hope if you pull the main into your PR and re-run the tests, it hopefully works. Also, on the CI platform, the tests run, so that's mostly what I care :)

sibirrer (approved):

sibirrer:
Looks great @mattgomer !
@ajshajib: this PR is supposed to be merged soon and that might (again) conflict your black formatting. Feel free to merge either one in the sequence you prefer

ajshajib:
@sibirrer, I will try merging #502 first, and then I can run the formatter on this PR myself to resolve any conflicts. 💪

ajshajib:
The test failed. Looks like it has to do with merging involving the earlier PR on `nautilus`.  I will do a couple of reverts now and try another way to merge #502 and this PR.

PR #492, opened 2023-08-08
Update coolest interface (coolest v0.1.0)

LyneVdV:
As COOLEST is evolving, I am updating the interface with lenstronomy. The actual version of coolest is now v0.1.0. The difference with the previous version is
- the external shear now belongs to a more general category of "Mass Field" (that could for instance also include flexion shifts).
- the exclude_keys (previously used to inform about non-existing keywords) do not exist anymore.


sibirrer (changes requested):
Thank you @LyneVdV !
Can you perhaps specify the specific COOLEST version in the requirements of lenstronomy, such that it's clear what version is required and tested against?
Thank you very much!

sibirrer (approved):

PR #497, opened 2023-08-16
Update coolest interface

LyneVdV:
As @mattgomer had issues when running the coverage for the SKiNN implementation #489 that were linked to the coolest interface, I updated the code to hopefully remove the error (at least, I don't have any issue when doing coverage on my side). The piece of code pointed in #489 used the `exclude_keys` that is an outdated feature of coolest. I removed any reference to this `exclude_keys` and also took the opportunity to make a small improvement (transforming MOCK type into MAP type when saving results).


sibirrer (approved):
Thank you @LyneVdV !
Let's see whether then @mattgomer's PR will work after this

PR #498, opened 2023-08-16
updated nautilus implementation, streamlined all nested samplers

johannesulf:
The following PR primarily updates the interface to nautilus. For example, previous versions did not allow the user to pass arguments to the `run` function of the sampler. I also updated the other samplers by removing a decent amount of redundant code. Sampler tests passed on my local computer. However, some workflow test including PolyChord fails. At the same time, it does so also for the `main` branch but for a different reason. I don't know why they fail. It's related to `nestcheck`. So some sort of quality control didn't pass. But the error message itself doesn't help me. I would definitely have some input and maybe some further tests. @sibirrer How would you like to proceed?


sibirrer (approved):
Thank you very much @johannesulf ! It looks like the tests are passing. It might be that some of the nested samplers are either outdated or require a bit more tedious installation, but they are not part of the CI test suite and only the user who wants to run them, should test them.

sibirrer:
Happy to merge this PR @johannesulf. You can also put your name in the AUTHORS.rst file, otherwise I will do it :)

johannesulf:
PolyChord is indeed tedious to install. As far as I can tell, the implementation currently may already be broken. So it doesn't seem like the PR causes any additional problems. Dynesty, nautilus and MultiNest seem to work. Eventually, it may make sense to have more extensive tests to ensure the sampler is somewhat accurate. Right now, the tests just check for crashes.

And yes, please add me to the author list. :)

PR #509, opened 2023-08-24
hardcode a more physical multipole convention in the EPL_BOXYDISKY profile

dangilman:
I add an automatic scaling of the multipole amplitude in the EPL_BOXYDISKY profile given by rescaling = theta_E / sqrt(q). @Maverick-Oh @LyneVdV maybe you guys have some input on this


lenstronomy/LensModel/Profiles/epl_boxydisky.py
```diff
@@ -54,24 +54,32 @@ def __init__(self):
...
-        :param theta_E: Einstein radius
-        :param gamma: log-slope of EPL mass profile
-        :param e1: ellipticity of EPL profile (along 1st axis)
-        :param e2: ellipticity of EPL profile (along 2nd axis)
-        :param a_m: amplitude of the multipole mass profile
-        :param center_x: center of the profile
-        :param center_y: center of the profile
-        :return: the keyword arguments for the joint profile
-        """
+        """This function splits the keyword arguments for the epl and multipole profiles
```
sibirrer, line 57: @ajshajib looks like pre-commit gives here some weird formatting that is not what you had when committing your code. Is there a way to change that?

ajshajib: I think if there were a new line after the description line/paragraph and the first `:param ...:` bit in the original commit, then `docformatter` would work more smoothly.

ajshajib: I already fixed it by first commenting a suggestion and then accepting that as a new commit. 😉 But, in the future, such small blemishes could be fixed by hand using this trick, or even better if the original commits had the blank line not to make `docformatter` go bonkers. 🙂

LyneVdV:
> I add an automatic scaling of the multipole amplitude in the EPL_BOXYDISKY profile given by rescaling = theta_E / sqrt(q). @Maverick-Oh @LyneVdV maybe you guys have some input on this

That's a great idea to have this factor implemented ! Nevertheless, the name "a_m" might not be appropriate anymore. First, the "m" in "a_m" is always 4 as we only speak about boxyness and diskyness, it would be better to have the subscript 4 for consistency. Secondly (and it's more tricky), the theta_E/sqrt(q) factor is useful to convert between the "mathematical quantity used for lensing" `a_4` introduced in the MULTIPOLE class and the "physical quantity" `a_4/a` which is the deformation of an ellipse (here, isodensity contour) of semi-major axis of length `a`. As the 2 concepts are different, different names might me needed to avoid confusion of the definitions. To sum up, why not renaming `a_m` in this EPL_BOXYDISKY profile into `a_4_over_a` or `a4_a` or `deviation_strength` or something else (pick your favorite or find a better one) ? And updating the documentation such that is it cleraly written that this parameter is the "strength of the multipolar deviation of order 4 of the elliptical isodensity contours, which is translated into the multipole strength from the MULTIPOLE class through $$the equation$$. If xx<0, the profile is boxy and if xx>0, the profile is discy" or something similar, you can even cite equation 3 of https://ui.adsabs.harvard.edu/abs/2022A%26A...659A.127V/abstract if it's easier. That's just a suggestion of course, feel free to ignore it if you think it would confuse people even more.

dangilman:
> > I add an automatic scaling of the multipole amplitude in the EPL_BOXYDISKY profile given by rescaling = theta_E / sqrt(q). @Maverick-Oh @LyneVdV maybe you guys have some input on this
> 
> That's a great idea to have this factor implemented ! Nevertheless, the name "a_m" might not be appropriate anymore. First, the "m" in "a_m" is always 4 as we only speak about boxyness and diskyness, it would be better to have the subscript 4 for consistency. Secondly (and it's more tricky), the theta_E/sqrt(q) factor is useful to convert between the "mathematical quantity used for lensing" `a_4` introduced in the MULTIPOLE class and the "physical quantity" `a_4/a` which is the deformation of an ellipse (here, isodensity contour) of semi-major axis of length `a`. As the 2 concepts are different, different names might me needed to avoid confusion of the definitions. To sum up, why not renaming `a_m` in this EPL_BOXYDISKY profile into `a_4_over_a` or `a4_a` or `deviation_strength` or something else (pick your favorite or find a better one) ? And updating the documentation such that is it cleraly written that this parameter is the "strength of the multipolar deviation of order 4 of the elliptical isodensity contours, which is translated into the multipole strength from the MULTIPOLE class through _the equation_. If xx<0, the profile is boxy and if xx>0, the profile is discy" or something similar, you can even cite equation 3 of https://ui.adsabs.harvard.edu/abs/2022A%26A...659A.127V/abstract if it's easier. That's just a suggestion of course, feel free to ignore it if you think it would confuse people even more.

Hi @LyneVdV, sorry for the delay in getting back with you. I think your suggestion regarding renaming the profile is a good one, I'll implement as a4_a

Maverick-Oh:
Hello @dangilman, thanks for your work! Sorry for my late response because of my vacation. The rescaling looks good!

I agree with the point @LyneVdV made. Because of the different conventions, it's easy to get confused. I would like to add some more references so that readers can look into them if they wonder what's going on. I would like to mention Keeton (2002) together. The arxiv link is here: [https://arxiv.org/abs/astro-ph/0210318]

I can work on adding comments too. Let me know. Thanks!

dangilman:
Hi everyone, I will go ahead and merge this pull request now

PR #511, opened 2023-08-31
Feature/nfwmc ellipse

ajshajib:
This PR adds an elliptical NFW mass model parametrized in the mass-concentration space, extending the previously available class for the spherical case.


lenstronomy/LensModel/Profiles/nfw_mass_concentration_ellipse.py
```diff
@@ -0,0 +1,107 @@
...
+    def __init__(self, z_lens, z_source, cosmo=None, static=False):
+        """
+
+        :param z_lens: redshift of lens
+        :param z_source: redshift of source
+        :param cosmo: astropy cosmology instance
+        :param static: boolean, if True, only operates with fixed parameter values
+        """
+        super(NFWMCEllipse, self).__init__(z_lens, z_source, cosmo, static)
+        self._nfw = NFW_ELLIPSE()
```
sibirrer, line 50: I tend to transition to the elliptical convergence NFW profile as the new default 'CSE', which is accurate and more physical, I am also thinking to change the convention overall of NFW_ELLIPSE to query the CSE decomposed profile

sibirrer (commented):
Looks good to me. I have one comment since you use an ellipse whether we should move it to use the ellipse defined in convergence, which is much more physical?

lenstronomy/LensModel/Profiles/nfw_mass_concentration_ellipse.py
ajshajib: That sounds good to me. It's done. 🙂

sibirrer (approved):
looks great, thank you!

PR #512, opened 2023-09-04
CosmoInterp with interpolated angular diameter distances as inputs

sibirrer:
This update adds the option for the CosmoInterp function (with is mirroring the Astropy.cosmology API) to have interpolated angular diameter distances as inputs instead of an astropy cosmology class


sibirrer:
I decide to merge it as I want to make a new release to update hierArc aside.

PR #517, opened 2023-09-15
Implementation of `hessian_z1z2` in `LensModel.lens_model`

trivialTZ:
**Implementation of `hessian_z1z2` in `LensModel`**
Added a method to handle the Hessian calculation in multi-plane for arbitrary redshifts of the lens and source in `LensModel.lens_model`. This builds upon the core function from `multi_plane.hessian_z1z2`.

### DIFF
--- a/lenstronomy/LensModel/lens_model.py
+++ b/lenstronomy/LensModel/lens_model.py
@@ -516,3 +516,29 @@ class LensModel(object):
         f_yy = (alpha_dec_pp - alpha_dec_pn + alpha_dec_np - alpha_dec_nn) / diff / 2
 
         return f_xx, f_xy, f_yx, f_yy
+
+    def hessian_z1z2(self, z1, z2, theta_x, theta_y, kwargs_lens, diff=0.00000001):
+        """Computes Hessian matrix when Observed at z1 with rays going to z2 with z1 <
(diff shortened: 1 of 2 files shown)


lenstronomy/LensModel/lens_model.py
```diff
@@ -516,3 +516,31 @@ def _hessian_differential_square(self, x, y, kwargs, k=None, diff=0.00001):
...
+        :param z1: Observer redshift
+        :param z2: source redshift
+        :param theta_x: angular position and direction of the ray
+        :param theta_y: angular position and direction of the ray
+        :param kwargs_lens: list of keyword arguments of lens model parameters matching
+            the lens model classes
+        :param diff: numerical differential step (float)
+        :return: f_xx, f_xy, f_yx, f_yy
+        """
+        if diff is None:
```
sibirrer, line 533: I don't think you need this raise error here. diff has a default argument that works and that can be set differently, and the documentation states that it needs to be a float

sibirrer (changes requested):
Looks really good. I made a small comment to simplify the code as I don't think you need a raise message when diff=None as this is not a valid input argument of the definition

sibirrer (approved):
Looks great! Thank you!

PR #520, opened 2023-09-24
Tracer model for metallicity gradient calculations

sibirrer:
This PR adds a Tracer model module for separate calculations of luminosity weighted tracers (such as metallicities)


sibirrer:
under time-pressure I will merge this (as there is a paper to be put on arXiv. 
@ajshajib: seems like coveralls is not displaying the missed lines anymore. Should we move to another service (such as the SLSim one)?

PR #524, opened 2023-09-28
Debug findOverlap parantheses of image_util.py

Maverick-Oh:
Previously, the parentheses of abs() in findOverlap was wrongly placed. It is a crucial typo but somehow it wasn't fixed earlier.

For example, with the previous code, if you give 
findOverlap([1, 0, 2, 3], [0, 0, 2, 3], 0.1) 

then the output comes out as (array([1, 2, 3]), array([0, 2, 3])), deleting the second coordinate even though it does not overlap in 0.1 distance.

After correcting the debug, now it works okay.

### DIFF
--- a/lenstronomy/Util/image_util.py
+++ b/lenstronomy/Util/image_util.py
@@ -179,9 +179,8 @@ def findOverlap(x_mins, y_mins, min_distance):
         else:
             for j in range(0, i):
                 if abs(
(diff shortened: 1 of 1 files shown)


sibirrer (approved):
uh, that is important. Hard to believe that this slipped through testing. I'll add a test as well.
Thank you very much @Maverick-Oh !!

PR #531, opened 2023-10-19
first commit of decoupled multiplane class

dangilman:
Hi @sibirrer, have a look at my initial attempt at implementing this methodology. As I mentioned earlier, modifying the lenstronomy source is not strictly necessary, but doing it in this way should make it much easier to use the new class within the existing lenstronomy code framework.


sibirrer:
@dangilman do you want me to still have a look at it?

PR #539, opened 2023-10-25
decoupled multi-plane lensing 

dangilman:
I implement a new multi-plane lens model class called MultiPlaneDecoupled that effectively transforms a multi-plane realization into a single plane realization. The assumption being made is that there is one massive deflector that dominates over line-of-sight objects, and that the coupling between the main deflector and halos in between the main lens plane and the source is unimportant compared to the existence of a non-linear deflection field in the first place. 

We make an initial guess of the main deflector to compute a deflection field associated with deflectors between the main lens plane and the source plane. This deflection field is then interpolated and used for any subsequent multi-plane lensing computations, leading to a factor 10-100 increase in speed for ray-tracing simulations with many halos in between the main deflector and the source plane. 

I also have an example notebook that does lens mass/source light reconstruction with the new class and a full realization of subhalos and line-of-sight halos generated with pyHalo.

### DIFF
--- a/lenstronomy/LensModel/Profiles/radial_interpolated.py
+++ b/lenstronomy/LensModel/Profiles/radial_interpolated.py
@@ -125,8 +125,7 @@ class RadialInterpolate(LensProfileBase):
         :type kappa_r: array of same size as r_bin
         :return: radial deflection angle
         """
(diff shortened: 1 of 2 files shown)


sibirrer (changes requested):
Looks really good. There is an error in the CI however

PR #572, opened 2024-01-19
Fix unsupported arguments  for image_linear_solve methods

aymgal:
I came across this issue while retrieving the fitting sequence class best-fit kwargs, which are now augmented by a `kwargs_tracer_source` placeholder, but it seems that this kwargs is not (yet?) supported in linear solver classes created via `class_creator.create_im_sim`. This PR is an easy fix for the constructor of `MultiBandImageReconstruction` following what is already done in the `ModelBand` class. I don't know what `kwargs_tracer_source` is made for, so please let me know if this fix is not appropriate for some reason.


sibirrer (approved):
Thank you very much @aymgal ! The model describes a 'tracer', which can be metallicity, velocity or any other property whose measurement signal is scaled with the brightness but its behavior can be differently. This is work in progress by @benmetha

Issue #573, opened 2024-01-19
Multiple select of single band kwargs


aymgal:
I think [this line in `image_linear_solve`](https://github.com/lenstronomy/lenstronomy/blob/43b5223346bcd95ef2adc927b33bb71e767b69f0/lenstronomy/ImSim/image_linear_solve.py#L472C31-L472C51) should call `_update_linear_kwargs()` instead of `update_linear_kwargs()`, since `update_linear_kwargs()` is overridden by classes like `SingleBandMultiModel`. In a multi-band setting, this causes the `select_kwargs()` method to be called two times, which leads to an error because kwargs that do not belong to the current single band being updated have already been removed.

I don't propose a PR since it seems to me that it's related to recent changes that are work in progress? (plus it can get tricky very quickly with these linear solver classes, so I'd rather leave this to the core developers).

ajshajib:
If #574 has solved this, should this be closed? Or, is there more to be done?

sibirrer:
can be closed, thank you!

PR #574, opened 2024-01-19
fix issue #573

sibirrer:


aymgal (approved):
Thanks for the fix @sibirrer !

PR #575, opened 2024-01-22
Fix issues with partial kwargs not being updated

aymgal:
Fixing similar bugs as in last two fixes.
- in `MultiBandImageReconstruction`, due to the deepcopy, the linear amplitudes were not updated by the constructor, which is here fixed by popping and adding back the `tracer_source` kwargs.
- in `ModelBandPlot`, the method `self._bandmodel.image(...)` was re-selecting the (already) single-band kwargs, so it should call `._image()` instead.

### DIFF
--- a/lenstronomy/Analysis/image_reconstruction.py
+++ b/lenstronomy/Analysis/image_reconstruction.py
@@ -58,8 +58,8 @@ class MultiBandImageReconstruction(object):
         source_marg = kwargs_likelihood.get("source_marg", False)
         linear_prior = kwargs_likelihood.get("linear_prior", None)
         bands_compute = kwargs_likelihood.get("bands_compute", None)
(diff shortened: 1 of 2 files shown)


sibirrer (approved):
Thank you very much @aymgal !
Might need in the future an other way of not accessing private definitions but this fixes it, indeed

ajshajib:
> Thank you very much @aymgal ! Might need in the future an other way of not accessing private definitions but this fixes it, indeed

Thanks a lot, @sibirrer, for the quick approval and merging. I was just about to provide some comments, and it's indeed about accessing the private method. I prefer to have it fixed now instead of leaving it for the future with the risk of forgetting about it. The quickest workaround would be to make a public method in the relevant class that would wrap around the private method.

sibirrer:
That makes sense @ajshajib . Sorry about the too quick merging

PR #579, opened 2024-01-29
Add option to use EPL instead of PEMD in COOLEST interface

aymgal:


aymgal:
@sibirrer Ok makes sense, I changed the default value then.

sibirrer:
@aymgal can you merge the latest  main version into your PR? The CI tests should pass now. Thank you!

PR #581, opened 2024-02-02
Update requirements.txt

ohannuks:
Currently numba_util.py implementation uses numba.generated_jit which is no longer available in numba 0.57.0 and further.

https://github.com/lenstronomy/lenstronomy/blob/d81ab6402e5727fc01a31d31713a8b4215ae51c5/lenstronomy/Util/numba_util.py#L79-L86

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ mpmath
 emcee>=3.0.0
 matplotlib
 scikit-learn
-numba>=0.43.1
+numba>=0.43.1,<=0.56.4
 dynesty
 pymultinest
 nestcheck


sibirrer:
Thank you very much @ohannuks for the PR and the notification that the very latest numba version does not support generate_jit anymore. Unfortunately, for the version you set, the test fails due to incompatibility of numba with python 3.11. See version compatibility here: https://numba.readthedocs.io/en/stable/user/installing.html#version-support-information do we need to downgrade to 0.56 or would 0.57 also work?

ajshajib:
Thanks @ohannuks for the PR! After looking into it, `numba.generated_jit()` has been deprecated since `numba.extending.overload()` provides a more powerful implementation of the same features (https://github.com/numba/numba/issues/8466). So, instead of downgrading the `numba` version, can you kindly try replacing `generated_jit()` function with `overload()` in this PR?

sibirrer:
@ohannuks fyi, another PR has set the requirements of numba to be <0.58 such that tests do not break down (numba version 0.57 for Python 3.11). This is just a temporary fix and as @ajshajib suggested, having the deprecated functions replaced is the way forward.

sibirrer:
@ohannuks I made a few changes such that with the latest version on the main branch, the recent numba version is compatible. Let me know if this does not work for you. It's not yet a pip version, but if it solves your issue, then we can make a PyPi release soon - thank you!

ohannuks:
Thanks a lot @sibirrer , @ajshajib . Yes, I think the latest version fixes this compatibility issue. Everything looks green now.

PR #583, opened 2024-02-06
simplify and fix numba issue

sibirrer:
more recent numba versions support np.nan_to_num() and do not require anymore a rather complex work around (that became deprecated). I removed this work around and now no deprecated functions are being called, while the same speed is hopefully allowed. This might also solve the issue raised in #581


ajshajib (approved):
Looks good to me! 🙂

PR #585, opened 2024-02-14
Add tracer types, allowing metallicity data to be used in observer units.

astrobenji:
Adds a new parameter, "tracer_type", to the tracer module. If this is linear, components are added linearly. If this is metallicity, then the metallicity in units of log(O/H)+12 is first converted to linear units, then added, then converted back into convenient logarithmic units. The default setting for tracers is still linear.

### DIFF
--- a/lenstronomy/ImSim/tracer_model.py
+++ b/lenstronomy/ImSim/tracer_model.py
@@ -139,7 +139,7 @@ class TracerModelSource(ImageModel):
         if self._tracer_type == "LINEAR":
             return tracer_brightness_conv / source_light_conv
         if self._tracer_type == "LOG":
-            return np.log10(tracer_brightness_conv / source_light_conv) 
+            return np.log10(tracer_brightness_conv / source_light_conv)
 
     def _tracer_model_source(
         self, kwargs_tracer_source, kwargs_lens, de_lensed=False, k=None
--- a/lenstronomy/ImSim/tracer_model.py
+++ b/lenstronomy/ImSim/tracer_model.py
@@ -45,8 +45,8 @@ class TracerModelSource(ImageModel):
         :param tracer_partition: in case of tracer models for specific sub-parts of the surface brightness model
          [[list of light profiles, list of tracer profiles], [list of light profiles, list of tracer profiles], [...], ...]
         :type tracer_partition: None or list
-        :param tracer_type: LINEAR or METALLICITY; to determine how tracers are summed between components, as metallicities
-         which use log units should not be added linearly.
+        :param tracer_type: LINEAR or LOG. If the tracer is in log units, it is converted to linear units, summed,
+         and then converted back to log units.
         :type tracer_partition: str
         """
         if likelihood_mask is None:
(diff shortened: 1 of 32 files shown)


lenstronomy/ImSim/tracer_model.py
```diff
@@ -44,6 +45,9 @@ def __init__(
         :param tracer_partition: in case of tracer models for specific sub-parts of the surface brightness model
          [[list of light profiles, list of tracer profiles], [list of light profiles, list of tracer profiles], [...], ...]
         :type tracer_partition: None or list
+        :param tracer_type: LINEAR or METALLICITY; to determine how tracers are summed between components, as metallicities
```
sibirrer, line 48: can you rename the type to LOG instead of METALLICITY? This is more explicitly, and it might not always be true that when you have line ratios, that you want to sum them in log space (the units of metallicity are log, but you can also define a linear line ratio unit)

sibirrer (changes requested):
Thank you @benmetha. I have one request to change, and then there are no tests for the  newly added component. Can you write a test function that will call this feature and test it? You can have a look at the test folder of the tracer_model. Thank you!

test/test_ImSim/test_tracer_model.py
```diff
@@ -112,3 +113,37 @@ def test_tracer_model(self):
         )
         tracer_model_2 = tracer_brightness_conv / light_convolved
         npt.assert_almost_equal(tracer_model_2, tracer_model, decimal=5)
+        
+    def test_log_and_linear_tracer_modes(self):
+    	"""Convolve a tracer with log units with a surface brightness profile in two ways:
+
+     (1) using the tracer_type="LOG" argument (2) converting units to linear ones, using
```
sibirrer, line 120: something is odd here with the re-formatting (the pre-commit also failed). Can you check on this and perhaps run black locally with

```
black .
docformatter -r ./* --black --in-place
```

sibirrer (changes requested):
Thank you very much! It looks good and thanks also for the test.
There is an issue with the formatting in the test file. Hope you can solve it either manually or through the terminal running black

test/test_ImSim/test_tracer_model.py
```diff
@@ -117,8 +117,9 @@ def test_tracer_model(self):
     def test_log_and_linear_tracer_modes(self):
     	"""Convolve a tracer with log units with a surface brightness profile in two ways:
 
-     (1) using the tracer_type="LOG" argument (2) converting units to linear ones, using
-     tracer_type="LINEAR", and     converting back to log units.
+     (1) using the tracer_type="LOG" argument
```
sibirrer, line 120: you need to move these strings one tab inwards, and perhaps not having a blank line in-between

sibirrer (changes requested):

test/test_ImSim/test_tracer_model.py
```diff
@@ -115,11 +115,10 @@ def test_tracer_model(self):
         npt.assert_almost_equal(tracer_model_2, tracer_model, decimal=5)
         
     def test_log_and_linear_tracer_modes(self):
-    	"""Convolve a tracer with log units with a surface brightness profile in two ways:
-
+    """Convolve a tracer with log units with a surface brightness profile in two ways:
```
sibirrer, line 118: sorry, mixed up inwards vs outwards. Everything needs to be aligned with the code block that is in the definition (see all other docstrings)

sibirrer:
@benmetha the two tests failed, see the link to the Tests/build

sibirrer:
Thank you very much @benmetha! I see that the de_lens=True option in the _tracer_model_source() definition has not been tested. Can you please add a test that executes this part. I doubt that it fails but this is just to make sure that this feature gets preserved by future development - thank you!

astrobenji:
Hi Simon, I never wrote the code for the `de_lens` option. I don't know how it is used. If you'd like, I could look into how its used in some of the other test cases that test image simulation, and try to write something that uses this option for traced quantities? But I don't understand how making a test case for that feature is related to this pull request.

PR #593, opened 2024-02-29
Fix typo in list of supported models

nataliehogg:
Extremely small PR to fix a typo in the list of supported models in ProfileListBase of the LensModel class.


sibirrer (approved):
Thank you for the fix @nataliehogg ! Sorry that this screwed up the "LOS" model.  There perhaps should be a better testing of the import of all models

nataliehogg:
No worries @sibirrer ! This hasn't caused any obvious problems, I just happened to see the typo whilst looking for something else. I can try to write a test for the imports :+1:

PR #596, opened 2024-03-11
add random seed to Optimizer class

dangilman:
I add a random seed parameter as an optional argument for the optimizer class. This allows the user to regenerate a lens model from a random seed when using this class


ajshajib (approved):

ajshajib:
Thanks @dangilman.

PR #597, opened 2024-03-12
more robust lens equation solver settings when dealing with imaging data

sibirrer:
I set the default as at least 200x200 grid sampling the imaging data when looking for solution of the lens equation. Prior to this PR, it takes the pixel resolution of the data, which can be not accurate when the pixels are too sparse (large pixel for small Einstein radius).


dangilman (approved):

dangilman:
thanks @sibirrer!

PR #600, opened 2024-03-23
definition for whether analytical solver is available

sibirrer:
this definition can help to decide whether or not to use the analytical lens equation solver


sibirrer:
I merge it and make a mini-release such that this functionality will be available for the SLSim package

PR #603, opened 2024-04-08
Fix bug in handing source positions constrained on image plane

ajshajib:
This PR fixes a bug where the code accepts a list of indices, but it's actually a boolean list.


sibirrer (approved):
Thank you very much @ajshajib!

PR #604, opened 2024-04-14
Create and Add Adaptive Boundary Mesh

avlando:
Implements an Adaptive Boundary Mesh (ABM) algorithm to achieve high-resolution microlensing simulations. Gravitational lensing effects caused by point mass lens and elliptical source. ABM algorithm dynamically adapts the mesh resolution around the source, efficiently capturing lensed image for M0 = 0.01M_sol to 0.1M_sol.


lenstronomy/LensModel/Microlensing/ABM.py
```diff
@@ -0,0 +1,333 @@
+import itertools
```
sibirrer, line 1: convention on file names and folders. File names are lower_case, folder names are UpperCase

lenstronomy/LensModel/Microlensing/ABM.py
```diff
@@ -0,0 +1,333 @@
...
+
+def splitting_centers(center, side_length, n_p):
+    """Takes square centered at center = (x, y) with side_length = side length.
+
+    (float) and divides it into n_p (integer) squares along each axis - so
+    returns n_p * n_p subsquares from the original square. In particular,
+    the function returns the coordinates of the centers of these subsquares,
+    along with the final side length of the subsquare.
+
+    :param center: nparray; center of square
```
sibirrer, line 15: a slightly better convention for the documentation strings is using the explicit :type <param_name>: convention
:param center: center of square
:type center: nparray

lenstronomy/LensModel/Microlensing/ABM.py
```diff
@@ -0,0 +1,333 @@
...
+                center_y - side_length / 2 + (i + 0.5) * side_length / n_p
+            )
+            k += 1
+
+    centers = np.column_stack((new_x, new_y))
+    new_side_length = side_length / n_p
+    return centers, new_side_length
+
+
+# define the microlens
```
sibirrer, line 43: all these lines of code that are not a function/definition I would suggest not having in the software package, or in the form of functions. The idea is that the user has full control over those and not being hard coded in a software package. You can keep using a notebook for your investigations using functions the functions from this file

lenstronomy/LensModel/Microlensing/ABM.py
```diff
@@ -0,0 +1,333 @@
...
+    {"amp": 1, "radius": size_s / 2, "e1": 0, "e2": 0, "center_x": 0, "center_y": 0}
+]
+surface_brightness = ligth.surface_brightness(beta_x, beta_y, kwargs_light)
+lensed_flux = np.sum(surface_brightness)
+
+reference_magnification = lensed_flux / unlensed_flux
+
+# define source parameters
+
+L = (
```
sibirrer, line 97: do you want these settings for example as default settings in your ABM function, or even be part of the high-level function for the user?

lenstronomy/LensModel/Microlensing/ABM.py
```diff
@@ -0,0 +1,333 @@
...
+n_p = 30
+eta = 0.7 * n_p
+source_position = (0, 0)
+
+# helper functions
+
+
+# defines loop_information to defines number of iterations and final scale factor
+def loop_information(eta, beta_0, beta_s):
+
```
sibirrer, line 113: can you provide the docstrings for this definition and the input and output variables as well?

lenstronomy/LensModel/Microlensing/ABM.py
```diff
@@ -0,0 +1,333 @@
...
+    :param center_points: nparray; Array of center points, each row containing (x, y)
+        coordinates. Source coordrates of grid
+    :param test_point: nparray; Coordinates of the test point (x, y). Source position
+    :param threshold: float; Distance threshold.
+    :return: nparray; Boolean array indicating whether each point is within the
+        threshold distance.
+    """
+
+    test_point = np.array(test_point)
```
sibirrer, line 140: since you require in the input statement that test_point and center_points are already numpy arrays, you don't need to explicitly convert those (in fact, this could possibly take quite a bit of computational cost)

lenstronomy/LensModel/Microlensing/ABM.py
```diff
@@ -0,0 +1,333 @@
...
+    L,
+    beta_0,
+    beta_s,
+    n_p,
+    eta,
+    number_of_iterations,
+    final_eta,
+    kwargs_lens,
+):
+    """Returns list of those high resolution image-plane pixels that were mapped to
```
sibirrer, line 164: this routine goes through an iterative adaptive process. Can you link to the original paper that introduced this algorithm?

lenstronomy/LensModel/Microlensing/ABM.py
```diff
@@ -0,0 +1,333 @@
...
+        # Increment iteration counter
+        i += 1
+
+    return subset_centers, side_length, total_number_of_rays_shot
+
+
+from lenstronomy.LensModel.lens_model import LensModel
+from lenstronomy.LightModel.light_model import LightModel
+
+# compute unlensed surface brightness
```
sibirrer, line 230: can you either define a function(s) for this task or move those to a jupyter notebook (not to be committed for now)

sibirrer (changes requested):
Thank you very much Alex! I have some minor suggestions/requests. Primarily that you should not have lines of code outside of definitions for the user to use. Think of this routines being the software that you (or any user) import on your notebook to facilitate the tasks you want to calculate

lenstronomy/LensModel/Microlensing/ABM.py
avlando: that is a python module like numpy, voided from code (used in previous commit)

lenstronomy/LensModel/Microlensing/ABM.py
avlando: I have added them into the ABM function with comments

test/test_LensModel/test_Microlensing/test_ABM.py
```diff
@@ -0,0 +1,75 @@
+import numpy as np
+import pytest
+from Microlensing.adaptive_boundary_mesh import loop_information
```
sibirrer, line 3: you have to use here the full lenstronomy paths, i.e. from lenstronomy.LensModel.Micorlensing.... import

sibirrer (changes requested):
The tests are failing due to incomplete import statements. Check my comments and I hope the automatic tests will then pass

lenstronomy/LensModel/Microlensing/adaptive_boundary_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+
+    centers = np.column_stack((new_x, new_y))
+    new_side_length = side_length / n_p
+    return centers, new_side_length
+
+
+def loop_information(eta, beta_0, beta_s):
+    """# Defines loop_information to defines number of iterations and final scale factor
+
+    :param eta: 0.7 * n_p
```
sibirrer, line 50: define what eta is as the input of this function, not how another function defined it prior to input of this function

lenstronomy/LensModel/Microlensing/adaptive_boundary_mesh.py
```diff
@@ -0,0 +1,202 @@
...
+    :type source_position: tuple
+    :param L: Side length of square area in image plane. Same as lenstronomy grid width
+    :type L: float
+    :param beta_0: Initial search radius (delta_beta)
+    :type beta_0: float
+    :param beta_s: Factor of 1/2 because radius
+    :type beta_s: float
+    :param n_p: Number of pixels
+    :type n_p: int
+    :param eta: 0.7 * n_p
```
sibirrer, line 129: define what eta is, not what input relations are used by users

sibirrer (changes requested):
some minor comments on the documentation

lenstronomy/LensModel/Microlensing/adaptive_boundary_mesh.py
```diff
@@ -0,0 +1,212 @@
...
+            (subset_image_centers_x, subset_image_centers_y)
+        )  # not used in calculation, used for debugging
+
+        # Update side length
+        side_length /= n_p
+
+        # Update delta_beta based on iteration number
+        if i < number_of_iterations:
+            delta_beta /= eta
+        elif i == number_of_iterations:
```
sibirrer, line 201: looking at the test coverage, it worries me that this line is not being tested. Can you check (and modify if needed) that this criteria is applied at the very last iteration?

sibirrer (changes requested):
Thank you! I have one comment/request that came up looking at the test coverage

lenstronomy/LensModel/Microlensing/adaptive_boundary_mesh.py
avlando: changed main loop to include final iteration, counter now starts at 2 to avoid extra iteration at end.

sibirrer (approved):
Thank you very much @avlando !

sibirrer (approved):
wohooo!

PR #611, opened 2024-05-19
Simple line segment LightModel for jet modeling

michael7198:
Hi all,

I've added a very simple line segment LightModel in order to facilitate modeling lensed AGN jets. In its current state the model simply produces a constant surface brightness rectangle, which could be used for a naive continuum subtraction. I'd very much appreciate input on how to structure tests for an addition like this, as well as any other general feedback. As implemented the model is pixel-based and therefore much less smooth than other light profiles, and I understand that could be a potential issue.

Thanks!

-Michael


lenstronomy/LightModel/Profiles/lineprofile.py
```diff
@@ -0,0 +1,69 @@
...
+        :param length: length of line (arcseconds)
+        :param width: width of line (arcseconds), line width extends symmetrically
+        :param start_x: ra coordinate of start of line
+        :param start_y: dec-coordinate of start of line
+        :return: surface brightness, raise as definition is not defined
+        """
+        ang = -np.deg2rad(angle)
+        x_ = np.cos(ang) * (start_x - x) + np.sin(ang) * (start_y - y)
+        y_ = np.cos(ang) * (start_y - y) - np.sin(ang) * (start_x - x)
+        return amp * (x_ > 0) * (x_ < length) * (abs(y_) < width / 2)
```
sibirrer, line 47: does this work when you are outside the bounds? I would suggest you use array convention and explicitly set it to zero outside the bounds, and amp inside it.

lenstronomy/LightModel/Profiles/lineprofile.py
```diff
@@ -0,0 +1,69 @@
+import numpy as np
+
+
+__all__ = ["LineProfile"]
+
+
+class LineProfile(object):
+    """Horizontal line segment class."""
```
sibirrer, line 8: might be worth describing this class a bit better

lenstronomy/LightModel/light_model_base.py
```diff
@@ -41,6 +41,7 @@
     "SLIT_STARLETS_GEN2",
     "LINEAR",
     "LINEAR_ELLIPSE",
+    "LINEPROFILE",
```
sibirrer, line 44: I'd suggest you naming it LINE_PROFILE instead for better readability

sibirrer (changes requested):
Hi @michael7198 thank you very much for the PR! It looks compatible with lenstronomy.
Now for testing, have a look at the test/test_LightModel/test_Profiles folder where for each light profile, there should be a test file. The idea is to test the specific functions with some known input-output behavior and see whether they execute.

And then you should test that the model can be properly integrated in LightModel, in the test_LightModel/test_light_model.py file

Let us know if you have questions!

lenstronomy/LightModel/Profiles/lineprofile.py
michael7198: I believe I've done this (created an array of zeros, then made a mask that selects elements in the bounds and sets them to amp) but let me know if you meant something else by "array convention".

test/test_LightModel/test_Profiles/test_lineprofile.py
```diff
@@ -0,0 +1,28 @@
...
+class TestLineProfile(object):
+    def setup_method(self):
+        self.lineprofile = LineProfile()
+
+    def test_function(self):
+        amp = 1
+        length = 1
+        width = 0.01
+        angle = 57
+        x = np.array([0, 1, 0.5*np.cos(np.deg2rad(57))])
```
sibirrer: can you make a test with an array structure (you did this) and a single-valued input? Thank you!

sibirrer (changes requested):
Looks great. I have one minor request in testing and then it should be good to go.

michael7198:
I've added a test to the test_profiles folder but am not sure what to do with the test_light_model.py file. For now I've added the model and its kwargs where relevant under self.setup_method and self.test_total_flux.

test/test_LightModel/test_Profiles/test_lineprofile.py
michael7198: Done!

sibirrer:
There are errors in the test @michael7198. Check out the link above. You can also execute these tests locally, by typing e.g.:
$ py.test test/test_LightModel

sibirrer (approved):

PR #612, opened 2024-05-22
added test function that show the centroiding of the PSF convolution

sibirrer:


sibirrer:
@martin-millon here is a PR with the test functions for centroiding of the convolutions. I checked that the light center is kept in the center pixel for all the convolution types in lenstronomy

PR #618, opened 2024-06-11
bug fix for GAUSSIAN PSF when being supersampled (kernel size)

sibirrer:
this fixes a bug for the GAUSSIAN PSF kernel size when being supersampled. FWHM was not considered in the size of the kernel


sibirrer:
@ajshajib any idea why the workflow did not proceed doing the CI?

Issue #621, opened 2024-06-19
Easy check of parameter bounds


amn3142:
It would be nice to get a quick alert if any of the parameters for the PSO are initialized outside of the parameter bounds.

sibirrer:
I close this issue for now @amn3142. Please reopen if needed.

PR #622, opened 2024-06-24
convenient checking of parameter settings to be in bounds

sibirrer:
This PR should address issue #621


sibirrer:
I merge it such that this feature gets available. Another PR coming with other features too...

PR #625, opened 2024-07-09
Update to lens_plot.py

jocelyn-mcmahon-16:
Added feature to lens_plot.py to allow for the plotting of multiple sources/images in the point_source_plot function. Allows for input of parameters _name_list_ and _index_, which respectively specify a new list of labels to be used for the images and, when used with multiple sources, labels each source's images with a number to more easily determine which image comes from which source.


lenstronomy/Plots/lens_plot.py
```diff
@@ -37,7 +39,9 @@ def lens_model_plot(
     coord_center_dec=0,
     coord_inverse=False,
     fast_caustic=True,
-    **kwargs
+    name_list=None,
```
sibirrer, line 42: can you document this new input variables in the doc strings

lenstronomy/Plots/lens_plot.py
```diff
@@ -283,12 +290,36 @@ def point_source_plot(
...
+    :param index: number of sources, an integer number. Default None.
     :param kwargs: additional plotting keyword arguments
     :return: matplotlib axis instance with figure
     """
     from lenstronomy.LensModel.Solver.lens_equation_solver import LensEquationSolver
 
+    if name_list is None and index is None:
+        name_list = _NAME_LIST
+    elif name_list is None and index is not None:
+        name_list = [
```
sibirrer, line 303: this line of code is not tested. Can you add a test that goes through this option?

sibirrer: do you really need this line of code? can't you just get name_list = _NAME_LIST and then also go through the for-loop the same way as for custom input for name_list?

lenstronomy/Plots/lens_plot.py
```diff
@@ -283,12 +290,36 @@ def point_source_plot(
...
+            f"{index+1}E",
+            f"{index+1}F",
+            f"{index+1}G",
+            f"{index+1}H",
+            f"{index+1}I",
+            f"{index+1}J",
+            f"{index+1}K",
+        ]
+    elif name_list is not None and index is None:
+        name_list = name_list
```
sibirrer, line 317: can you also make a test case with this option?

sibirrer (changes requested):
Looks great, thank you @jocelyn-mcmahon-16 ! I have some suggestions and a request for better test coverage

lenstronomy/Plots/lens_plot.py
jocelyn-mcmahon-16: Added a test for this one!

lenstronomy/Plots/lens_plot.py
jocelyn-mcmahon-16: I added a test for each case. I also updated it to loop over _NAME_LIST like you suggested, but I left it as a separate case.

lenstronomy/Plots/lens_plot.py
jocelyn-mcmahon-16: Added the variables!

sibirrer (approved):
Thank you very much @jocelyn-mcmahon-16 !

PR #627, opened 2024-07-15
Fixing a typo in point source plotting utility 

Hadrien-Pgnt:
When plotting the reconstructed source with ModelPlot.source_plot, the y coordinate of the point source should be `y_source * delta_pix` and not `y_source + 0.5 * delta_pix` to be consistent with the x coordinate (otherwise the y coordinate is out of range and the point source's location is not shown)


sibirrer (approved):
Thank you very much @Hadrien-Pgnt! This was a remnant from where the pixel grid was defined simply by mats how and pixel centers where not defined correctly.

PR #630, opened 2024-07-17
Update profile names for lens models with Gaussians

ajshajib:
This PR updates several lens-model class names and their associated keyword to make it explicit when ellipticity is defined in the potential, and make this naming convention uniform across all profiles.

These changes in the profile-name keywords break backward compatibility:

```
"CNFW_ELLIPSE" → "CNFW_ELLIPSE_POTENTIAL"
"HERNQUIST_ELLIPSE" → "HERNQUIST_ELLIPSE_POTENTIAL"
"NFW_ELLIPSE" → "NFW_ELLIPSE_POTENTIAL"
"NFW_MC_ELLIPSE" → "NFW_MC_ELLIPSE_POTENTIAL"
"PJAFFE_ELLIPSE" → "PJAFFE_ELLIPSE_POTENTIAL"
"TNFW_ELLIPSE" → "TNFW_ELLIPSE_POTENTIAL"
"MULTI_GAUSSIAN_KAPPA" → "MULTI_GAUSSIAN"
"MULTI_GAUSSIAN_KAPPA_ELLIPSE" → "MULTI_GAUSSIAN_ELLIPSE_POTENTIAL"
"GAUSSIAN" → "GAUSSIAN_POTENTIAL"
"GAUSSIAN_KAPPA" → "GAUSSIAN"
```


ajshajib:
@sibirrer, this PR didn't add any new code; it only renamed class names and keywords. I think the 1 line missing from the test coverage was already being missed, but since a longer class/variable name broke a previously single line into 3, the coverage statistics went down a bit.

sibirrer (approved):
Thank you @ajshajib ! This makes it indeed more transparent what the assumptions are in the different profiles.
On a similar note, I wonder whether we want also to re-name NFW_ELLIPSE_CSE into NFW_ELLIPSE since this is the much better numerical situation, and the same for HERNQUIST_ELLIPSE_CSE -> HERNQUIST_ELLIPSE? This can also be done in a different PR. Might just be good thinking about all of them at once that backwards incompatibility can be traced back to a specific version

ajshajib:
@sibirrer, I think "NFW_ELLIPSE_CSE" and "HERNQUIST_ELLIPSE_CSE" are fine as is, as they explicitly convey the numerical approximation used. There are similar names for "NFW_ELLIPSE_GAUSS_DEC" and "SERSIC_ELLIPSE_GAUSS_DEC".

PR #640, opened 2024-08-09
Small bug fix in CSE profile

ahuang314:
In rare situations when x is an integer, the code can crash on the += operation when the LHS is a numpy scalar int and the RHS is a float. This PR fixes that bug.

Fixed incorrect values in the upper limit defaults for the classes


sibirrer (approved):
Thank you very much @ahuang314 ! Can be merged after the tests passed

PR #644, opened 2024-08-30
Updated treatment of non-normalized PSF

sibirrer:
re-structured kernel_util and PSF() functions to keep track of normalized or non-normalized PSF


sibirrer:
@ahuang314 this PR should fix the issue with the PSF normalization (other than having the keywords passed in SimulationAPI

ahuang314 (approved):
Looks like it works, I'll follow up with the changes to the Observation and SingleBand classes in another PR

PR #645, opened 2024-08-30
adds normalisation kwarg to SingleBand and Observation classes

ahuang314:
Continuation of PR #644 to allow for the normalisation keyword to be passed into the SingleBand and Observation class


sibirrer (approved):
Thank you @ahuang314 !

PR #646, opened 2024-09-05
Update fitting_sequence.py

ajshajib:
This PR restores consistency with other places where it is expected to be "emcee".


sibirrer (approved):
oh, was not even noticing it. Thank you for the fix @ajshajib !

PR #648, opened 2024-09-07
Add "GNFW_ELLIPSE_GAUSS_DEC" lens model

ajshajib:
This PR adds the elliptical generalized-NFW lens model through the Gauss decomposition method.

### DIFF
--- a/lenstronomy/LensModel/Profiles/gauss_decomposition.py
+++ b/lenstronomy/LensModel/Profiles/gauss_decomposition.py
@@ -7,6 +7,7 @@ __author__ = "ajshajib"
 import numpy as np
 import abc
 from scipy.special import comb
+from scipy.special import hyp2f1
 
 from lenstronomy.LensModel.Profiles.multi_gaussian_ellipse_kappa import (
     MultiGaussianEllipseKappa,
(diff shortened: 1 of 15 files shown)


lenstronomy/LensModel/Profiles/gauss_decomposition.py
```diff
@@ -430,6 +435,111 @@ def get_scale(self, **kwargs):
...
+            * **alpha_Rs** (``float``) --
+              Deflection angle at ``Rs``
+            * **R_s** (``float``) --
+              gNFW scale radius
+
+        :return: projected NFW profile at y
+        :rtype: ``type(y)``
+        """
+        R_s = kwargs["Rs"]
+        alpha_Rs = kwargs["alpha_Rs"]
```
sibirrer, line 516: do you need the inputs to this definition as **kwargs when it is determined that the arguments required are alpha_Rs and gamma_in? Or what is the reason for it?

sibirrer (commented):
Thank you @ajshajib! I have one comment only, otherwise it's good to go.

lenstronomy/LensModel/Profiles/gauss_decomposition.py
ajshajib: Thanks, @sibirrer, for the review. The reason for using `**kwargs` instead of explicit argument definition is that these "GAUSS_DEC" classes use an abstract parent class to handle all Gaussian decomposition and lensing calculation, and the job of these individual child classes is only to define the 1D profile shape. Python has a non-strict rule requiring children-class methods to have the same function signature as the ones in the parent class. To avoid breaking this rule, I have left the arguments in the child class methods unspecified as `**kwargs`. However, the expected parameters for each class are still specified above the `__init__()` method.

ajshajib:
@sibirrer, I have added some more updates and bug fixes; please re-review. I have shuffled some lines in `likelihood.py` to potentially speed up sampling, the lines missing coverage in there were already missing it.

sibirrer (approved):
Looks good. Thank you very much @ajshajib !

PR #653, opened 2024-09-16
Reorder parameter extraction in LikelihoodModule

ajshajib:
This PR modifies the `LikelihoodModule` to ensure that the bounds check occurs before parameter extraction. This change validates the parameters first, potentially avoiding unnecessary computations if the bounds are not met.


sibirrer (approved):
Thank you @ajshajib !

PR #654, opened 2024-09-21
Make point source and lens equation flexible to enable different redshifts

sibirrer:
This PR implements:
- SinglePlane lens model with a deflection scaling accepting different source redshift for different z_source_conventions
- Makes analytical and numerical lens equation solver adaptive to different source redshifts
- Makes PointSource model flexible to have lensed point sources at different redshifts


lenstronomy/LensModel/Solver/lens_equation_solver.py
```diff
@@ -210,6 +219,21 @@ def image_position_analytical(
             raise ValueError(
                 "Only SIE, EPL, EPL_NUMBA (+shear +convergence) supported in the analytical solver for now."
             )
+
+        if self.lensModel.type != "SinglePlane":
+            raise ValueError(
```
ajshajib, line 224: Can you kindly add a test case with `with self.assertRaises(ValueError)` to cover this line?

ajshajib (approved):

PR #655, opened 2024-09-24
Enable batched fermat potential computation with EPL_NUMBA

smericks:
- I want to use the static method EPL_numba.function() to compute many fermat potentials from many lenses at one time
- Basically, I want to be able to add a batch dimension to the inputs, and keep everything working
- It turns out this already almost works, the only thing that is not batch compatible is the way in which niter is computed in the subfunction omega()
- If I take the max of the quantity f = (1 - q) / (1 + q) across the many lenses, and use that to determine niter, everything works.
- Since np.max(scalar) just returns that scalar, this should not change the interface for existing use cases


smericks:
Problem with numba overwrite of np.max() not handling an argument of type float64, I will edit accordingly

smericks:
Hi @sibirrer , test cases are now passing for this change, let me know if this is OK!

PR #663, opened 2024-10-14
Non-italic text in absolute residual plot colourbar label

nataliehogg:
Very tiny PR to change colourbar labels to non-italic text in the absolute residuals plot!


sibirrer (approved):

sibirrer:
Thank you very much @nataliehogg !
We ignore the pre-commit.ci for the moment and happy to merge

PR #673, opened 2024-11-22
minor change in the structure of lens_profile.py

huangxy256:
moved part of the "effective_einstein_radius" outside the LensProfileAnalysis class to be better accessible from outside


lenstronomy/Analysis/lens_profile.py
```diff
@@ -409,3 +379,48 @@ def einstein_radius_from_grid(
         return np.nan, 0
     else:
         return np.nan
+
+
+def effective_einstein_radius_from_radial_profile(r_array, kappa_r):
```
ajshajib, line 384: Can you make it still part of the `LensProfileAnalysis` class, but a static method? Or, are there downsides if that is done?

ajshajib (changes requested):

ajshajib (approved):
Thanks a lot, @huangxy256!

PR #674, opened 2024-11-25
add cb.ax.tick_params to all plot functions

nataliehogg:
Small PR which ensures colourbar tick labels of individual model plots follow the chosen fontsize.


sibirrer (approved):
Looks great, thank you!

Issue #678, opened 2024-12-02
Time-delay likelihood is incorrect if several point sources are used. 


martin-millon:
This statement is always False when self._measurement_bool_list is a list of lists, which is the case when you have several 'LENSED_POSITION' point sources: https://github.com/lenstronomy/lenstronomy/blob/f731c7c9fe2fd91ba6921184baaf151074a02ea5/lenstronomy/Sampling/Likelihoods/time_delay_likelihood.py#L68

I am working on a PR to solve this.

sibirrer:
@martin-millon the original use of this is intent to only have one True/False statement per point_source_model, either a source is time-variable and measurements exist, or the source is static. We either need to specify the use or indeed fix it to not cause confusions. Thank you for pointing this out and looking forward to the PR!

martin-millon:
The proposed solution should conserve the previous behavior if default arguments are provided. You can now provide a list of list of boolean in `time_delay_measurement_bool_list`. For example, for 2-point sources, imaged 4 times each (3 time-delays), you can provide `time_delay_measurement_bool_list=[[True, True, True], [True, False, False]]`.

martin-millon:
Fixed in #680

PR #680, opened 2024-12-03
Time-delay likelihood correction for multiple point sources

martin-millon:
Solve issue #678. 

Changes : 
- Allows to provide list of list of booleans for each point source and for each image to determine which time-delays are used
- Propagation of the z_lens_convention argument in multiplane 
- Introduce the color bar option in the convergence plot 
- Propagate the lens equation solver options to the point source plots


lenstronomy/Sampling/Likelihoods/time_delay_likelihood.py
```diff
@@ -23,7 +23,9 @@ def __init__(
         :param lens_model_class: instance of the LensModel() class
         :param point_source_class: instance of the PointSource() class, note: the first point source type is the one the
          time delays are imposed on
-        :param time_delay_measurement_bool_list: list of bool to indicate for which point source model a measurement is available
+        :param time_delay_measurement_bool_list: list of list of bool to indicate for which point source model a measurement is available.
+        This list must have the same length as time_delays_measured and time_delays_uncertainties.
```
sibirrer, line 27: for the second line, there should be a space inwards to be rendered properly

lenstronomy/Sampling/Likelihoods/time_delay_likelihood.py
```diff
@@ -49,7 +51,23 @@ def __init__(
...
+                for i in range(self._num_point_sources):
+                    time_delay_measurement_bool_list.append(
+                        [True] * len(time_delays_measured[i])
+                    )
+        else:
+            for i in range(self._num_point_sources):
+                if len(time_delay_measurement_bool_list[i]) != len(
+                    self._delays_measured[i]
+                ):
+                    raise ValueError(
```
sibirrer, line 67: this error does not seem to be backwards compatible. A single True per point source model should also be possible and then resulting in the delays being calculated and evaluated between all of the point sources

lenstronomy/Sampling/Likelihoods/time_delay_likelihood.py
```diff
@@ -76,9 +95,17 @@ def logL(self, kwargs_lens, kwargs_ps, kwargs_cosmo):
...
+                mask_full = np.concatenate(
+                    ([True], mask)
+                )  # add the first image to the mask
+                if len(delay_days) - 1 != len(self._delays_measured[i]):
+                    logL += -(10**15)
+                else:
+                    logL += self._logL_delays(
+                        delay_days[mask_full],
+                        self._delays_measured[i][mask],
+                        self._delays_errors[i][mask],
```
sibirrer, line 107: i am not sure this works when providing a n x n covariance matrix as the errors in the delays

sibirrer (changes requested):
Thank you @martin-millon ! I have some comments and questions whether the changes will work with error covariance matrices. I'd still like to have a single boolean per point source model be supported, if possible. 
There is also some text coverage dropping due to mostly small changes in the plotting routine.

martin-millon:
@sibirrer I am still working on this PR, especially the automated test, which needs to be updated. I'll let you know when it's ready and include your comments in the new version.

lenstronomy/Sampling/Likelihoods/time_delay_likelihood.py
martin-millon: This is now tested in L220 of test_time_delay_likelihood.py and it works

lenstronomy/Sampling/Likelihoods/time_delay_likelihood.py
martin-millon: It should be backward compatible now

lenstronomy/Sampling/Likelihoods/time_delay_likelihood.py
martin-millon: corrected

martin-millon:
@sibirrer I think this is now ready to be merged.

lenstronomy/Plots/lens_plot.py
```diff
@@ -161,14 +164,15 @@ def convergence_plot(
     :param vmax: matplotlib vmax
     :param cmap: matplotlib cmap
     :param kwargs: keyword arguments for matshow
+    :param with_color_bar: bool, if True, shows color bar
```
sibirrer, line 167: can you add docstrings for colorer_label and font_size?

sibirrer (changes requested):
Looks good. Just some minor docstrings to be added and I am waiting for the coverage report

martin-millon:
I don't get why these two lines of code do not get executed by codecov whereas the program enters the if statement...

PR #681, opened 2024-12-03
Fix typo in shapelets docstring

nataliehogg:
Tiny typo fix so the html equation renders correctly.


sibirrer (approved):
Thank you @nataliehogg !

PR #684, opened 2024-12-08
Use distance ratio betas instead of a and b ratios for sampling free parameters in multi-lens-plane

ajshajib:
This pull request includes the introduction of cosmology sampling and refactoring of distance calculations in multi-plane lensing scenarios.

* Added cosmology sampling functionality to the `image2source_mapping.py` by introducing the `_cosmology_sampling` attribute and updating the `__init__` method to handle cosmology sampling. (`lenstronomy/ImSim/image2source_mapping.py`) [[1]](diffhunk://#diff-47b1e62e722e6f93bff10e6f19f7e1aafac46d08881d733bf35f1c515d6c780cR43) [[2]](diffhunk://#diff-47b1e62e722e6f93bff10e6f19f7e1aafac46d08881d733bf35f1c515d6c780cR55-R57)

* Refactored the `MultiPlaneOrganizer` class to use `beta` parameters instead of `a` and `b` factors for distance calculations, and updated the relevant methods accordingly. (`lenstronomy/ImSim/multiplane_organizer.py`) [[1]](diffhunk://#diff-acb17c0080cd8f715ae22d72f2946c7b9da66dece76a1dc6bc85feef43a1bbabL16-R28) [[2]](diffhunk://#diff-acb17c0080cd8f715ae22d72f2946c7b9da66dece76a1dc6bc85feef43a1bbabR67-R78)

### DIFF
--- a/lenstronomy/ImSim/image2source_mapping.py
+++ b/lenstronomy/ImSim/image2source_mapping.py
@@ -3,6 +3,7 @@ import warnings
 from astropy.cosmology import *
 from lenstronomy.Cosmo.background import Background
 from lenstronomy.ImSim.multiplane_organizer import MultiPlaneOrganizer
+from lenstronomy.Util.cosmo_util import get_cosmology
 
 __all__ = ["Image2SourceMapping"]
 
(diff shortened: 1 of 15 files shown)


lenstronomy/LensModel/MultiPlane/multi_plane.py
```diff
@@ -53,11 +57,64 @@ def __init__(
...
         :param kwargs_multiplane_model: keyword arguments for the MultiPlaneDecoupled class, if specified
         :param distance_ratio_sampling: bool, if True, will use sampled
          distance ratios to update T_ij value in multi-lens plane computation.
+        :param cosmology_sampling: bool, if True, will use sampled cosmology
+        :param cosmology_model: str, name of the cosmology model to use for
         """
+        self.cosmology_sampling = cosmology_sampling
+        self.cosmology_model = cosmology_model
+        if self.cosmology_sampling:
+            supported_cosmologies = [
```
sibirrer, line 71: there are some things here that are duplicate/very similar in another class already. Is it possible to have an utility function for this?

lenstronomy/ImSim/image2source_mapping.py
```diff
@@ -1,4 +1,6 @@
 import numpy as np
+import warnings
+from astropy.cosmology import *
```
sibirrer, line 3: I don't really like imports with * as it makes it less transparent from where certain objects/variables come from

sibirrer (commented):
Thank you @ajshajib. I have two comments that might help to be addressed. Otherwise it is looking good. Thank you very much!

sibirrer (approved):

martin-millon:
@ajshajib is this meant only to sample the _expansion history_ in multiplane lensing or could this be extended to SinglePlane lens models and to the TimeDelayLikelihood ? In its current state, Ddt is sampled separately in the TimeDelayLikelihood and does not seem integrate the new `cosmology_sampling` option. Should I implement this option?

ajshajib:
@martin-millon, you are right that, currently, cosmology sampling is only implemented for the multi-plane case. But, yes, it would be great to connect it to the `TimeDelayLikelihood`. 🙏

martin-millon:
OK, I've implemented it on my fork. Once I've convinced myself that this is propagated correctly everywhere, I'll make a PR. 
It also impacts the `SourcePositionLikelihood`, `ImagePositionLikelihood`, ect ...

martin-millon:
@sibirrer do you think that this option `cosmology_sampling` should also exist for `SinglePlane` lenses, or is this going to break too many things?

sibirrer:
> @sibirrer do you think that this option `cosmology_sampling` should also exist for `SinglePlane` lenses, or is this going to break too many things?

@martin-millon I think it works, it should just default into not doing any unnecessary calculations when not being used (default)

PR #685, opened 2024-12-09
m=1 multipole (spherical/elliptical) + updated solution of m=3 multipole

Hadrien-Pgnt:
Implements solutions for the spherical m=1, elliptical m=1 and elliptical m=3 that have fully continuous deflection fields (old m=3 version had discontinuities).

### DIFF
--- a/lenstronomy/LensModel/Profiles/multipole.py
+++ b/lenstronomy/LensModel/Profiles/multipole.py
@@ -14,18 +14,19 @@ class Multipole(LensProfileBase):
     This uses the same definitions as Xu et al.(2013) in Appendix B3 https://arxiv.org/pdf/1307.4220.pdf, Equation B12
     Only the q=1 case (ie., spherical symmetry) makes this definition consistent with interpretation of multipoles as a deformation of the isophotes with an order m symmetry (eg., disky/boxy in the m=4 case).
 
-    m : int, multipole order, m>=2
+    m : int, multipole order, m>=1
     a_m : float, multipole strength
     phi_m : float, multipole orientation in radian
     """
 
-    param_names = ["m", "a_m", "phi_m", "center_x", "center_y"]
+    param_names = ["m", "a_m", "phi_m", "center_x", "center_y", "r_E"]
     lower_limit_default = {
-        "m": 2,
+        "m": 1,
         "a_m": 0,
         "phi_m": -np.pi,
         "center_x": -100,
         "center_y": -100,
+        "r_E": 0,
     }
     upper_limit_default = {
         "m": 100,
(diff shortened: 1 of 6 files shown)


dangilman:
@Hadrien-Pgnt Thanks for adding this! Regarding the tests, the main thing that needs to happen is for the test_numeric_lens_differentials to pass, and for the solutions to match spherical multipoles for q=1. You could probably simplify the test suite to just do those checks

lenstronomy/LensModel/Profiles/multipole.py
```diff
@@ -47,37 +49,49 @@ def function(self, x, y, m, a_m, phi_m, center_x=0, center_y=0):
...
         """
-        Deflection of a multipole contribution (for 1 component with m>=2)
+        Deflection of a multipole contribution (for 1 component with m>=1)
         This uses the same definitions as Xu et al.(2013) in Appendix B3 https://arxiv.org/pdf/1307.4220.pdf
         Equation B12
 
         :param x: x-coordinate to evaluate function
         :param y: y-coordinate to evaluate function
-        :param m: int, multipole order, m>=2
+        :param m: int, multipole order, m>=&
```
dangilman, line 73: possible typo in docstring?

lenstronomy/LensModel/Profiles/multipole.py
```diff
@@ -144,47 +177,58 @@ class EllipticalMultipole(LensProfileBase):
...
         r, phi = param_util.cart2polar(x, y, center_x=center_x, center_y=center_y)
+        r = np.maximum(r, 0.000001)
 
-        if np.abs(1 - q) < 1e-4:  # avoid numerical instability when q is too close to 1
-            f_ = r * a_m / (1 - m**2) * np.cos(m * (phi - phi_m))
+        if (
+            np.abs(1 - q**2) ** ((m + 1) / 2) < 1e-8
+        ):  # avoid numerical instability when q is too close to 1
+            if m == 1:
+                f_ = r * np.log(r / r_E) * a_m / 2 * np.cos(phi - phi_m)
```
dangilman, line 206: I think this line, the m=1 case of the q -> 1 special case, was missed by tests

dangilman (changes requested):
Thanks a lot @Hadrien-Pgnt. Everything looks great, one very minor comment about a line that gets missed by tests, but otherwise it looks good to go

dangilman (approved):

lenstronomy/LensModel/Profiles/multipole.py
Hadrien-Pgnt: I test that the elliptical solution with q approaching 1 is approximately the spherical solution.
This line is here (for all orders m) to avoid numerical instabilities due to terms (1-q)^(-n) when q is too close to 1, they are copies of the spherical multipole solution - I'll make that explicit in another PR and add a test to check that solutions are exactly the same if the "q too close to 1" condition is verified.

PR #686, opened 2024-12-11
Changes to q=1 case handling for elliptical multipoles

Hadrien-Pgnt:
Explicitely use spherical multipole class for elliptical multipoles when q is dangerously close to 1+ add a test of this in test_multipoles.py


dangilman:
Thank you for adding the additional checks, @Hadrien-Pgnt! I think the test coverage % is going down with this pull request only because there are fewer overall lines of code. I will go ahead and merge this

PR #687, opened 2024-12-16
Expand cosmological parameters sampling to SinglePlane models

martin-millon:
I've also fixed a bug at https://github.com/lenstronomy/lenstronomy/blob/be142b8e0bc11f55302a99a9fcc1145aeb554407/lenstronomy/LensModel/MultiPlane/multi_plane_base.py#L107 

where the new comoving distances were appended to the list at each call of this function.


lenstronomy/LensModel/lens_model.py
```diff
@@ -271,7 +281,14 @@ def fermat_potential(
...
     def arrival_time(
-        self, x_image, y_image, kwargs_lens, kappa_ext=0, x_source=None, y_source=None
+        self,
+        x_image,
+        y_image,
+        kwargs_lens,
+        kappa_ext=0,
+        x_source=None,
+        y_source=None,
+        kwargs_cosmo=None,
```
sibirrer, line 291: I am not so sure this is the best way doing it, since in multiplane, also the ray shooting and any other call is also cosmology dependent. Perhaps the sampling class is the only one that needs to get a new cosmo object and updates all the classes?

lenstronomy/Sampling/Likelihoods/position_likelihood.py
```diff
@@ -179,18 +182,32 @@ def astrometric_likelihood(kwargs_ps, kwargs_special, sigma):
...
+        if self._lensModel.type == "MultiPlane":  # only for multi-plane lens models
+            if (
+                self._lensModel.lens_model.cosmology_sampling is True
+                and kwargs_special is not None
+            ):
+                cosmo = get_astropy_cosmology(
+                    cosmology_model=self._lensModel.lens_model.cosmology_model,
+                    param_kwargs=kwargs_special,
+                )
+                self._lensModel.lens_model.set_background_cosmo(cosmo)
```
sibirrer, line 209: chances are that the self._lensModel instance already got updated (there should only be one instance that gets moved to different classes. I would prefer the setting of the cosmology in these classes to be done at a higher level and that these individual classes do not need to know them

sibirrer (commented):
Thank you @martin-millon ! I have a few comments and let @ajshajib review it

lenstronomy/Sampling/Likelihoods/position_likelihood.py
martin-millon: I agree, ideally, this should all be in the LensCosmo class, as for the Single plane case, but currently Multiplane model does not have such an attribute. It would need some serious refactoring of the LensCosmo class to support multiplane.

lenstronomy/LensModel/MultiPlane/multi_plane.py
```diff
@@ -68,9 +68,16 @@ def __init__(
         """
         self.cosmology_sampling = cosmology_sampling
         self.cosmology_model = cosmology_model
-        if self.cosmology_sampling:
+        if cosmo is None and cosmology_model == "FlatLambdaCDM":
```
ajshajib, line 71: As a failsafe, please also add the case of `if cosmology_model is None` here.

martin-millon:
@ajshajib This is now ready for review. I had to adapt this PR to Simon's latest update (#691 )

lenstronomy/LensModel/MultiPlane/multi_plane.py
martin-millon: Default value for cosmology is 'FlatLambdaCDM,' so I think this is fine, unless you explicitly pass None. Let me know if you think the default behavior is not optimal.

PR #690, opened 2024-12-19
Initialize T_ij_list and T_z_list in MultiPlaneBase class

ajshajib:
This fixes a bug that @martin-millon pointed out in #687; many thanks! Although @martin-millon included the bugfix in #687, I am pushing it here already so that the currently implemented multi-plane feature can be used by other users.


ajshajib:
@sibirrer, I am merging this PR, bypassing the review requirement, so that my collaborator can work with the bug-fixed version. This bugfix already came from Martin in #687, so I think that kind of acts as a soft "review", and this is a small fix. But, please feel free to revert if something is not right.

PR #691, opened 2024-12-19
Lens model cosmology refactoring

sibirrer:
This PR adds a definition in LensModel to update the cosmology instance. The idea is that the API should support all downstream cosmology definitions and serves as a single point to update the cosmology of the lens model configuration.

### DIFF
--- a/test/test_LensModel/test_lens_model.py
+++ b/test/test_LensModel/test_lens_model.py
@@ -409,8 +409,6 @@ class TestLensModel(object):
         dt_new = lens_model_new.arrival_time(1, 1, kwargs_lens=kwargs_lens)
         npt.assert_almost_equal(dt, dt_new, decimal=5)
 
-        # TODO: raise testing
-
 
 class TestRaise(unittest.TestCase):
     def test_raise(self):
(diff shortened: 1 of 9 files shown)


test/test_LensModel/test_lens_model.py
```diff
@@ -354,6 +354,62 @@ def test_change_source_redshift(self):
...
+        from astropy.cosmology import FlatwCDM
+
+        cosmo = FlatwCDM(H0=67, Om0=0.3, w0=-0.8)
+        cosmo_new = FlatwCDM(H0=73, Om0=0.3, w0=-1)
+
+        z_lens = 0.5
+        z_source_convention = 2
+        z_source_new = 1
+        kwargs_lens = [{"theta_E": 1, "center_x": 0, "center_y": 0}]
+        # multi-plane lens model
```
martin-millon, line 367: This seems to have only one plane. Can you add a test with two lens planes ?

martin-millon (approved):
This looks good, thanks @sibirrer ! I'll try to run my tests with this changes,

martin-millon:
are you planning to have the modifications of the Param class to allow the sampling of the cosmological parameter in this PR as well, or should we already merge this?

sibirrer:
I think we can merge this one @martin-millon. I will open a new PR for the parameter class

PR #696, opened 2025-01-09
a few small bug fixes

ahuang314:
1. Changes the way num_pix is calculated in the PSF class when generating gaussian kernels to more closely match that of scipy's definition
2. Fixed a bug in the MultiGaussianConvolution class where the deltapix argument should be equal to 1, since sigma_scaled is in units of pixels
3. In the Gaussian light model, explicitly declare the dtype when initializing an array


sibirrer (approved):
Thank you @ahuang314 !

PR #706, opened 2025-01-28
psf_error_map normalization

sibirrer:
normalizes the psf_error_map (currently defined as the variance at each pixel for the input PSF). Applies if the PSF input is not normalized

### DIFF
--- a/lenstronomy/ImSim/Numerics/point_source_rendering.py
+++ b/lenstronomy/ImSim/Numerics/point_source_rendering.py
@@ -73,8 +73,8 @@ class PointSourceRendering(object):
         :param dec_pos: image positions of point sources
         :param amp: amplitude of modeled point sources
         :param data: 2d numpy array of the data
-        :param fix_psf_variance_map: bool, if True, estimates the error based on the input
-            (modeled) amplitude, else uses the data to do so.
+        :param fix_psf_variance_map: bool, if True, estimates the error based on the
+            input (modeled) amplitude, else uses the data to do so.
         :return: 2d array of size of the image with error terms (sigma**2) expected from
             inaccuracies in the PSF modeling
         """
--- a/lenstronomy/ImSim/Numerics/point_source_rendering.py
+++ b/lenstronomy/ImSim/Numerics/point_source_rendering.py
@@ -66,7 +66,7 @@ class PointSourceRendering(object):
             )
         return self._kernel_supersampled_instance
 
-    def psf_variance_map(self, ra_pos, dec_pos, amp, data, fix_psf_error_map=False):
+    def psf_variance_map(self, ra_pos, dec_pos, amp, data, fix_psf_variance_map=False):
         """Variance of PSF error.
 
         :param ra_pos: image positions of point sources
--- a/lenstronomy/ImSim/Numerics/point_source_rendering.py
+++ b/lenstronomy/ImSim/Numerics/point_source_rendering.py
@@ -83,7 +83,7 @@ class PointSourceRendering(object):
         psf_variance_map = self._psf.psf_variance_map
         variance_map = np.zeros_like(data)
         for i in range(len(x_pos)):
-            if fix_psf_error_map is True:
+            if fix_psf_variance_map is True:
                 amp_estimated = amp
             else:
                 amp_estimated = kernel_util.estimate_amp(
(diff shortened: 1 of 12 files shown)


ajshajib:
Hi @sibirrer, while this is a small change and it looks fine to me, I wonder if we should consider renaming `psf_error_map` to `psf_variance_map` to clarify the variable's purpose, or alternatively expect the uncertainty being passed to that variable instead of the variance.

sibirrer:
> Hi @sibirrer, while this is a small change and it looks fine to me, I wonder if we should consider renaming `psf_error_map` to `psf_variance_map` to clarify the variable's purpose, or alternatively expect the uncertainty being passed to that variable instead of the variance.

That's a good point. Both are not quite backwards compatible (one requiring a renaming of the variable, the other change the input). Which one would be your preferred way?

ajshajib:
I think renaming to `psf_variance_map` would be better because it will cause all backward-incompatible code to throw an error, forcing the necessary change in the code. The alternative may make this change go unnoticed from the user's point of view.

sibirrer:
ok @ajshajib I changed the naming conventions throughout lenstronomy. For the user, the only change is form 'psf_error_map' to 'psf_variance_map' in kwargs_psf

lenstronomy/ImSim/Numerics/point_source_rendering.py
```diff
@@ -66,29 +66,30 @@ def _kernel_supersampled(self):
             )
         return self._kernel_supersampled_instance
 
-    def psf_error_map(self, ra_pos, dec_pos, amp, data, fix_psf_error_map=False):
-        """
+    def psf_variance_map(self, ra_pos, dec_pos, amp, data, fix_psf_error_map=False):
```
ajshajib, line 69:
```suggestion
    def psf_variance_map(self, ra_pos, dec_pos, amp, data, fix_psf_variance_map=False):
```

lenstronomy/ImSim/Numerics/point_source_rendering.py
```diff
@@ -66,29 +66,30 @@ def _kernel_supersampled(self):
...
+        """Variance of PSF error.
 
         :param ra_pos: image positions of point sources
         :param dec_pos: image positions of point sources
         :param amp: amplitude of modeled point sources
         :param data: 2d numpy array of the data
-        :param fix_psf_error_map: bool, if True, estimates the error based on the input (modeled) amplitude, else uses
-         the data to do so.
-        :return: 2d array of size of the image with error terms (sigma**2) expected from inaccuracies in the PSF modeling
+        :param fix_psf_error_map: bool, if True, estimates the error based on the input
```
ajshajib, line 76:
```suggestion
        :param fix_psf_variance_map: bool, if True, estimates the error based on the input
```

lenstronomy/ImSim/Numerics/point_source_rendering.py
```diff
@@ -66,29 +66,30 @@ def _kernel_supersampled(self):
...
+            inaccuracies in the PSF modeling
         """
         x_pos, y_pos = self._pixel_grid.map_coord2pix(ra_pos, dec_pos)
         psf_kernel = self._psf.kernel_point_source
-        psf_error_map = self._psf.psf_error_map
-        error_map = np.zeros_like(data)
+        psf_variance_map = self._psf.psf_variance_map
+        variance_map = np.zeros_like(data)
         for i in range(len(x_pos)):
             if fix_psf_error_map is True:
```
ajshajib, line 86:
```suggestion
            if fix_psf_variance_map is True:
```

lenstronomy/Workflow/psf_fitting.py
```diff
@@ -90,7 +90,7 @@ def update_iterative(
         :param kwargs_psf: keyword arguments to construct the PSF() class
         :param kwargs_params: keyword arguments of the parameters of the model components (e.g. 'kwargs_lens' etc)
         :param num_iter: number of iterations in the PSF fitting and image fitting process
-        :param keep_psf_error_map: boolean, if True keeps previous psf_error_map
+        :param keep_psf_error_map: boolean, if True keeps previous psf_variance_map
```
ajshajib, line 93:
```suggestion
        :param keep_psf_variance_map: boolean, if True keeps previous psf_variance_map
```

ajshajib (changes requested):
Thanks a lot, @sibirrer. I have added some suggestions to also update the associated variable names to `keep_psf_variance_map` and `fix_psf_variance_map`, but there may be other occurrences where similar changes may be needed, but didn't come up within the view of the review page on GitHub for me to flag them.

ajshajib (approved):

PR #707, opened 2025-01-31
enable param class to be used in jaxtronomy

ahuang314:
I add one line of code to make the Param class compatible with jaxtronomy for now.
If in the future jaxtronomy requires significant changes to the Param class, then jaxtronomy will have its own version of the class, but for now we use the lenstronomy version.

### DIFF
--- a/lenstronomy/Sampling/parameters.py
+++ b/lenstronomy/Sampling/parameters.py
@@ -530,7 +530,7 @@ class Param(object):
         """
         return self._linear_solver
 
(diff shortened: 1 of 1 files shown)


sibirrer (commented):
If this makes the entire Param() class compatible with JAX, that is great.
If there are more changes necessary, perhaps make one PR with one of them such that only one new version of lenstronomy is required.

sibirrer (approved):

PR #708, opened 2025-02-02
Bug fix in MultiPatchReconstruction.image_joint()

ahuang314:
This fixes a bug that I introduced in PR #702 where select_kwargs() is executed twice, once in `model_band.kwargs_model` and once in `SingleBandMultiModel.image()` resulting in a list index error.
I've also made the test function more robust to be able to catch this sort of error in the future.

### DIFF
--- a/lenstronomy/Analysis/multi_patch_reconstruction.py
+++ b/lenstronomy/Analysis/multi_patch_reconstruction.py
@@ -127,9 +127,7 @@ class MultiPatchReconstruction(MultiBandImageReconstruction):
         for model_band in self.model_band_list:
             if model_band is not None:
                 image_model = model_band.image_model_class
-                model = image_model.image(
-                    **self._kwargs_params_no_tracersource
-                ) 
+                model = image_model.image(**self._kwargs_params_no_tracersource)
                 data_class_i = image_model.Data
                 # evaluate pixel of zero point with the base coordinate system
                 ra0, dec0 = data_class_i.radec_at_xy_0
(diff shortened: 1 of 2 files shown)


lenstronomy/Analysis/multi_patch_reconstruction.py
```diff
@@ -126,10 +127,7 @@ def image_joint(self):
         for model_band in self.model_band_list:
             if model_band is not None:
                 image_model = model_band.image_model_class
-                kwargs_params = model_band.kwargs_model
-                model = image_model.image(
-                    **kwargs_params
-                )  # TODO: avoid using private definitions uses sub-set of the model parameters
+                model = image_model.image(**self._kwargs_params_no_tracersource)
```
sibirrer, line 130: I am confused this did not work previously. The image_model class should be a simple ImageModel() class without the SingleBandMultiImage() class to down-select the arguments. Hence the image_model.image() call should operate on the down-selected parameters. If this has changed, then there is a broader issue with the ImageReconstruction() class return of model_band.image_model.class

sibirrer (commented):
I do have a comment as I think this might solve this bug but points to a broader issue of the ImageReconstruction class

lenstronomy/Analysis/multi_patch_reconstruction.py
ahuang314: In the Modelband class, instances of SingleBandMultiModel are created, not ImageModel classes.
Perhaps the Modelband class should be changed so that ImageModel or ImageLinearFit is created depending on whether linear solve is true or not?

sibirrer (approved):

PR #711, opened 2025-02-08
fix bug in TNFWC lens profile class

dangilman:
I fix a bug in the TNFWC profile class when the parameter r_trunc is greater than Rs. A misplaced abs() operator would make the profile have negative convergence. This is now fixed, and a test covering the case r_trunc < Rs was added


sibirrer (approved):
Thank you @dangilman !

PR #714, opened 2025-02-26
fixed a few small bugs

ahuang314:
1. Fixed lower limit default and upper limit default dicts for gaussian profiles
2. Fixed a bug in LightParam where the code would crash if n_max was not in kwargs even if it was in kwargs_fixed
3. Fixed a bug in Gaussian deflector profile where the numerical integral returned nan instead of 0 when r = 0
4. Deleted duplicate lines of code in the Param class
5. Changed param name "A" to "a" in CSE profile to stay consistent with function keyword arguments

### DIFF
--- a/lenstronomy/LightModel/light_param.py
+++ b/lenstronomy/LightModel/light_param.py
@@ -138,7 +138,7 @@ class LightParam(object):
                         model in ["SHAPELETS", "SHAPELETS_POLAR", "SHAPELETS_POLAR_EXP"]
                         and name == "amp"
                     ):
-                        n_max = kwargs_fixed.get("n_max", kwargs.get(["n_max"]))
+                        n_max = kwargs_fixed.get("n_max", kwargs.get("n_max"))
                         if model in ["SHAPELETS_POLAR_EXP"]:
                             num_param = int((n_max + 1) ** 2)
                         else:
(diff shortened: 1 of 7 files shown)


sibirrer:
@ahuang314 there is an error in the tests. I don't think you need the bracket around "n_max" in the inner statement.

lenstronomy/LightModel/light_param.py
```diff
@@ -138,7 +138,7 @@ def set_params(self, kwargs_list):
                         model in ["SHAPELETS", "SHAPELETS_POLAR", "SHAPELETS_POLAR_EXP"]
                         and name == "amp"
                     ):
-                        n_max = kwargs_fixed.get("n_max", kwargs["n_max"])
+                        n_max = kwargs_fixed.get("n_max", kwargs.get("n_max"))
```
sibirrer, line 141: @ahuang314 there is an error in the tests. I don't think you need the bracket around "n_max" in the inner statement.

ahuang314: Fixed now

sibirrer (approved):
Thank you very much @ahuang314

sibirrer (changes requested):
actually, can you please add a test case for r==0 for the Gaussian profile? The newly added line is not tested

sibirrer (approved):

PR #717, opened 2025-03-02
Remove kwargs_synthesis, kwargs_interp, numerical_alpha_class from LensModel and ProfileListBase

ahuang314:
This PR deprecates several LensModel arguments and instead introduces a new argument profile_kwargs_list, which allows the user to change the initialization settings of any deflector profile. This also means we don't need to include more and more separate LensModel arguments every time we add a profile that requires kwargs at initialization.

I plan to introduce the same feature to LightModel in a future PR.


lenstronomy/LensModel/profile_list_base.py
```diff
@@ -110,87 +110,47 @@ class ProfileListBase(object):
...
-                lensmodel_class = lens_class(
-                    lens_type,
-                    custom_class=custom_class,
-                    z_lens=lens_redshift_list[i],
-                    z_source=z_source_convention,
-                    kwargs_interp=kwargs_interp,
-                    kwargs_synthesis=kwargs_synthesis,
-                )
-            else:
-                if lens_type not in imported_classes.keys():
```
sibirrer, line 184: This feature of not creating a new class instance and using instead the same class instance for lens models that are not dependent on initialization is removed in your PR

lenstronomy/LensModel/LightConeSim/light_cone.py
```diff
@@ -58,7 +58,7 @@ def cone_instance(self, z_source, cosmo, multi_plane=True, kwargs_interp=None):
             z_source_convention=z_source,
             cosmo=cosmo,
             z_source=z_source,
-            kwargs_interp=kwargs_interp,
+            profile_kwargs_list=[kwargs_interp] * len(self._mass_map_list),
```
sibirrer, line 61: I am not sure this is the intent here. The idea is that there are different convergence maps for the different redshift and the input should support this

lenstronomy/LensModel/MultiPlane/decoupled_multi_plane.py
```diff
@@ -50,13 +48,10 @@ def __init__(
...
             deflection angles of the lens models. If None, 'z_source' is used.
         :param cosmo: instance of astropy.cosmology
-        :param numerical_alpha_class: an instance of a custom class for use in
-            NumericalAlpha() lens model (see documentation in Profiles/numerical_alpha)
-        :param kwargs_interp: interpolation keyword arguments specifying the numerics.
-            See description in the Interpolate() class. Only applicable for 'INTERPOL'
-            and 'INTERPOL_SCALED' models.
-        :param kwargs_synthesis: keyword arguments for the 'SYNTHESIS' lens model, if
-            applicable
+        :param profile_kwargs_list: list of dicts, keyword arguments used to initialize
```
sibirrer, line 51: while the input is correctly supported, this class deliberately intends to use NumericalAlpha() classes and perhaps the documentation of the class should reflect that

sibirrer (changes requested):
Thank you very much @ahuang314 ! I have two requests and then let @dangilman review it further as this changes will mostly affect the use and documentation of the code he uses

lenstronomy/LensModel/MultiPlane/decoupled_multi_plane.py
dangilman: @ahuang314 Actually this class doesn't use the NumericalAlpha classes, I think these changes are fine

lenstronomy/LensModel/profile_list_base.py
dangilman: @ahuang314 Sometimes with substructure in the lens model you'll have 10s of thousands of lens profile classes, this feature was implemented to avoid instantiating the same class over and over again

dangilman (commented):
@sibirrer @ahuang314 thanks for looping me in. The MultiPlaneDecoupled class actually does not make explicit use of the NumericalAlpha class, so the changes won't directly impact the methods

lenstronomy/LensModel/profile_list_base.py
ahuang314: Gotcha. I took it out because it's a bit tedious keeping track of all the profiles which have `set_static()` methods and/or different initialization settings. But I'll re-implement it; maybe there is a cleaner way to do this.

sibirrer (approved):
Looks good to me. I leave it up to @dangilman

dangilman (approved):

PR #719, opened 2025-03-06
some docstring changes, small bug fix in LensedPosition

ahuang314:
Small bug fix in the amplitude calculations in LensedPosition when self.k_list is not None

point_source_frame_list is a bit confusing and the docstring is not very clear; we should address this at some point

### DIFF
--- a/lenstronomy/PointSource/Types/base_ps.py
+++ b/lenstronomy/PointSource/Types/base_ps.py
@@ -20,19 +20,23 @@ class PSBase(object):
 
         :param lens_model: instance of the LensModel() class
         :param fixed_magnification: bool. If True, magnification
-         ratio of point sources is fixed to the one given by the lens model
+            ratio of point sources is fixed to the one given by the lens model
         :param additional_images: bool. If True, search for additional images of the same source is conducted.
         :param index_lens_model_list: list (length of different patches/bands) of integer lists, e.g., [[0, 1], [2, 3]];
-         evaluating a subset of the lens models per individual bands. If this keyword is set, the image positions need
-         to have a specified band/frame assigned to it
-        :param point_source_frame_list: list of lists mirroring the structure of the image positions.
-         Integers correspond to the i'th list entry of index_lens_model_list indicating in which frame/band the image is
-         appearing
+            evaluating a subset of the lens models per individual bands. If this keyword is set, the image positions need
+            to have a specified band/frame assigned to it
+        :param point_source_frame_list: list of ints, e.g. we can assign the i-th image to the second band
+            with point_source_frame_list[i] = 2
         :param redshift: redshift of the source, only required for multiple source redshifts
         :type redshift: None or float
         """
         self._redshift = redshift
         self._lens_model = lens_model
+
+        # k_list is a list of lists which combines index_lens_model_list with point_source_frame_list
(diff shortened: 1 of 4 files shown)


lenstronomy/PointSource/Types/lensed_position.py
```diff
@@ -130,7 +130,7 @@ def image_amplitude(
                 for i in range(len(ra_image)):
                     mag.append(
                         self._lens_model.magnification(
-                            ra_image, dec_image, kwargs_lens, k=self.k_list[i]
+                            ra_image[i], dec_image[i], kwargs_lens, k=self.k_list[i]
```
sibirrer, line 133: I see this was a bug. Is the format that is returned the same as for k_list = None (i.e. structure of the list), or do we have to re-map to a list of arrays?

lenstronomy/PointSource/Types/base_ps.py
```diff
@@ -20,14 +20,14 @@ def __init__(
...
         :param additional_images: bool. If True, search for additional images of the same source is conducted.
         :param index_lens_model_list: list (length of different patches/bands) of integer lists, e.g., [[0, 1], [2, 3]];
-         evaluating a subset of the lens models per individual bands. If this keyword is set, the image positions need
-         to have a specified band/frame assigned to it
-        :param point_source_frame_list: list of lists mirroring the structure of the image positions.
-         Integers correspond to the i'th list entry of index_lens_model_list indicating in which frame/band the image is
-         appearing
+            evaluating a subset of the lens models per individual bands. If this keyword is set, the image positions need
+            to have a specified band/frame assigned to it
+        :param point_source_frame_list: list of ints mirroring the structure of the image positions.
```
sibirrer, line 28: it is a list of lists of ints (complicated)
How about

:param point_source_frame_list: list of lists mirroring the structure of the image positions.
         Integers correspond to the i'th list entry of index_lens_model_list indicating in which frame/band the image is
         appearing, e.g., if you have one "LENSED_POSITION" model with four point sources appearing in four different
         frames, then point_source_frame_list = [[0, 1, 2, 3]]. If you have four types of "UNLENSED" point source models
         appearing in four different frames, then point_source_frame_list = [[0], [1], [2], [3]]

sibirrer (changes requested):
Thank you! I have some comments

lenstronomy/PointSource/Types/base_ps.py
ahuang314: Ah, this is in PSBase though. In PointSource class, this is a list of list of ints, but when it gets passed into PSBase, it would just be a list of ints

ahuang314: Also, point_source_frame_list seems to be completely unused in UNLENSED so I'm not sure that part is relevant.

lenstronomy/PointSource/Types/lensed_position.py
ahuang314: The output format in both cases will be a 1 dimensional list/array containing the mags of each image

ahuang314:
@sibirrer OK I updated the docstring again, it should be extremely clear now

sibirrer (approved):
Looks great, thank you very much @ahuang314 !

PR #721, opened 2025-03-07
fixed more bugs

ahuang314:
1. Fixed a bug in PointSource where point_source_list(with_amp=False) could raise an error when trying to apply np.ones_like(ra_list) in situations where the point sources have different number of images. Also added a test for this scenario
2. Fixed a bug in image_util.add_layer2image_int where a rectangular kernel size resulted in a shape error. Also added a test for this scenario


sibirrer (approved):
Looks great! I doubt the case that with_amp has a different number of images than ra_list, but better save then sorry :)

PR #734, opened 2025-03-18
fixed a bug in PositionLikelihood source_position_likelihood check

ahuang314:
In the initialization of LikelihoodModule, the argument source_position_tolerance has a default value of 0.001. This means that in PositionLikelihood, source_position_likelihood will be calculated even if source_position_likelihood is False (unless the user overwrites the default 0.001).

Additionally, source_position_tolerance has a default value of None in PositionLikelihood. We should synchronize all of the default values to avoid confusion. @sibirrer should we set all of the defaults to None or 0.001?


sibirrer (approved):
Provided that the tests pass, I think this is the best choice. Only when the user provides a tolerance level, this gets added. It should not really affect past occurrences and can speed things a bit up for people not requiring the source_position_likelihood to be evaluated

PR #735, opened 2025-03-18
add no_arrow option for decomposition_plot

dangilman:


lenstronomy/Plots/model_band_plot.py
```diff
@@ -1090,6 +1091,7 @@ def decomposition_plot(
         :param point_source_add:
         :param source_add:
         :param lens_light_add:
+        :param no_arrow:
```
sibirrer, line 1094: I know this was not well documented beforehand, but could you add documentation as we go along? Thank you!

PR #742, opened 2025-03-24
updates setup.cfg

ahuang314:
setuptools got updated ~5 hours ago


sibirrer (approved):
Ha, this things happen unfortunately too often and sometimes it's just bad luck which PR gets the bad press from it :) Thank you very much!

PR #747, opened 2025-04-22
Implement ability to use jaxtronomy profiles in lenstronomy

ahuang314:
This PR allows the user to import deflector profiles from jaxtronomy instead of lenstronomy by setting the `use_jax` flag to `True` during the initialization of the `LensModel` class. This is useful in situations where it is not possible to use the full jaxtronomy `LensModel` class, for example when using PyHalo, the `LensModel` class will have tens of thousands of deflector profiles for dark matter substructure, leading to infinite compile time in jaxtronomy. This is due to how the JAX compiler handles `for` loops, leading to issues with the `for` loops in `SinglePlane` or `MultiPlaneBase` iterating through every profile.

This is hopefully a temporary solution until we figure out a clean way of doing it in jaxtronomy.

### DIFF
--- a/lenstronomy/LensModel/profile_list_base.py
+++ b/lenstronomy/LensModel/profile_list_base.py
@@ -174,7 +174,7 @@ class ProfileListBase(object):
             lens_redshift_list = [None] * len(lens_model_list)
         if profile_kwargs_list is None:
             profile_kwargs_list = [{} for _ in range(len(lens_model_list))]
-        
+
         # use_jax can be either a bool or list of bools to select specific models to be imported from jaxtronomy
         # If it's a bool, convert to list of bools
         if isinstance(use_jax, bool):
--- a/lenstronomy/LensModel/profile_list_base.py
+++ b/lenstronomy/LensModel/profile_list_base.py
@@ -170,6 +170,7 @@ class ProfileListBase(object):
         self._param_name_list = name_list
 
     def _load_model_instances(
+        self,
         lens_model_list,
         profile_kwargs_list=None,
         lens_redshift_list=None,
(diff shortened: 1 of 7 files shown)


lenstronomy/LensModel/profile_list_base.py
```diff
@@ -164,30 +168,42 @@ def _load_model_instances(
         profile_kwargs_list=None,
         lens_redshift_list=None,
         z_source_convention=None,
+        use_jax=False,
     ):
         if lens_redshift_list is None:
             lens_redshift_list = [None] * len(lens_model_list)
         if profile_kwargs_list is None:
             profile_kwargs_list = [{} for _ in range(len(lens_model_list))]
+        if use_jax is True or use_jax is False:
```
sibirrer, line 177: I don't understand this bool statement. Isn't this always true?

lenstronomy/LensModel/profile_list_base.py
```diff
@@ -164,30 +168,42 @@ def _load_model_instances(
...
+            use_jax = [use_jax] * len(lens_model_list)
+        if True in use_jax:
```
sibirrer, line 179: that's also an odd bool statement. I would state: if use_jax is True

lenstronomy/LensModel/profile_list_base.py
```diff
@@ -164,30 +168,42 @@ def _load_model_instances(
...
+            )
+
         func_list = []
         imported_classes = []
         imported_profile_kwargs = []
         for i, lens_type in enumerate(lens_model_list):
             if lens_type in ["NFW_MC", "NFW_MC_ELLIPSE_POTENTIAL"]:
                 profile_kwargs_list[i]["z_lens"] = lens_redshift_list[i]
                 profile_kwargs_list[i]["z_source"] = z_source_convention
+            if use_jax[i] is True:
```
sibirrer, line 191: this can break if use_jax is a single bool value. I would internally make use_jax to be a list of length of the lens models first such that you can uniformly use it within the class

sibirrer (changes requested):
Thank you very much @ahuang314 ! I have a few hopefully minor requests as I don't see clearly how the boolean operators were applied

sibirrer (approved):
Thank you @ahuang314 !

PR #751, opened 2025-04-27
add non-linear solver option keyword arguement

dangilman:
I add a keyword argument to the optimize method of the QuadOptimizer class to allow different optimization routines from scipy.minimize


sibirrer (approved):
Looks good to me. Feel free to merge @dangilman

PR #752, opened 2025-04-30
when use_jax flag is true, convert jax arrays back into np arrays

ahuang314:
This avoids errors in situations where specifically numpy arrays are needed. This does not have a noticeable impact on performance.

I also added the use_jax flag to the MultiPlaneDecoupled class.


sibirrer (approved):
Thank you @ahuang314 !

PR #753, opened 2025-05-01
bug fix in HERNQUIT_ELLIPSE_CSE normalization

sibirrer:
the normalization between HERNQUIST and HERNQUIST_ELLIPSE_CSE were not done correctly. Tests were too simplistic


sibirrer:
I merge it as it is a relatively urgent task for having SLSim run properly with the halo model

PR #760, opened 2025-05-13
add CUSTOM solver_type option for Solver4Point class

dangilman:
I add the possibility to specify solver_type = 'CUSTOM' with a user-specified class to handle parameters during the application of the solver in the Solver4Point class. The test function gives an example of how to use this methodology to enforce a fixed axis ratio q while still sampling in the e1/e2 basis with an elliptical power law profile.

### DIFF
--- a/lenstronomy/LensModel/Solver/solver.py
+++ b/lenstronomy/LensModel/Solver/solver.py
@@ -9,18 +9,20 @@ class Solver(object):
     """Joint solve class to manage with type of solver to be executed and checks whether
     the requirements are fulfilled."""
 
-    def __init__(self, solver_type, lensModel, num_images):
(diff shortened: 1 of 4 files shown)


dangilman:
@sibirrer this was the least invasive way I could think of to add flexibility to the Solver4Point method -- I wanted to avoid adding another if/else clause inside the existing class. I think it will quickly get out of hand if people eventually start using other lens profiles more regularly with the solver option. Another option would be to simply replace the entire class. That is, when the solver_type='CUSTOM' then user has to provide an entire Solver4Point-like class with the only requirement being that it have a "constraint_lensmodel" method

sibirrer (approved):
Hi @dangilman, thank you very much! I like it this way to make it more generic, and in the future we can make all the current options in such pre-defined classes

PR #761, opened 2025-05-16
callable method for optimization routine

dangilman:
Here I add the option to pass a callable function as the "method" keyword argument in the Optimizer class


sibirrer (approved):
Thank you @dangilman !

PR #762, opened 2025-05-17
KinematicsAPI compatible with multiple IFU observations

sibirrer:
This PR is needed to allow for multiple IFU data sets (such as JWST and KCWI) to be processed at the same time. Simplified options for direct convolution when IFU data sets are provided to take away options from the users.


ajshajib (approved):
There is one potentially leftover debug print statement that Copilot caught, otherwise it looks good to me.

ajshajib:
Thanks a lot, @sibirrer! Merging this now.

PR #766, opened 2025-06-07
various small changes for convenience/consistency

ahuang314:
1. the psf_class argument is required in the initialization of the ImageModel class, so it should also be required in the initialization of the ImageLinearFit class
2. Since kwargs_special is optional in ImageLinearFit.point_source_linear_response_set(), I have added a default value of None to this argument
3. Improved some docstrings in point source classes
4. In PointSource.check_positive_flux(), convert kwargs['amp'] to an array if it is a list so that it does not crash when comparing to an int


sibirrer (approved):
Looks great, thank you very much!

PR #772, opened 2025-06-14
add feature to do joint linear inversion for all parameters except sk…

amn3142:
add feature to do joint linear inversion for all parameters except sky background which can vary from exposure to exposure not compatible with any other features of lenstronomy e.g. plotting ect.

This is useful for multi exposure ground based data, where we expect the sky background to vary differently relative to the image brightness from exposure to exposure, and for situations where we don't have a good way of subtracting the background a-priori.

The linear inversion is used with multiple exposures of integral field spectroscopy data to calculate the flux as a function of wavelength for quasar images over many wavelengths. The relative brightness of the quasars is assumed to be constant, while the sky throughput and the sky background are assumed to possibly vary.

When testing I verified that the correct value of simulated sky backgrounds were approximately recovered on mock data.

### DIFF
--- a/lenstronomy/ImSim/MultiBand/joint_linear_vary_background.py
+++ b/lenstronomy/ImSim/MultiBand/joint_linear_vary_background.py
@@ -1,0 +1,271 @@
+from lenstronomy.ImSim.MultiBand.multi_linear import MultiLinear
+import lenstronomy.ImSim.de_lens as de_lens
+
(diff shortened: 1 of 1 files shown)


sibirrer:
Hi @amn3142, thank you for the PR. Can you write test functions for this class? Can be simple with two bands testing the expected behavior. Thank you!

PR #773, opened 2025-07-05
Add pixelated source reconstruction functions

nanz6:
This Pull Request introduces a new module for pixelated source reconstruction, enhancing lenstronomy's capabilities for modeling extended sources without assuming a parametric light profile. This feature is particularly useful for reconstructing complex source morphologies and complements existing parametric source modeling.

Key Features Introduced:

1. pixelated_source_reconstruction.py:
Implements the core logic for reconstructing a source on a pixelated grid in the source plane.
Handles mapping from image plane to source plane.
Provides the framework for pixelated source reconstruction. 
Generates the M matrix (source-pixel to source-pixel) and b vector (source-pixel to data image) for the linear source reconstruction problem.
Supports both diagonal covariance matrix (spatially uncorrelated noise with uniform RMS across the image) and the natural weighting interferometric noise covariance matrix.

2. regularization_matrix_pixel.py:
Generates various regularization matrices for pixelated source reconstruction.
Includes "zeroth-order", "gradient", and  "curvature" regularization schemes.

3. solve_regularization_strength.py:
Provides tools for determining the optimal regularization strength (e.g., the lambda parameter) based on methods maximizing the log (Bayes Evidence).

Test Files Included:

test/test_SourceReconstruction/test_pixelated_source_reconstruction.py
test/test_SourceReconstruction/test_regularization_matrix_pixel.py
test/test_SourceReconstruction/test_solve_regularization_strength.py

These tests ensure the correctness of the newly added functions.


lenstronomy/SourceReconstruction/pixelated_source_reconstruction.py
```diff
@@ -0,0 +1,510 @@
...
+
+from lenstronomy.LensModel.lens_model import LensModel
+from lenstronomy.Data.imaging_data import ImageData
+import lenstronomy.Util.util as util
+
+__all__ = ["PixelatedSourceReconstruction"]
+
+
+class PixelatedSourceReconstruction(object):
+    """This class defines useful functions for pixelated source plane reconstruction in
```
sibirrer, line 12: As far as I understand, the class initiation requires a specific lens model with known parameters. So this class has to be initialized at every step of the likelihood sampling? Would it be better to do the ray-tracing on the fly instead as having it as a self variable?

lenstronomy/SourceReconstruction/pixelated_source_reconstruction.py
```diff
@@ -0,0 +1,510 @@
...
+    gravitational lensing.
+
+    It handles initialization of data, lens model, and PSF kernel, and provides methods
+    for generating the M and b matrices for source reconstruction based on different
```
sibirrer, line 16: can you specify (or give references to) what M and b matrices are?

lenstronomy/SourceReconstruction/pixelated_source_reconstruction.py
```diff
@@ -0,0 +1,510 @@
...
+                or shape_kernel_cut[check_dim] < 2 * self._numPix - 1
+            ):
+                raise ValueError(
+                    "PSF kernel size must be odd and at least (2 * numPix - 1) "
+                    "in each dimension for proper convolution."
+                )
+        self._kernel = kernel
+
+        # (RA, DEC) coordinate at pixel (0,0) of the image plane in angular units
+        self._minx = kwargs_data["ra_at_xy_0"]
```
sibirrer, line 62: naming of the variable might be a bit misleading as in general this is not the minimum along the x-axis

lenstronomy/SourceReconstruction/pixelated_source_reconstruction.py
```diff
@@ -0,0 +1,510 @@
...
+        self._miny = kwargs_data["dec_at_xy_0"]
+
+        # Calculate lensed source plane coordinates for each image pixel
+        x_grid_temp, y_grid_temp = data_class.pixel_coordinates
+        self._x_grid = util.image2array(x_grid_temp)
+        self._y_grid = util.image2array(y_grid_temp)
+        self._beta_x_grid, self._beta_y_grid = lens_model_class.ray_shooting(
```
sibirrer, line 69: do you need this to be a self variable or can you execute these lines on the fly?

lenstronomy/SourceReconstruction/pixelated_source_reconstruction.py
```diff
@@ -0,0 +1,510 @@
...
+            self._x_grid, self._y_grid, kwargs=kwargs_lens
+        )
+
+        self._primary_beam = kwargs_data.get(
```
sibirrer, line 73: I suggest you to access these quantities within the ImageData() class directly instead of through the dictionary (more stable)

lenstronomy/SourceReconstruction/pixelated_source_reconstruction.py
```diff
@@ -0,0 +1,510 @@
...
+            "antenna_primary_beam", None
+        )  # Optional primary beam map
+        self._logL_method = kwargs_data.get(
```
sibirrer, line 76: I suggest you to access these quantities within the ImageData() class directly instead of through the dictionary (more stable)

lenstronomy/SourceReconstruction/pixelated_source_reconstruction.py
```diff
@@ -0,0 +1,510 @@
...
+        data_class = ImageData(**kwargs_data)
+
+        self._numPix = len(
+            kwargs_data["image_data"]
+        )  # Number of pixels along one dimension of the image
+        self._deltaPix = kwargs_data["transform_pix2angle"][
+            0, 0
+        ]  # Angular size of a single pixel
+        self._image_data = kwargs_data["image_data"]  # Observed image data
+        self._noise_rms = kwargs_data["background_rms"]  # RMS noise of the background
```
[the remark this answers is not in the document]
sibirrer: also note, there are also different ways to define the background_rms. When using the ImageData() method, these are automatically supported

lenstronomy/SourceReconstruction/pixelated_source_reconstruction.py
```diff
@@ -0,0 +1,510 @@
...
+        self._verbose = verbose
+
+        if self._logL_method not in ["diagonal", "interferometry_natwt"]:
+            raise ValueError(
+                "likelihood_method '%s' not supported! It can only be 'diagonal' or 'interferometry_natwt'."
+                % self._logL_method
+            )
+
+    def generate_M_b(self, x_min, x_max, y_min, y_max):
+        """Generates the M and b matrices for source reconstruction based on the
```
sibirrer, line 88: shouldn't the source grid density also be a free parameter? I think it's important to have the source grid scale be different (smaller) than the image grid pixel. One might even want to consider fitting the source pixel scale as a free parameter, as it is related to the mass sheet degeneracy

sibirrer (changes requested):
Thank you very much @nanz6 for the PR!
I have some initial questions and suggestions, primarily to potentially also support other features within lenstronomy. Perhaps the class can be similarly structured as the Numerics() class: https://github.com/lenstronomy/lenstronomy/blob/main/lenstronomy/ImSim/Numerics/numerics.py

nanz6:
Hi @sibirrer,

I have addressed all the initial questions and suggestions you raised, with the exception of the point regarding the initialization of PixelatedSourceReconstruction class at every step of likelihood sampling.

To clarify the current design:

The PixelatedSourceReconstruction class now explicitly takes (data_class, psf_class, lens_model_class, source_pixel_grid_class, kwargs_lens) as input during its initialization. The source_pixel_grid_class is a PixelGrid instance, allowing users to fully customize their source grid for a specific reconstruction. Here, the source grid is restricted to be non-rotational and to have uniform pixel width along both the x and y axes.

For the meaning of "M and b matrices", I have made a placeholder for my paper.

Regarding the question "So this class has to be initialized at every step of the likelihood sampling?", currently this module is designed for a one-time source reconstruction.

Please let me know if there are more suggestions on this module. Thanks!

nanz6:
And there is one another issue that I have not realized when making this code. 
When we define the source grid which has a different pixel size of the image, the flux should be multiplied by a factor to conserve the total flux (flux * pixel_width**2). I need to solve this issue.

lenstronomy/SourceReconstruction/pixelated_source_reconstruction.py
```diff
@@ -0,0 +1,598 @@
...
+        """
+
+        self._numPix = data_class.num_pixel_axes[0]
+        self._image_data = data_class.data
+        self._noise_rms = data_class._background_rms
+        self._primary_beam = data_class.primary_beam
+        self._logL_method = data_class._logL_method
+
+        # prepare for the rayshooting
+        self._x_grid_data, self._y_grid_data = data_class.pixel_coordinates
```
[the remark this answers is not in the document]
nanz6: For this comment, are you referring to the source_pixel_grid_class?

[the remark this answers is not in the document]
sibirrer: No, for the data pixel grid. You can get a supersampled grid with the Numerics module and then there are also the routines that downsample/average the supersampled pixels. This may or may not work for your algorithm.

lenstronomy/SourceReconstruction/pixelated_source_reconstruction.py
```diff
@@ -0,0 +1,598 @@
...
+        :param lens_model_class: LensModel class instance
+        :param source_pixel_grid_class: PixelGrid() class instance (defining the source plane grid)
+        :raises ValueError:
+            - If the source pixel grid has rotational components or non-uniform pixel widths.
+            - If the PSF kernel size is improperly sized for interferometric likelihood methods.
+        """
+
+        self._numPix = data_class.num_pixel_axes[0]
+        self._image_data = data_class.data
+        self._noise_rms = data_class._background_rms
```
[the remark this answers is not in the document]
nanz6: I think we can use data_class.C_D to apply the well estimated noise covariance from the data in the source reconstruction. But we are not able to estimate C_D using the model (i.e. data_class.C_D_model()) because for the reconstruction we do not know the model. 

Do you think we should raise a warning telling about it?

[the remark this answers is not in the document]
sibirrer: I think this is ok. Any linear inversion source reconstruction can not apply the likelihood from the model (circular, or non-linear). We are generally only using the updated model to estimate the Poisson noise when evaluating the likelihood after the source reconstruction.

sibirrer (approved):
Thank you very much @nanz6 ! I approve this PR. I left one optional comment as I saw you in the past putting your name and email address in the init file. Happy to merge it when it's ok for you.

PR #784, opened 2025-09-30
Enable point source simulation for interferometric images, update tes…

nanz6:
…t files

This PR extends the image generation of the ImageModel to correctly simulate point sources when interferometric PSF and Primary Beam (PB) effects are applied.

To ensure the primary beam is applied before convolution, the PB values are interpolated at the exact coordinates of each point source. The point source amplitude is then scaled by this PB (if PB is applied) response before being rendered and convolved by the PSF.


sibirrer (approved):
Looks great, thank you @nanz6 !

PR #831, opened 2026-04-08
Fix gauss dec missing methods

furcelay:
Hi,

This is a minor change for `gauss_decomposition.py` to allow these profiles to be used when solving the lens equation. The lens solver requires the `set_static` and `set_dynamic` methods.


sibirrer (changes requested):
Thank you @furcelay! Can you try to instead inherit the LensProfileBase class? This should have been done for all profiles such that such features are supported. Thank you very much!

sibirrer (approved):

PR #833, opened 2026-04-08
Integrating jampy kinematics

furcelay:
Hi,

This pull request integrates the Jampy package as an alternative to Galkin for the `KinematicAPI`. It adds a new module `JAMPy` that implements a wrapper for Jampy, and relies on the Galkin classes (Aperture, Anisotropy, and PSF), which are updated for this purpose. 

This package can be selected by setting `kinematics_backend='jampy'` when instantiating `KinematicAPI`, and it supports both spherical and axisymmetric modeling via the `axial_symmetry` argument.

 I've set Jampy as the default backend, though Galkin could be used as well. Jampy requires the mass and light models to be a multi-Gaussian, and the MGE is handled by `KinematicAPI` before calling the `JAMWrapper`. The MGE is computed using the `mgefit` package rather than the Lenstronomy implementation, as I found it to be more accurate under the same hyperparameters. 

This pull request also includes new PSFs: multi-Gaussian and pixelated, which are currently only compatible with Jampy or Galkin over a regular grid (no MC integration). 

The apertures are sampled over a regular grid using the `aperture_sample` method, then integrated to the desired number of values using the `aperture_downsample` method. The sampling density can be set with the `delta_pix` argument when defining the aperture, which defaults to FWHM/3, and further increased using the `supersampling_factor` when computing the velocity dispersion. Also, it includes a paddig for convolution, which defaults to 3*psf_sigma.


furcelay:
I'll be working on the code coverage. I'll probably finish it by tomorrow.

furcelay:
The pull request is ready for review! :)

lenstronomy/Analysis/kinematics_api.py
```diff
@@ -461,44 +543,22 @@ def kinematic_lens_profiles(
                     kwargs_lens_i["grid_interp_x"] -= center_x_i
                     kwargs_lens_i["grid_interp_y"] -= center_y_i
                 else:
-                    kwargs_lens_i = {
-                        k: v
-                        for k, v in kwargs_lens[i].items()
-                        if not k in ["center_x", "center_y"]
```
sibirrer, line 467: just checking: if the centers of the mass model are not at (0,0), the mass profile would not be centered?

lenstronomy/Analysis/kinematics_api.py
```diff
@@ -461,44 +543,22 @@ def kinematic_lens_profiles(
...
         if MGE_fit is True:
-            if kwargs_mge is None:
-                raise ValueError("kwargs_mge needs to be specified!")
-            if theta_E is None:
-                raise ValueError(
-                    "rough estimate of the Einstein radius needs to be provided to "
-                    "compute the MGE!"
-                )
-            r_array = np.logspace(-4, 2, 200) * theta_E
-            if self._lens_model_list[0] in ["INTERPOL", "INTERPOL_SCLAED"]:
```
sibirrer, line 480: just checking, INTERPOL is not supported anymore or is the new way generic enough to do it?

lenstronomy/Analysis/kinematics_api.py
```diff
@@ -660,31 +744,22 @@ def _setup_light_parameters(
         for i, light_model in enumerate(self._lens_light_model_list):
             if model_kinematics_bool[i] is True:
                 light_profile_list.append(light_model)
-                kwargs_lens_light_i = {
-                    k: v
-                    for k, v in kwargs_lens_light[i].items()
-                    if not k in ["center_x", "center_y"]
```
sibirrer, line 666: Just checking, if the centers are not at (0,0), does this impact the calculation?

lenstronomy/GalKin/aperture_types.py
```diff
@@ -1,33 +1,146 @@
 __author__ = "sibirrer"
 
 import numpy as np
-
 from lenstronomy.Util.package_util import exporter
 
 export, __all__ = exporter()
 
 
+class ApertureBase:
```
sibirrer, line 9: can you document some definitions in this class a bit better? Thank you!

lenstronomy/GalKin/aperture_types.py
```diff
@@ -79,24 +184,73 @@ def slit_select(ra, dec, length, width, center_ra=0, center_dec=0, angle=0):
...
 @export
-class Frame(object):
+def make_slit_grid(delta_pix, length, width, center_ra=0, center_dec=0, angle=0):
+    """
+    :param delta_pix: size of the sub-pixels that samples the aperture for integration
+    :param length: length of slit
+    :param width: width of slit
+    :param center_ra: center of slit
+    :param center_dec: center of slit
+    :param angle: orientation angle of slit, angle=0 corresponds length in RA direction
```
sibirrer: I guess this is in radian?

lenstronomy/GalKin/aperture_types.py
```diff
@@ -152,10 +298,12 @@ def frame_select(ra, dec, width_outer, width_inner, center_ra=0, center_dec=0, a
 
 
 @export
-class Shell(object):
+class Shell(ApertureBase):
     """Shell aperture."""
 
-    def __init__(self, r_in, r_out, center_ra=0, center_dec=0):
+    def __init__(
+        self, r_in, r_out, center_ra=0, center_dec=0, delta_pix=0.1, padding_arcsec=0.0
```
sibirrer, line 328: can you document delta_pix and padding_arcsec in docstring?

lenstronomy/GalKin/aperture_types.py
```diff
@@ -265,115 +502,251 @@ def aperture_select(self, ra, dec):
...
+            x_grid_supersampled, y_grid_supersampled, angle=angle
+        )
+        return x_grid_supersampled, y_grid_supersampled
+    else:
+        return x_grid, y_grid
 
 
-@export
-def shell_ifu_select(ra, dec, r_bin, center_ra=0, center_dec=0):
+def _rotate(x, y, angle):
```
sibirrer, line 738: can you document in what direction the rotation is?

lenstronomy/JAMPy/mge.py
```diff
@@ -0,0 +1,252 @@
+from lenstronomy.LensModel.lens_model import LensModel
+from lenstronomy.LightModel.light_model import LightModel
+from lenstronomy.Analysis.lens_profile import LensProfileAnalysis
+from lenstronomy.Analysis.light_profile import LightProfileAnalysis
+from mgefit import mge_fit_1d as mge
+import numpy as np
+from copy import deepcopy
+
+
+class MGEMass:
```
sibirrer, line 10: I wonder how similar or different these MGE fits are relative to the ones in the KinematicsAPI / LightProfileAnalysis etc. A comment on it might be worth it.

sibirrer (changes requested):
Looks very good @furcelay! Thank you very much!
I have some minor comments only

lenstronomy/Analysis/kinematics_api.py
furcelay: That's right, unless the aperture center is at the same coordinates, it would be shifted. I can restore the previous behavior.

lenstronomy/Analysis/kinematics_api.py
furcelay: The case of an INTERPOL profile is managed inside the MGEMass class

lenstronomy/Analysis/kinematics_api.py
furcelay: The same as for the mass profile, it is not recentered at (0, 0)

lenstronomy/Analysis/kinematics_api.py
sibirrer: Yeah, would be great to do that as some of the pipelines effectively assumed the mass density profile to be centered at (0,0), or at least make a warning when they are not the same. Mass and light parameter fits might always slightly misalign but we assume for the kinematics usually that they are perfectly aligned

lenstronomy/JAMPy/mge.py
furcelay: The difference is only at large radii, but it can be quite significant. For example, here both fits are with 20 gaussians and the same initial radial profile to be fitted:
mge_sersic
mge_epl

sibirrer: this also depends on the choices within the lenstronomy mge fit, correct? See box [4] in this notebook: https://github.com/TDCOSMO/7LensMilestone/blob/main/Likelihoods/MilestoneLikelihood/TDCOSMO_sample/tdcosmo_sample.ipynb

furcelay: Yes, there is more control if using the mge fit function directly instead of the `multi_gaussian_decomposition` method of `LensProfileAnalysis` /`LightProfileAnalysis`, which doesn't allow setting the limits of the radial profile. But for the Isothermal example here, I've used the same limits.

lenstronomy/Analysis/kinematics_api.py
furcelay: Okay, I will set the center at (0, 0)

furcelay:
Hi @sibirrer! The comments are addressed in the last commit

sibirrer:
@furcelay the pre-commit CI fails. Perhaps there needs to be a blank line between general description of a definition and the input parameter description

furcelay:
@sibirrer, thanks for the review! I believe now it should be ready :)

sibirrer (approved):

PR #837, opened 2026-04-17
Enable mask_starred in PSF iteration with JAX compatibility fixes

Guanlin-Liu:
This PR fixes several issues encountered when using the `starred` method

Changes:
- `lenstronomy/Data/psf.py`: fix numpy and JAX array type conflict
- `lenstronomy/Workflow/psf_fitting.py`: add PSF iteration mask support for starred 
- `lenstronomy/Sampling/Pool/multiprocessing.py`: fix deadlock caused by JAX and fork mode in multiprocessing (just a simple and straightforward approach)

### DIFF
--- a/lenstronomy/Data/psf.py
+++ b/lenstronomy/Data/psf.py
@@ -119,7 +119,9 @@ class PSF(object):
             if kernel_point_source_normalisation is True:
                 # self._psf_variance_map /= np.sum(kernel_point_source) ** 2
                 # Handle the inconsistency between jax and numpy arrays for the psf variance map.
-                if isinstance(self._psf_variance_map, jax.Array) or isinstance(kernel_point_source, jax.Array):
+                if isinstance(self._psf_variance_map, jax.Array) or isinstance(
+                    kernel_point_source, jax.Array
+                ):
                     psf_var = np.asarray(self._psf_variance_map)
                     kernel = np.asarray(kernel_point_source)
                     psf_var /= np.sum(kernel) ** 2
--- a/lenstronomy/Data/psf.py
+++ b/lenstronomy/Data/psf.py
@@ -1,5 +1,5 @@
 import copy
-
+import jax
 import numpy as np
 import lenstronomy.Util.kernel_util as kernel_util
 import lenstronomy.Util.util as util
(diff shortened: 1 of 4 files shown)


Guanlin-Liu:
Hi @martin-millon , I've opened this PR but found I can't choose reviewer from settings. So I’m afraid this is the only way I can inform you. Could you please review it? I'm very sorry to disturb you. Thanks!

sibirrer:
@Guanlin-Liu thank you very much! I am not sure you can test the jax setting (see coverage report). If it's possible, that would be great.

lenstronomy/Data/psf.py
```diff
@@ -117,7 +117,17 @@ def __init__(
                         "psf_variance_map are on the down-sampled pixel scale."
                     )
             if kernel_point_source_normalisation is True:
-                self._psf_variance_map /= np.sum(kernel_point_source) ** 2
+                # self._psf_variance_map /= np.sum(kernel_point_source) ** 2
+                # Handle the inconsistency between jax and numpy arrays for the psf variance map.
+                if isinstance(self._psf_variance_map, jax.Array) or isinstance(
```
ahuang314: It might be better to just directly convert kernel_point_source and psf_variance_map to numpy arrays at the start, rather than performing this type-check. That way, an explicit JAX import is not necessary.

Guanlin-Liu:
@sibirrer hi professor, I've added a JAX test using pytest.importorskip, so the test 
will be skipped if JAX is not available in the CI environment. 

The CI failure appears to be unrelated to my changes, it's caused by 
a compatibility issue between the docformatter/untokenize dependency 
and Python 3.14 in the CI environment.

sibirrer:
Thank you very much @Guanlin-Liu! It looks like the CI fails on a line that you introduced: https://github.com/lenstronomy/lenstronomy/actions/runs/24614070880/job/71973915300?pr=837#step:5:510
The pre-commit part indeed, you can ignore (try to sync all the changes from main into your PR, perhaps that helps as well)

sibirrer (approved):
Thank you very much @Guanlin-Liu !

PR #844, opened 2026-04-29
introduce jax flag for when param class is called from jaxtronomy

ahuang314:
1. introduce jax flag for when param class is called from jaxtronomy
2. Allow `solver_type` to be `None`

### DIFF
--- a/lenstronomy/Sampling/parameters.py
+++ b/lenstronomy/Sampling/parameters.py
@@ -291,6 +291,7 @@ class Param(object):
         # Required for JAXtronomy to initialize Image2SourceMapping class with JAXtronomy profiles
         if _jax:
             from jaxtronomy.Util import class_creator as class_creator_jax
+
             _class_creator = class_creator_jax
         else:
             _class_creator = class_creator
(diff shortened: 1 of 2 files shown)


lenstronomy/Sampling/parameters.py
```diff
@@ -286,13 +288,20 @@ def __init__(
         else:
             num_lens_planes = len(list(set(self._lens_redshift_list)))
 
+        # Required for JAXtronomy to initialize Image2SourceMapping class with JAXtronomy profiles
+        if _jax:
+            from jaxtronomy.Util import class_creator as class_creator_jax
```
sibirrer, line 293: would it be possible to change the design of the lenstronomy class creator instead of the option switch?
I am just worried a bit about importing (too often) jaxtronomy in lenstronomy and then doing some of it wise versa, potentially creating a infinity loop

sibirrer: or can you have the jax_flag further down within the class_creator definition?

sibirrer (commented):
I have a comment to see if there might be a more elegant solution

lenstronomy/Sampling/parameters.py
ahuang314: I think moving the jax flag further down into the class creator might create more clutter; it may be cleaner to keep such flags at the top level. Moving the flag also would not solve the potential issue of circular imports.

I cannot think of a good solution to this, other than just duplicating code into JAXtronomy if we want to avoid having excessive JAX flags and imports within lenstronomy.

sibirrer (approved):

PR #848, opened 2026-05-13
turn off bound checks for kwargs_special when using jaxtronomy

ahuang314:
For compatibility with JAXtronomy, bound checks at the Param class level are turned off.
Also removes a redundant jax flag in the arguments of args2kwargs()


sibirrer (approved):
Thank you!

PR #850, opened 2026-05-16
fixed if statement in AdaptiveGrid

ahuang314:
This PR fixes an if statement that would _always_ execute during image simulation if using AdaptiveGrid, since the `_x_sub_grid` attribute doesn't actually exist.

### DIFF
--- a/lenstronomy/ImSim/Numerics/grid.py
+++ b/lenstronomy/ImSim/Numerics/grid.py
@@ -87,7 +87,7 @@ class AdaptiveGrid(Coordinates1D):
 
         :return: 1d arrays of subpixel grid coordinates
         """
-        if not hasattr(self, "_x_sub_grid"):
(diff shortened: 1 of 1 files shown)


sibirrer (approved):
Thank you for the catch! Did not notice it. Hope things go faster with this fix

sibirrer:
@ahuang314 not sure it is related to this PR or the recent change in JAXtronomy, there is now a CI failing: https://github.com/lenstronomy/lenstronomy/actions/runs/25951174725/job/76289192081 
Looks like the JAX test is triggered but then breaks

ahuang314:
Yup, there are a few places in lenstronomy where it tests the use_jax flag (like in LensModel). And since I turned off default 64 bits in JAXtronomy, those lenstronomy tests are failing unless I turn 64 bits back on

PR #851, opened 2026-05-16
fix tests involving use_jax flag to enable 64 bit floats

ahuang314:


sibirrer (approved):

PR #862, opened 2026-05-28
fix bug in jampy kinematics backend

furcelay:
There is a bug in the Jampy kinematics backend: in KinematicAPI, if MGE is on, the ellipticities are not copied from the original profile to the MGE, so the resulting model is always spherical. This pull request fixes this bug and includes a related test.

Also, I'm adding an option, kwargs_jampy, to pass additional Jampy options, which can be used, for example, to change the convolution step size for higher accuracy.


sibirrer (approved):

PR #866, opened 2026-06-11
fix JAMWrapper bug due to ellipticity convention

furcelay:
Fix bug in the Jampy wrapper due to different ellipticity convention between lenstronomy (product average radius) and Jampy (major axis radius).


sibirrer (approved):
Thank you @furcelay !

Issue #845, opened 2026-05-06
Compatibility with Python 3.12


Hellseher:
Hi,

We, at Guix, initiated migration to Python 3.12 from 3.11.

I've noticed `lenstronomy@1.13.5` started failing on one test where `pkgutils` is involved:

```
[gw0] linux -- Python 3.12.12 /gnu/store/nizx2naydfz73pjng6yyf7y1588c30qg-python-wrapper-3.12.12/bin/python

    def test_short_and_laconic():
        import lenstronomy as ls
    
>       ls.short()

test/test_Util/test_package_util.py:7: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

_laconic = False

    @export
    def short(_laconic=False):
        """Activate shortcut mode.
    
        Usage:
            import lenstronomy as ls
            ls.short()
    
            lens_model = ls.LensModel.lens_model.LensModel(...)
            source_model = ls.LightModel.light_model.LightModel(...)
            image_model = ls.ImSim.image_model.ImageModel(...)
        """
        import pkgutil
        import lenstronomy
    
        to_add = dict()
        all_modules = dict()
    
        for loader, module_name, is_pkg in pkgutil.walk_packages(lenstronomy.__path__):
            # This deep internal module relies heavily on 'multiprocessing',
            # which may not be installed
            if module_name == "Sampling.Pool.multiprocessing":
                continue
    
            # Load the module
>           module = all_modules[module_name] = loader.find_module(module_name).load_module(
                                                ^^^^^^^^^^^^^^^^^^
                module_name
            )
E           AttributeError: 'FileFinder' object has no attribute 'find_module'

```
I could not spot difference in documentation, maybe it was kept undocumented somewhere:
https://docs.python.org/3.12/library/pkgutil.html#pkgutil.walk_packages
https://docs.python.org/3.11/library/pkgutil.html#pkgutil.walk_packages

ahuang314:
Addressed in #870. Python 3.12 deprecated the `find_module` and `load_module` utilities in `importlib`.

PR #870, opened 2026-06-22
expand support to python 3.12 and 3.13

ahuang314:
JAX will be dropping support for Python 3.11 on July 15th. Since JAXtronomy relies partially on lenstronomy, this PR ensures that lenstronomy supports Python versions 3.12 and 3.13.

This PR also addresses #845.

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -8,15 +8,12 @@ on:
     branches:
       - main
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.version }}
-  cancel-in-progress: false
-
 jobs:
   build:
 
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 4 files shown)


sibirrer:
It looks like jampy has an issue. @furcelay did jampy release a new version that is incompatible? Perhaps downgrade jampy for the time being

furcelay:
> It looks like jampy has an issue. @furcelay did jampy release a new version that is incompatible? Perhaps downgrade jampy for the time being

That's right, there was a jampy upgrade. You can keep the version fixed for now, I'll check it next week to make sure ir works with the new one.

ahuang314:
Setting scipy<1.18.0 for the time being due to a bug with interpolation (https://github.com/scipy/scipy/issues/25471) causing tests to fail on Python version 3.12+

sibirrer (approved):

PR #872, opened 2026-07-11
Fixed bug in elliptical implementation for non-zero PA

Hadrien-Pgnt:
Fixing a bug noticed by @ Hengkai Ding in the implementation of elliptical multipoles: the reference ellipse used to define the multipole coordinates was not properly rotated.

Now, the EllipticalMultipole() class takes an additional argument (`phi_ref`) which corresponds to the PA of the reference ellipse in the xy plane. I also renamed `phi_m` into  `varphi_m` in that class to emphasize the fact that it is not a polar angle but a difference in eccentric anomaly. For backwards compatibility, the name of this parameter in EPL_MULTIPOLE_M1M3M4_ELL and EPL_MULTIPOLE_M3M4_ELL are still `delta_phi_m[.]`.

Also added some additional tests to make sure that the potential, deflection field, and hessian get correctly rotated.


sibirrer:
Thank you @Hadrien-Pgnt! There are two tests failing in the CI

dangilman:
@Hadrien-Pgnt thank you very much! When sampling parameters of the multipole terms, do you think it makes more sense to implement uniform priors on the eccentric anomaly, or a uniform prior on the polar angle?

sibirrer (approved):
Thank you very much @Hadrien-Pgnt. 
@dangilman feel free to merge when you think it is ok

Hadrien-Pgnt:
> @Hadrien-Pgnt thank you very much! When sampling parameters of the multipole terms, do you think it makes more sense to implement uniform priors on the eccentric anomaly, or a uniform prior on the polar angle?

@dangilman 
There is perhaps a physical prior that can be determined, but in the absence of further information, I believe it makes more sense to use uniform priors on the eccentric anomaly by default: since it is the parameter that goes into the elliptical multipole formulas, it will be the most uninformative choice. I see 2 ways of thinking about this:
- if we use a uniform prior on the eccentric anomaly, we have more samples for perturbations near the direction of the semi-major axis, which is where things vary more rapidly, so we are "compensating" for the geometric flattening effect.
- with a uniform prior on the eccentric anomaly (in red below), the sine and cosine components have similar distributions, whereas if we assume a uniform prior on the polar angle (in blue below), we are introducing an asymmetry between the two (see below for m=1, q=0.6) 
image

Let me know what you think !

dangilman:
@Hadrien-Pgnt thank you for checking this!

PR #873, opened 2026-07-17
Update lenstronomy Jampy Wrapper to use the latest Jampy version.

furcelay:
Update lenstronomy Jampy Wrapper to use the latest Jampy version with the spectral method.

This method is faster and allows more general descriptions for the anisotropy and light/mass distribution. However, the axi_cyl alignment is deprecated in this new version in favor of an angular variable anisotropy within the spherical alignment. For the moment, only radially variable anisotropies are implemented in this wrapper.

----
Edit: 
The axi_cyl alignment is implemented using an angle-dependent anisotropy function within the spherical alignment following Cappellari 2026.


lenstronomy/GalKin/anisotropy.py
```diff
@@ -78,8 +78,33 @@ def delete_anisotropy_cache(self):
         if hasattr(self._model, "delete_cache"):
             self._model.delete_cache()
 
-    def jampy_params(self, kwargs):
-        return self._model.jampy_params(**kwargs)
+    def jampy_beta(self, kwargs, symmetry="spherical"):
+        """Returns the anisotropy parameter in the new Jampy spectral method format.
+        Either a constant anisotropy or a callable beta(r, theta). Currently only a
```
sibirrer: can you describe this a bit in more detail?

test/test_JAMPy/test_jam_wrapper.py
```diff
@@ -473,110 +473,112 @@ def test_axi_dispersion_shells(self):
...
-            [self.kwargs_light_spherical],
-            self.kwargs_anisotropy,
-            supersampling_factor=5,
-        )
-        npt.assert_allclose(sigma_v_jam, sigma_v_galkin, rtol=1e-2)
+        npt.assert_allclose(sigma_v_jam, sigma_v_galkin, rtol=5e-2)
+
+
+# The axi_cyl was deprecated in favor of an angular variable anisotropy
+# TODO: needs to be updated to the new Jampy spectral method
```
sibirrer: do you want to update add these tests?

sibirrer (changes requested):
it looks like the new jampy version does not support python 3.11 anymore. This is fine. Perhaps add a condition that the specific calls only get executed if the python environment is >3.11 and otherwise give a raise with a proper error message?

furcelay:
Thank you @sibirrer for the review! I will make the requested changes and see if I can come up with an alternative for the anisotropy in the cylindrical alignment.

furcelay:
Hi @sibirrer, I've re-implemented the axi_cyl alignment with the new Jampy spectral method (Cappellari 2026). This is not equivalent to the previous axi_cyl from Jampy <9 (Cappellari 2008), but it is instead closer to what axi_cyl was originally intended.

sibirrer:
Thank you @furcelay!
I see that the python version 3.11 raises an error. Instead of breaking the CI, perhaps, we can change the tests such that if 3.11 or earlier, certain tests with jampy are simply not executed

furcelay:
Hi @sibirrer, now it is skipping the Jampy tests if Python <3.12

sibirrer (approved):
Wonderful! This looks great - thank you @furcelay !

PR #874, opened 2026-07-17
Minor upgrade to Jampy/Galkin kinematics

furcelay:
Added an option for a general aperture described by a non-regular grid of (x, y, bins), and also allow for non-consecutive bin numbers.


sibirrer (approved):
Thank you very much @furcelay!

PR #877, opened 2026-07-28
Fixed caustics_plot color assignment

### LINKED ISSUE #876: Critical curve color is always set to the caustic color for `caustics_plot`
Hello Lenstronomy Enjoyers,

There is an issue with the `caustics_plot` function for plotting critical curves and caustics. Currently, the code forces the critical curve color (the "critical_curve_color" argument in the `kwargs_caustics` dictionary) to be the same color as the caustic ("color" in `kwargs_caustics`). At line 330, the code attempts to overwrite "color" with `critical_curve_color` using `setdefault`, but since "color" is already assigned (by default to 'g'), the overwrite does not happen. Instead, "color" should be directly assigned: `kwargs_caustics["color"] = critical_curve_color`. 

https://github.com/lenstronomy/lenstronomy/blob/bf50e455de0a8661732c656db938095ad03556fc/lenstronomy/Plots/lens_plot.py#L330

seanxuseanxu:
Fixes #876 

Hello Lenstronomy Enjoyers,

Here is a fix for the critical curve and caustics plotting color assignment behavior. Is it possible to also change the "color" key in `kwargs_caustics` to "caustic_color" so that it is more informative?

### DIFF
--- a/lenstronomy/Plots/lens_plot.py
+++ b/lenstronomy/Plots/lens_plot.py
@@ -327,7 +327,7 @@ def caustics_plot(
         color=caustic_color,
         **kwargs_caustics,
     )
-    
+
     plot_util.plot_line_set(
         ax,
         pixel_grid,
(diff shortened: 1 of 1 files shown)


sibirrer (approved):
Thank you very much @seanxuseanxu !

PR #878, opened 2026-07-31
Fix DeprecationWarning in Numpy>=2

bryce-wedig:
`lenstronomy.ImSim.Numerics.convolution.PixelKernelConvolution.convolution2d()` raises the following deprecation warning on Numpy>=2.0:

`DeprecationWarning at /Users/brycewedig/Documents/GitHub/lenstronomy/lenstronomy/ImSim/Numerics/convolution.py:121
    `axes` should not be `None` if `s` is not `None` (Deprecated in NumPy 2.0). In a future version of NumPy, this will raise an error and `s[i]` will correspond to the size along the transformed axis specified by `axes[i]`. To retain current behaviour, pass a sequence [0, ..., k-1] to `axes` for an array of dimension k.`

Following the recommendation, pass a tuple of the appropriate length as `axes` to retain the current behavior.

### DIFF
--- a/lenstronomy/ImSim/Numerics/convolution.py
+++ b/lenstronomy/ImSim/Numerics/convolution.py
@@ -118,8 +118,10 @@ class PixelKernelConvolution(object):
         # sure we only call rfftn/irfftn from one thread at a time.
         if not complex_result and (_rfft_mt_safe or _rfft_lock.acquire(False)):
             try:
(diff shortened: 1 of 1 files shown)


sibirrer (approved):
Thank you very much @bryce-wedig! This removes a lot of warnings and makes lenstronomy fit for numpy>2.0 :)

PR #879, opened 2026-08-02
Add multiband offset sampling support

HuiminQu:
PR description：
This PR adds support for jointly sampling multi-band astrometric offsets (shift and rotation) in the likelihood module.

Main changes:
- Add multi-band offset parameters in SpecialParam.
- Apply temporary coordinate transformations during likelihood evaluation.
- Restore original coordinate states after evaluation to avoid side effects.
- Add tests covering parameter conversion, coordinate updates, image response, and repeated evaluations.
- Add a demonstration notebook.

The implementation updates internal coordinate grids during likelihood evaluation while preserving the original data state.

### DIFF
--- a/lenstronomy/Sampling/Likelihoods/image_likelihood.py
+++ b/lenstronomy/Sampling/Likelihoods/image_likelihood.py
@@ -1,5 +1,6 @@
 import numpy as np
 from lenstronomy.Util import class_creator
+from lenstronomy.Util import util
 
 __all__ = ["ImageLikelihood"]
 
@@ -50,6 +51,7 @@ class ImageLikelihood(object):
         self._source_marg = source_marg
         self._linear_prior = linear_prior
         self._check_positive_flux = check_positive_flux
+        self.num_bands = len(multi_band_list)
 
     def logL(
         self,
@@ -71,19 +73,27 @@ class ImageLikelihood(object):
         :param kwargs_extinction: extinction parameter keyword argument list according to LightModel module
         :return: log likelihood of the data given the model, linear parameter inversion list
         """
-        logL, param = self.imSim.likelihood_data_given_model(
-            kwargs_lens,
-            kwargs_source,
-            kwargs_lens_light,
-            kwargs_ps,
-            kwargs_extinction=kwargs_extinction,
-            kwargs_special=kwargs_special,
-            source_marg=self._source_marg,
-            linear_prior=self._linear_prior,
-            check_positive_flux=self._check_positive_flux,
-        )
-        if np.isnan(logL) is True:
-            return -(10**15), param
+
+        backup_grids = self._apply_multiband_offsets(kwargs_special)
+        try:
+            logL, param = self.imSim.likelihood_data_given_model(
+                kwargs_lens,
+                kwargs_source,
+                kwargs_lens_light,
+                kwargs_ps,
+                kwargs_extinction=kwargs_extinction,
+                kwargs_special=kwargs_special,
+                source_marg=self._source_marg,
+                linear_prior=self._linear_prior,
+                check_positive_flux=self._check_positive_flux,
+            )
+
+            if np.isnan(logL) is True:
+                return -(10**15), param
+            
+        finally:
+            self._restore_multiband_offsets(backup_grids)
+        
         return logL, param
 
     @property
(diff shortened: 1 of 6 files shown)


test/test_Sampling/test_multiband_offset.py
```diff
@@ -0,0 +1,746 @@
...
+
+    numerics_kwargs = {
+        "supersampling_factor": 1,
+        "supersampling_convolution": False,
+    }
+
+    # same coordinate system initially
+    _, _, ra_at_xy_0, dec_at_xy_0, _, _, Mpix2coord, _ = (
+        util.make_grid_with_coordtransform(
+            numPix=num_pix,
```
sibirrer: this needs to be re-named to 'num_pix' to be up to date with latest lenstronomy version

lenstronomy/Sampling/param_group.py
```diff
@@ -344,3 +344,152 @@ def kwargs_upper(self):
...
+
+            offset = sampled_offsets[band] if band < len(sampled_offsets) else {}
+            fixed_band = fixed_offsets[band] if band < len(fixed_offsets) else {}
+            for param in ["dx", "dy", "angle"]:
+                if param not in fixed_band:
+                    args.append(offset.get(param, 0))
+        return args
+
+    def get_params(self, args, i, kwargs_fixed, kwargs_lower=None, kwargs_upper=None):
+        if not self.on:
```
sibirrer, line 426: is self.on defined?

lenstronomy/Sampling/Likelihoods/image_likelihood.py
```diff
@@ -71,19 +73,27 @@ def logL(
...
-            kwargs_special=kwargs_special,
-            source_marg=self._source_marg,
-            linear_prior=self._linear_prior,
-            check_positive_flux=self._check_positive_flux,
-        )
-        if np.isnan(logL) is True:
-            return -(10**15), param
+
+        backup_grids = self._apply_multiband_offsets(kwargs_special)
+        try:
```
sibirrer: I don't like try: except: statements as it makes it in transparent why certain things are not executed and the error messages will not show up

lenstronomy/Sampling/Likelihoods/image_likelihood.py
```diff
@@ -119,3 +129,149 @@ def reset_point_source_cache(self, cache=True):
         :return: None
         """
         self.imSim.reset_point_source_cache(cache=cache)
+
+    def _apply_multiband_offsets(self, kwargs_special):
```
sibirrer: this function accesses many private variables in other classes.
I would suggest implementing a feature through the ImageModel() class in the Data() class that allows to change temporarily the coordinate frame instead

sibirrer (changes requested):
Thank you very much @HuiminQu!
I made some comments. I am also happy to provide a better API for these changes in the Data() and ImageModel() classes

lenstronomy/Sampling/param_group.py
HuiminQu: Yes, self.on is defined as a property in MultiBandOffsetParam, returning self._on, which is initialized in __init__.

test/test_Sampling/test_multiband_offset.py
HuiminQu: Updated to num_pix to match the current lenstronomy API.

lenstronomy/Sampling/Likelihoods/image_likelihood.py
HuiminQu: Removed the try/finally block and the manual coordinate backup/restore. The coordinate updates are now handled through the ImageModel coordinate-frame API.

lenstronomy/Sampling/Likelihoods/image_likelihood.py
HuiminQu: Addressed by using the coordinate-frame update API introduced in #880. The multiband offsets are now passed through ImageModel.update_pixel_grid_coordinates() to update the underlying Data coordinate frame, rather than accessing private variables directly : )

sibirrer:
Thank you very much @HuiminQu! Glad to see that this could be a bit simplified.
As you may see, some lines of code that you added are not represented in a test function.
If you can expand the test function to make sure that all the added lines of code are covered in a test, that would be great

PR #882, opened 2026-08-18
perturber model to subtract shift and hessian 

sibirrer:
This PR adds a feature "perturber_model_list" that can flag deflector models that are only treated at higher order (without constant shift, shear and convergence) to not cause degeneracy with other modeled quantities


ahuang314 (approved):
Looks good to me. Seems straightforward to implement in JAX as well

sibirrer:
Thank you @ahuang314 !
Would be great to having this in JAXtronomy soon to test it out with the JWST lenses

PR #883, opened 2026-08-23
consolidate perturber model class

ahuang314:
Currently, using the PerturberModel class requires 3 extra keyword arguments to be supplied to the LensModel class. In order to use this feature in lens modeling, these 3 arguments would have to be included in `kwargs_model` and propagated through the class creator util functions.

To simplify usage and implementation in e.g. dolphin, this PR removes these extra keyword arguments in favor of placing them within `profile_kwargs_list`. Models in the `lens_model_list` are no longer flagged with the `perturber_model_list` but are instead initialized directly with the `"PERTURBER"` lens type.


sibirrer (approved):
Looks good and makes sense for easier usage - thank you @ahuang314 !

PR #884, opened 2026-08-23
Fix eccentricity upper limit defaults

bryce-wedig:
This PR updates the default upper limits for the ellipticity parameters (`e1` and `e2`) in several light profile classes to use positive values. I suspect this was a typo that got propagated with copy-paste.


ajshajib (approved):
Good catch, thanks a lot for fixing it! 🚀

PR #885, opened 2026-08-24
change scipy requirement to allow >= 1.18.1

ahuang314:
Previously scipy was restricted to <1.18.0 due to a bug (see [PR 870](https://github.com/lenstronomy/lenstronomy/pull/870#issuecomment-4772950157)). The bug has been fixed in 1.18.1.


sibirrer (approved):
Thank you!

PR #887, opened 2026-08-30
Add gradient beyond prior bound to bring back walkers

ajshajib:
## Summary
`Likelihood.logL` returned a flat `-1e18` penalty for any out-of-bounds parameter vector, regardless of how far outside the bounds it was. Because that constant offset cancels out of the Metropolis acceptance ratio between two out-of-bounds points, walkers that land in that region see zero likelihood gradient. This can happen if a set of walkers are initiated in the MCMC that land outside the prior bounds. With `emcee`'s stretch move (multiplicative, referenced to another walker's position), this manifests as an unconstrained multiplicative random walk: a walker can drift from O(1) to O(10^100+) over a few hundred steps with 100% acceptance the entire time, which diminishes the effectiveness of the number of walkers used.

## Change
`check_bounds` now accumulates a penalty that scales with the (bound-range normalized) distance past each violated bound, summed over all violated parameters, instead of a fixed value. `logL` adds this penalty on top of the `-1e18` floor. The floor still guarantees any out-of-bounds point scores worse than a legitimate finite likelihood, but the penalty term no longer cancels between two out-of-bounds points — restoring a gradient that pulls divergent walkers back toward the allowed region.

### DIFF
--- a/lenstronomy/Sampling/likelihood.py
+++ b/lenstronomy/Sampling/likelihood.py
@@ -351,11 +351,6 @@ class Likelihood(object):
                 args, self._lower_limit, self._upper_limit, verbose=verbose
             )
             if bound_hit is True:
-                # the -1e18 floor guarantees any out-of-bounds point scores
-                # worse than a legitimate finite likelihood; it cancels out of
-                # any Metropolis ratio between two out-of-bounds points, so the
-                # (distance-based) penalty is what still provides a restoring
-                # gradient back toward the allowed region
                 return -(10**18) - penalty
         # extract parameters
         kwargs_return = self.param.args2kwargs(args)
--- a/lenstronomy/Sampling/likelihood.py
+++ b/lenstronomy/Sampling/likelihood.py
@@ -449,15 +449,15 @@ class Likelihood(object):
     @staticmethod
     def check_bounds(args, lowerLimit, upperLimit, verbose=False):
         """Checks whether the parameter vector has left its bounds. If so, returns a
-        penalty that grows with the (bound-range-normalized) distance past the
-        violated bound(s), summed over all violated parameters.
+        penalty that grows with the (bound-range-normalized) distance past the violated
+        bound(s), summed over all violated parameters.
 
-        A flat penalty (independent of how far a parameter is out of bounds) creates
-        a plateau with no likelihood gradient, which lets samplers with proposals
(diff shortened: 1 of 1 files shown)


lenstronomy/Sampling/likelihood.py
```diff
@@ -443,21 +448,34 @@ def log_likelihood(self, kwargs_return, verbose=False):
...
+        """Checks whether the parameter vector has left its bounds. If so, returns a
+        penalty that grows with the (bound-range-normalized) distance past the violated
+        bound(s), summed over all violated parameters.
+
+        A flat penalty (independent of how far a parameter is out of bounds) creates a
+        plateau with no likelihood gradient, which lets samplers with proposals
+        referenced to other walkers' positions (e.g. emcee's stretch move) drift
+        arbitrarily far from the bounds once they land in that region, since every out-
+        of-bounds point looks equally "bad". Scaling the penalty with distance keeps a
+        restoring gradient pointing back toward the allowed region.
```
ajshajib:
```suggestion
```

lenstronomy/Sampling/likelihood.py
```diff
@@ -443,21 +448,28 @@ def log_likelihood(self, kwargs_return, verbose=False):
 
     @staticmethod
     def check_bounds(args, lowerLimit, upperLimit, verbose=False):
-        """Checks whether the parameter vector has left its bound, if so, adds a big
-        number."""
+        """Checks whether the parameter vector has left its bounds. If so, returns a
+        penalty that grows with the (bound-range-normalized) distance past the violated
+        bound(s), summed over all violated parameters.
+
```
ajshajib, line 447:
```suggestion
```

lenstronomy/Sampling/likelihood.py
```diff
@@ -351,7 +351,12 @@ def logL(self, args, verbose=False):
                 args, self._lower_limit, self._upper_limit, verbose=verbose
             )
             if bound_hit is True:
-                return -(10**18)
+                # the -1e18 floor guarantees any out-of-bounds point scores
+                # worse than a legitimate finite likelihood; it cancels out of
+                # any Metropolis ratio between two out-of-bounds points, so the
+                # (distance-based) penalty is what still provides a restoring
+                # gradient back toward the allowed region
```
ajshajib:
```suggestion
```

sibirrer (approved):

PR #888, opened 2026-09-06
Pixelated source reconstruction py update

nanz6:
Update:

Updated pixelated_source_reconstruction.py to make it run faster for pixelated source reconstruction tasks. The main reason it now runs faster is that the double for loops in the relevant functions are reduced by using np.matmul or scipy sparse matrix @ np.array, replacing them with either no loop or only one loop.

Additional functions were added to manipulate scipy csc matrices. The inputs and outputs of all original functions remain unchanged, but the internal flow has changed. The modified functions passed all the old test functions, except for the test that checks the printed text of the progress bar, which changed slightly. I've also tested the new and old versions against a large number of different cases to make sure their outputs remain the same with rtol=1e-9 and atol=1e-12.

All documentation was also modified to make the ideas clearer.

pixelated_source_reconstruction.py

- (+) _lens_pixel_source_of_a_rectangular_region_csc_matrix

This function performs the original task of lens_pixel_source_of_a_rectangular_region. It computes the lens mapping from image pixels to source pixels, but the result returned by this function is now stored in scipy csc_matrix form.

- (+) _csc_matrix_to_lensed_sp

Converts the csc_matrix into the old nested-list form used to store the lensing map.

- (changed) lens_pixel_source_of_a_rectangular_region

Combined the above two functions to keep its input and output unchanged from the older version.

- (changed) generate_M_b_diagonal_likelihood and generate_M_b_interferometry_natwt_likelihood

Changed the internal flow and reduced the depth of the for loops to accelerate the computation.

- (changed) lens_pixel_source_of_a_rectangular_region

Also changed the internal flow and reduced the double for loops.

- (changed) sum_sparse_elementwise_product

Changed into a ``@staticmethod`` because it does not rely on self or other instance properties.

- (+) _sparse_convolution_from_image_indices

This function was added to perform the convolution of a set of lensed pixels using their pixel indices and values. It performs the same task as the original sparse_convolution function takes the image pixel indices as input which works better with csc_matrix.

- (changed) sparse_convolution

Now invokes _sparse_convolution_from_image_indices to perform sparse image convolution.

test_pixelated_source_reconstruction.py

- (changed) test_generate_M_b_print

Changed the detection of the printed text from "Running (iteration times vary)" to "Running".

- (+) test_lens_pixel_source_of_a_rectangular_region_csc_matrix

- (+) test_csc_matrix_to_lensed_sp

- (changed) test_sparse_matrix_manipulation_functions

Added tests for _sparse_convolution_from_image_indices.

AI acknowledgement:

This version was inspired by an AI review of the older version of pixelated_source_reconstruction.py. After the AI-generated changes, I reviewed them carefully and found that the main changes were the reduction of for loops and the use of the CSC matrice. I decided to keep these changes, but revised the code myself. The core part of the code, including lensing and bilinear interpolation in the source plane, was not touched by AI. 
AI model used: GPT-5.6 Sol.

### DIFF
--- a/lenstronomy/ImSim/SourceReconstruction/pixelated_source_reconstruction.py
+++ b/lenstronomy/ImSim/SourceReconstruction/pixelated_source_reconstruction.py
@@ -217,9 +217,9 @@ class PixelatedSourceReconstruction(object):
         
         # b_i = (Ls_i)^T(d/sigma_map)
         b = np.matmul(lensed_pixel_conv_set, (image_data * inverse_variance))
-        lensed_pixel_conv_set *= np.sqrt(inverse_variance)
         
         # M_ij = (Ls_i/sqrt(sigma_map))^T(Ls_j/sqrt(sigma_map))
+        lensed_pixel_conv_set *= np.sqrt(inverse_variance)
         M = np.matmul(lensed_pixel_conv_set, lensed_pixel_conv_set.T)
 
         # Enforce exact symmetry to remove small floating-point asymmetries from matrix multiplication.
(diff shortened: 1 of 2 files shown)


sibirrer (approved):
Thank you very much @nanz6 ! Looks all good. Curious what the speed up here is :)

PR #889, opened 2026-09-16
Select the positive quadratic root before the caustic power

### LINKED ISSUE #821: Problem in selecting roots solvequadeq in caustics_epl_shear
In this [function]( https://github.com/lenstronomy/lenstronomy/blob/08bb8dd3de65319f4ddfce80889663a3e49b5216/lenstronomy/LensModel/Solver/epl_shear_solver.py#L313) that produce caustic points with analytical method, there is 1D root calculation and selection <[link](https://github.com/lenstronomy/lenstronomy/blob/08bb8dd3de65319f4ddfce80889663a3e49b5216/lenstronomy/LensModel/Solver/epl_shear_solver.py#L374)> as,

```python
usol = np.array(solvequadeq(cc, bb, aa)).T
xcr_4, ycr_4 = pol_to_cart(b * usol[:, 1] ** (-1 / t) * frac_roverR, theta)
```
if `usol[:, 1]` is -ve `usol[:, 1] ** (-1 / t)` can be nan as these parameters,

```python
theta_E=2.1479577335194115e-06
q=0.595
phi=5.761
e1=0.127
e2=-0.220
gamma=2.636
gamma1=-0.009
gamma2=0.171
```

The correction should be, if `usol[:, 1]<0.`, use `usol[:, 0]` instead.

Regards,
Hemanta
PhD student 
CUHK Hong Kong

shaneraphel:
Closes #821

`caustics_epl_shear` sends the second `solvequadeq` root through `u ** (-1 / t)`. When that root is negative, the real power is NaN and the quad caustic (and the `t > 1` cut) is unusable. The `t <= 1` cut already uses the first root.

This change takes the positive member of the pair before the power. It does not take `abs` of the second root and does not rewrite `solvequadeq`.

On a two-root sample with `t = 2.636`:

| sheet | `u ** (-1/t)` |
| --- | --- |
| leftover `usol[:, 1] = (-0.3, 0.2)` | `(nan, 1.841…)` |
| dest positive member | `(1.416…, 1.841…)` |


sibirrer (commented):
Thank you very much @shaneraphel!
Is there a potential test scenario that tests t>1 since these lines of code are not tested (also not tested in the past). Would be nice to have one.

sibirrer (approved):
Thank you very much @shaneraphel !

shaneraphel:
tanks for review!

